PR #51, opened 2012-12-16
Added two settings to optionally perform stricter json deserializations

daniellm:
- MemberNameMatchHandling: forces case-sensitive lookup of member name.
- MemberTypeConversionHandling: prevents implicit type conversion, e.g. assigning a json string to an int member.

### DIFF
--- a/Src/Newtonsoft.Json/JsonSerializer.cs
+++ b/Src/Newtonsoft.Json/JsonSerializer.cs
@@ -220,6 +220,12 @@ namespace Newtonsoft.Json
       }
     }
 
+
+    /// <summary>
+    /// Gets or sets whether the deserializer replaces lists with default values
+    /// </summary>
+    public virtual NonEmptyListHandling NonEmptyListHandling { get; set; }
+
     /// <summary>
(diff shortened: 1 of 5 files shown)


andresmoschini:
Hi @JamesNK, I am using ASP.NET 5 MVC 6 and I would like to have a case sensitive JSON deserialization and I do not want to add a dependency on Newtonsoft Schema.

I think that the natural place to configure or override it is on `InputFormatter`'s `ContractResolver`, but I cannot find a way to do it since `GetClosestMatchProperty` is who make it case insensitive.

Could you give me any clue?

andresmoschini:
:watch:

daniellm:
Hello again @JamesNK 

7+ years have passed. What's your stance on json validations? My use case is a customer-facing REST API accepting large json objects. I don't want to be lax during deserialization, so that I don't break my customer implementations in case I ever need to become strict (e.g. if I move the endpoint to a different development stack). Could I prepare a new version of this MR for the latest version? Would you accept it?

If not, could you explain how Json.net schema can be used to validate casing and prohibit implicit type conversions (e.g. mapping a json string literal such as "3" to an integer property)? Can it be done without generating actual schemas, just based off C# POCOs?

Thank you!

PR #85, opened 2013-05-09
Allow a serialization binder to be specified for a particular property (using jsonpropertyattribute)

martinbooth:
When attempting to create a SerializationBinder to customize the $type property on json, I realised that if I customized it for one property, i'd be customizing it for them all.

Since there is no context available within BindToType and BindToName, its impossible to know which property is currently being serialized/deserialized. 

Therefore it is possible with this modification to specify a different SerializationBinder per property.. each with their own behaviour.


Bab95:
I am not sure if this is the right place to ask this question but does anybody knows how to make this binder property disappear from Json's?

PR #292, opened 2014-06-02
Add ability to clone the reader. Cloning can be used to return to parsin...

PawelStroinski:
...g position at a time when clone was created given that source stream is in the same position/buffer as when clone was created.


chadwalters:
Any chance you might reconsider this issue? We're using Json.NET over in [Bond ](https://github.com/Microsoft/bond) and this capability would be handy. Implementing this from the outside is even more fiddly.

JamesNK:
It doesn't make sense with JsonTextReader and BsonReader because they over the top of forward only text and binary readers - and reading from one will break the copies - and for JTokenReader it is simple to create a new JTokenReader from the current node.

PR #295, opened 2014-06-02
Add ability to grab delimiters and whitespace during parsing and make it...

PawelStroinski:
... available for read (create JsonTextReader with grabDelimiters option).


JamesNK:
Hi

I've thought about this and I don't want to add this complexity to JsonTextReader, which very few people will use. It is possible for those people to implement there own JsonReader with this feature if they want to.

PR #376, opened 2014-09-12
[ObjectContract] Added support for custom member serialization

OndrejPetrzilka:
Custom member (de)serialization allows me to have more control over serialization and decide at runtime what and how to serialize. Member abstraction (TypeHandling.Auto) works with this too. It was hard/impossible to get it working with Extension Data.

Using custom member (de)serialization and few other classes makes possible this, I'm used to this syntax from SharpDX.IDataSerializable:

```
public class Control : IJsonSerializable
{
    public string Name;
    public Vector2 Position;
    public Vector2 Size { get; set; }
    public HashSet<Control> Controls = new HashSet<Control>();

    // Same method used for serialization and deserialization
    public virtual void Serialize(JsonSerializableContext context)
    {
        if (context.Writing)
        {
            context.Serialize("Name", "--unnamed--");
        }
        else
        {
            context.Serialize("OldName", ref Name);
            context.Serialize("Name", ref Name);
        }

        if (Position != Vector2.Zero) // Decide here if serialize or not (similar to ShouldSerialize)
        {
            context.Serialize("Position", ref Position);
        }
        context.Serialize(() => Size); // Possible like this (this expression gets cached)
        //context.Serialize("Size", ref Size); // Or like this
        context.Serialize("Controls", ref Controls);
    }
}
```

### DIFF
--- a/Src/Newtonsoft.Json/Formatting.cs
+++ b/Src/Newtonsoft.Json/Formatting.cs
@@ -38,6 +38,11 @@ namespace Newtonsoft.Json
         /// <summary>
         /// Causes child objects to be indented according to the <see cref="JsonTextWriter.Indentation"/> and <see cref="JsonTextWriter.IndentChar"/> settings.
         /// </summary>
-        Indented
+        Indented,
+
+        /// <summary>
+        /// Indented formatting with ObjectStart and ArrayStart starting on new line.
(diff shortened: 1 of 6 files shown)


JamesNK:
Thanks for the PR but I don't want to add this feature right now.

PR #387, opened 2014-10-06
Removes the install.ps1 script that opens a browser when installing the package.

damianh:
We all love the lib, but we all don't like having a browser open every time we do a package install. It's borderline obnoxious. Thank you.


SimonCropp:
sorry for the spam people... i am doing a straw pole on nugets that open doco urls on install https://twitter.com/SimonCropp/status/768999257021878272

PR #410, opened 2014-10-31
-Ported to Xamarin.iOS and Xamarin.Android

jstedfast:
Added Visual Studio projects and solutions for Xamarin.iOS (Unified API) and Xamarin.Android


JamesNK:
I prefer to use PCLs.

PR #420, opened 2014-11-09
Deserializing anonymous json into ordinary dictionaries and arrays using a converter

wallymathieu:
A converter to deserialize into array and dictionary. This is useful if you have code that have been using the standard javascript deserializer in .net, but where you want better error messages ;) .


JamesNK:
I think this would be a good option for a converter in a separate NuGet package. I'm trying hard to keep the size of the core Json.NET library small.

wallymathieu:
Will do. I've moved it here:
https://github.com/NewtonsoftJsonExt

Issue #401, opened 2014-10-23
Object reference equality should be used when checking for circular references


D3-LucaPiombino:
In Newtonsoft.Json\Src\Newtonsoft.Json\Serialization\JsonSerializerInternalWriter.cs(252): 

```csharp
private bool CheckForCircularReference(JsonWriter writer, object value, JsonProperty property, JsonContract contract, JsonContainerContract containerContract, JsonProperty containerProperty)
```

(Last commit: https://github.com/JamesNK/Newtonsoft.Json/commit/70d3e9327d5dcd9a1770b3aed5437d6aa0b15802) the code:
```csharp
if (_serializeStack.IndexOf(value) != -1)
```
should be replaced with:
```csharp
if (_serializeStack.IndexOf(itemOnStack => object.ReferenceEquals(itemOnStack,value)) != -1)
```
(or a custom ReferenceEqualityComparer can be created and passed to the other overload)
because if the instance contained in value or any entry in the stack overrides object.Equals it can cause the code to report a false positive.

**Test case:**

```csharp
        public class ValueContainer
        {
            public object Value;

            public override bool Equals(object obj)
            {
                if (obj == null)
                    return false;

                var otherContainer = obj as ValueContainer;

                if (otherContainer != null)
                    return EqualityComparer<object>.Default.Equals(Value, otherContainer.Value);

                return EqualityComparer<object>.Default.Equals(Value, obj);
            }

            public override int GetHashCode()
            {
                return EqualityComparer<object>.Default.GetHashCode(Value);
            }
        }

        [Test]
        public void ShouldCheckForCircularReferencesUsingReferenceEquality()
        {
            var source = new ValueContainer
            {
                Value = new object()
            };

            string json = JsonConvert.SerializeObject(source, Formatting.Indented);
        }
```
I think this behavior was not intended by the author, since checking for circular references in an object graph should only be about referential identity.

JamesNK:
I prefer the current behavior which lets devs customize logic by overriding Equals.

Besides, this is a big breaking change.

johnsparrowuk:
This causes a lot of unnecessary calls to `object.Equals` when your classes have value semantics (which after all, Json.net doesn't care about, it's looking for referential loops). Even when `ReferenceLoopHandling.Serialize` is set the call to object.Equals still fires.

[...]

PR #423, opened 2014-11-11
Circular reference detection reference equality handling

D3-LucaPiombino:
I' ve closed the previous pull request and opened a new one.
I still think that the current behavior(see https://github.com/JamesNK/Newtonsoft.Json/issues/401)  **IS** very fragile.
I undestand that if people are depending on it, the default
behavior cannot be changed, but maybe it should be at least overridable.
In this pull request i've added a new enum(ReferenceComparisonHandling) and a new property that allow to override the behavior, available to all the various levels available for other settings like **ReferenceLoopHandling**, **ItemReferenceLoopHandling** etc... that specify the method that will be used to compare references during the circular reference detection phase.
If the settings is not provided the old behavior is used.

```csharp
new JsonSerializerSettings
{
	Formatting = Formatting.Indented,
	ReferenceComparisonHandling = ReferenceComparisonHandling.ReferenceEquals
}

[JsonObject(ItemReferenceComparisonHandling = ReferenceComparisonHandling.ReferenceEquals)]
[JsonProperty(ReferenceComparisonHandling = ReferenceComparisonHandling.ReferenceEquals)]
[JsonProperty(ItemReferenceComparisonHandling = ReferenceComparisonHandling.ReferenceEquals)]

```

and of course all the other classes that partecipate in the serialization pipeline(**JsonProperty**,**JsonContract**, etc...).
I've added a new set of unit test that verify the intended behavior as the guidelines indicates.
This could have also been implemented in a different way, by creating a new service (like IReferenceResolver) that allow to change the default behavior, but i didn't want to change the code to much. If you prefer this solution let me know , i can make another pull request.

Thanks.

### DIFF
--- a/Build/build.ps1
+++ b/Build/build.ps1
@@ -2,7 +2,7 @@
   $zipFileName = "Json70r1.zip"
   $majorVersion = "7.0"
   $majorWithReleaseVersion = "7.0.1"
-  $nugetPrelease = "beta2"
+  $nugetPrelease = "beta3"
   $version = GetVersion $majorWithReleaseVersion
   $packageId = "Newtonsoft.Json"
   $signAssemblies = $false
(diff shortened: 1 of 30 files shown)


iSynaptic:
On a project that needs this enhancement - any idea when it will be merged?

iSynaptic:
Agreed, I think the default is wrong, but this solution is agreeable. Perhaps 7.0 can make the default use `object.ReferenceEquals`.

JamesNK:
Added EqualityComparer to JsonSerializer https://github.com/JamesNK/Newtonsoft.Json/commit/3cc797c24071b35429674205712ed0a5df8cf7ad

PR #424, opened 2014-11-13
Added test and fix for race condition in contract resolving

asgerhallas:
We encountered an issue when initializing multiple serializers at the same time using shared contract resolver cache. 

The test written does not actually use shared cache, but imitates it to ensure that the test fails more often. If not written this way (but with sharedCache = true) the race condition would still appear but only not as often.

### DIFF
--- a/Src/Newtonsoft.Json/Serialization/JsonSerializerInternalReader.cs
+++ b/Src/Newtonsoft.Json/Serialization/JsonSerializerInternalReader.cs
@@ -27,6 +27,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Threading;
 #if !(NET35 || NET20 || PORTABLE40)
 using System.ComponentModel;
 using System.Dynamic;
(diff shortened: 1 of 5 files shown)


Src/Newtonsoft.Json/Serialization/JsonSerializerInternalReader.cs
```diff
@@ -1913,9 +1914,20 @@ private object PopulateObject(object newObject, JsonReader reader, JsonObjectCon
...
+            Dictionary<JsonProperty, PropertyPresence> propertiesPresence;
+            if (contract.HasRequiredOrDefaultValueProperties || HasFlag(Serializer._defaultValueHandling, DefaultValueHandling.Populate))
+            {
+                propertiesPresence = contract.Properties.ToDictionary(m => m, m => PropertyPresence.None);
+            }
+            else
+            {
+                if (contract.CreatedType.Name == "Root")
+                {
+                    Console.WriteLine("HERE on " + Thread.CurrentThread.ManagedThreadId);
```
ericnewton76: Why Console.WriteLine here?  Is there any logging or can you use builtin System.Diagnostics.Trace instead?

asgerhallas: I honestly can't remember anything about this (two years old) PR, sorry :) 
I guess the Console.Write is a left over from some debugging.

PR #435, opened 2014-12-03
Snake Case Contract resolvers

RaphHaddad:
Creating Snake case contract resolvers to both forms of snake case.

1. snake_case1 or
2. snake_case_2


dschniepp:
:+1:

JamesNK:
This can be a separate package or be something that a user does.

RaphHaddad:
@JamesNK is right.

FYI @dschniepp https://github.com/RaphHaddad/Newtonsoft.Json.Serialization.ContractResolverExtentions

PR #440, opened 2014-12-10
Adding support to use XML schemas for XML to JSON conversion

microsoftazure:
As we know xml only contains data it doesn't have data type information. While converting from XML to JSON we need to identify data type of each xml node, So We have modified the library to take a xml schema and read the data type from the schema. 
There exist a similar problem to identify a json array of single element. the current version of this library rely on a "ns:Array" attribute to identify single element array in json, but we can not expect a xml coming from any arbitrary system to have this kind of attribute present. this problem can also we solved by using xml schemas.


slabarque:
Any plans on adding this functionality? I also need to parse xml over which I do not have control, but I do have an xsd schema.

PR #443, opened 2014-12-11
A way to simplify customized serialization usage

qaqz111:
A **static Dictionary\<Type, JsonConverter[]\>** was added to JsonConvert class, which the "final" **SerializeObjectInternal** and **DeserializeObject** method will automaticlly get the right **JsonConverter[]** from, if any.

So, by registering a **custom JsonConverter[]** in advance, the serialization/deserialization code need **not** to pass **JsonConverter[]** everytime any more.

But, if **JsonConverter[]** was passed explicitly, the passed **JsonConverter[]** will override the registered ones. So that the behavior of existing code will not change.

This idea comes from MongoDB serialization mechanism, which is very convenient.


qaqz111:
I see the response on codeplex, and respect the author's decision.

PR #457, opened 2014-12-31
Update JToken.cs (JObject.ToString throws OutofMemory exception)

fuyifan:
JObject.ToString throws OutofMemory exception, when JToken uses StringWriter to generate string and this string is quite large. If the StringWriter.length is larger than 120MB, exception occurs very often. The reason is that StringBulder uses StringBuilder as its base data structure and StringBuilder requires a continuous memory. If the StringBuilder cannot get such a continuous memory it will throw OutofMemory exception.


JamesNK:
Use WriteTo instead of ToString for large objects.

PR #474, opened 2015-01-28
Improve Self referencing loop exceptions on properties by including a recommendation

b9chris:
The solution to a self-referencing loop on most Properties is to add an ignore reference loop attribute, so the Exception may as well be there to tell novices how to most easily resolve it. Most devs running into this for the first time are likely to be surprised at the depth of Json.Net in having ways of handling this, and it may lead them to explore JsonProperty(ReferenceLoopHandling more deeply.


ericnewton76:
I wholeheartedly endorse this, I ran into this problem before and was scratching my head as to what could've been circular.

And adding more information is always helpful.

PR #485, opened 2015-02-13
Added the Xamarin platforms to the NuGet

mattleibow:
Added the 3 Xamarin platforms to the NuGet:

 - Xamarin.iOS (Unified)
 - Xamarin.Android
 - MonoTouch (Xamarin.iOS Classic)


JamesNK:
Have you tested that both these builds work on iOS and Android? For example Newtonsoft.Json.Portable uses dynamic which traditionally has broken in iOS.

JamesNK:
No reply, no merge.

PR #487, opened 2015-02-13
added support for multiple deserialization aliases

BrannonKing:
I was looking for ways to make it easier to rename properties. I don't like having to keep both the old and new properties on my objects. For that reason, I added some code to allow me to specify multiple property aliases with a new property on the `JsonPropertyAttribute`. It required me to significantly change the `JsonPropertyCollection` as well. I could not see any way to accomplish what I wanted using hooks in the existing system.


JamesNK:
This is too disruptive and I think will be rarely used.

dgstpierre:
This feature is exactly what I need.  I need to support a long and short name for each property.  Thus two names, not just one.  Could you recommend a technique outside of staticly adding multiple properties to my view models?  Thanks

PR #489, opened 2015-02-19
Added support for \v char with tests

ptrstpp950:
According to https://www.json.com/json-object#object-with-strings valid json can contains '\v' char. It was discussed on SO on link: http://stackoverflow.com/questions/19176024/how-to-escape-special-characters-in-building-a-json-string

That is why I added support for this char :)

I hope You will agree with me


JamesNK:
I don't know why json.com includes it because it is not part of the spec: http://json.org/

PR #507, opened 2015-03-12
Adding Support for Writing String values from Streams

elarkin:
I have a need to write very large JSON strings (100MB+), and storing the entire string in memory before writing it to the JSON output stream is causing my agents to go OOM.

This change allows me to stream into a large string.

NOTE:

* In order to keep the new Streaming method in line with the String method, I made them quickly converge to the same code path. This causes writing String values to become a bit slower.
* This change extends the core JsonWriter interface. If you don't want to do this, an alternate implementation can be found here: https://github.com/elarkin/Newtonsoft.Json/tree/no-new-interface-method


JamesNK:
I think this is too niche. You might be able to get the same result using multiple calls to WriteRaw.

Issue #512, opened 2015-03-17
F# discriminated union deserialization bug


Horusiath:
I've discovered that default DiscriminatedUnionConverter fails when trying to deserialize a F# discriminated union instance with settings type name handling turned to `TypeNameHandling.All` (only with this configuration).

Exception message:
> An unhandled exception of type 'Newtonsoft.Json.JsonSerializationException' occurred in Newtonsoft.Json.dll
> Additional information: Union fields must been an array. Path 'Fields', line 1, position 28.

Test code used to recreate an error:
```fsharp
let settings = JsonSerializerSettings()
settings.TypeNameHandling <- TypeNameHandling.All

let serialize o =
    let json = JsonConvert.SerializeObject(o, Formatting.None, settings)
    Encoding.Default.GetBytes json

let deserialize bytes t =
    let json = Encoding.Default.GetString bytes
    JsonConvert.DeserializeObject (json, t, settings)

type TestUnion =
    | CaseOne of string
    | CaseTwo of int

let a = CaseOne "hello"

let abin = serialize a
let ades = deserialize abin (a.GetType())
let a2 = ades :?> TestUnion
```

Serialized union JSON: `{"Case":"CaseOne","Fields":{"$type":"System.Object[], mscorlib","$values":["hello"]}}`

I've also tracked down source of an error in [Newtonsoft.Json source code](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Converters/DiscriminatedUnionConverter.cs#L178).

JamesNK:
Fixed https://github.com/JamesNK/Newtonsoft.Json/commit/a3278ccd6a7ac88c3c5ae85d46a3ec46a6f438ec#diff-cb3fb9511e571c94735be121f80b6fe5R133

rogeralsing:
This only works if the type is known on the deserializing side.
if the discriminated union is stored in an object array, it still fails.

```fsharp
open Newtonsoft.Json
open System.Text;


let settings = JsonSerializerSettings()
settings.TypeNameHandling <- TypeNameHandling.All

let serialize o =
    let json = JsonConvert.SerializeObject(o, Formatting.None, settings)
    Encoding.Default.GetBytes json

let deserialize bytes t =
    let json = Encoding.Default.GetString bytes
    JsonConvert.DeserializeObject (json, t, settings)

type TestUnion =
    | CaseOne of string
    | CaseTwo of int

[...]

PR #513, opened 2015-03-17
DiscriminatedUnionConverter can deserialize fields from object

Horusiath:
This pull solves the problem described in issue #512. It gives a `DiscriminatedUnionConverter` ability to deserialize union fields from JSON serialized in `{ Fields: { $type: ..., $values: [] } }` notation (which is an array JSON format when serialization setting `TypeNameHandling` is set to `All`).

Changes have been reflected in test cases (one new added to confirm proper deserialization, one removed because it's not an error anymore).

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Converters/DiscriminatedUnionConverterTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Converters/DiscriminatedUnionConverterTests.cs
@@ -172,6 +172,22 @@ namespace Newtonsoft.Json.Tests.Converters
             Assert.AreEqual(10.0, r.width);
         }
 
+        [Test]
+        public void DeserializeUnionWithFieldsWithTypeNameHandlingAll()
+        {
+            JsonSerializerSettings settings = new JsonSerializerSettings
+            {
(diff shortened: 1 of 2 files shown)


JamesNK:
I think I'd prefer to fix this from the other side and not output $type and friends in a DU when type name handling and fields is enabled.

Horusiath:
I was thinking about this too. But how does it concern backward compatibility?

JamesNK:
Well it is broken now so it's not like it can become double broken for anyone.

JamesNK:
Fixed https://github.com/JamesNK/Newtonsoft.Json/commit/a3278ccd6a7ac88c3c5ae85d46a3ec46a6f438ec#diff-cb3fb9511e571c94735be121f80b6fe5R133

Horusiath:
@JamesNK This issue is necessary for a proper work of the [Akka.NET](http://getakka.net/) F# API plugin, which uses Newtonsoft.Json pretty extensively. Therefore my question is: can your fix code be included inside Akka.NET source code until it will be published with official Newtonsoft.Json release?

PR #521, opened 2015-03-26
DataTableConverter is using Serializer for cell data.

pischke:
Hello.
This is my first pull request.

When you want to deserialize JSON to a typed DataSet, which uses custom types (DataColumn.DataType) there was no possibility to use own JsonConverter.

This change make this possible, because of using the serializer for converting JSON value to cell value.


pischke:
Sorry, why build failed?
CSC : error CS1548: Cryptographic failure while signing assembly 'c:\projects\newtonsoft-json\Src\Newtonsoft.Json\obj\Release\Net45\Newtonsoft.Json.dll' -- 'Error signing assembly -- The parameter is incorrect. '

I haven't changed anything there, or?

PR #526, opened 2015-04-02
Fix for "inherited" interfaces

ravensorb:
It appears that the current GetProperties methods in the ReflectionUtils class does not handle the scenario were an interface is derived from 1 or more other interfaces.  This is a fix (and a unit test) for that issue


JamesNK:
What fails without this change? The test you modified still passes when the changes to ReflectionUtils are removed...

ravensorb:
Sorry, I checked in the wrong Unit test.  Please take a look now.

JamesNK:
Done

ravensorb:
Thanks!  Any thought on when you might be pushing out an updated beta?

JamesNK:
I did last night.

ravensorb:
Excellent -- thanks!

PR #533, opened 2015-04-08
Document the meaning of "default settings"

VictorNicollet:
The term "default settings" did not obviously reference the `JsonConvert.DefaultSettings` global variable (especially to those who were not aware of its existence), and so the actual difference between `Create()` and `CreateDefault()` was unclear. By referencing it explicitly, the code, and therefore its generated documentation, will be clearer.


VictorNicollet:
CI appears to have failed on an unrelated test. Is it possible for a BSON test to break by adding a few comments line to JSON serialization code ?

nitrocaster:
@VictorNicollet It fails even on clean master checkout. Unrelated to your commit.

JamesNK:
Done

PR #535, opened 2015-04-08
Add support for reading json from stream without buffering

nitrocaster:
Consider a stream that is being read by a two different entities. Let's say, stream contains a small json 'header' object followed by a large json 'data' object. We want the first entity to read header (that contains data size) and the second entity to read the data. Current `JsonReader` implementation doesn't allow to implement this task, because it caches data from stream, consuming beginning of 'data' object and so making impossible for the second entity to read that object from the stream.

My solution introduces `JsonReader.Unbuffered` property (false by default) that allows to disable caching so `JsonReader` will read as many chars as it actually needs.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/JsonTextReaderTest.cs
+++ b/Src/Newtonsoft.Json.Tests/JsonTextReaderTest.cs
@@ -3358,7 +3358,7 @@ null//comment
             if (disposing)
             {
                 if (own)
-                    r.Dispose();
+                    r.Close();
             }
             disposed = true;
         }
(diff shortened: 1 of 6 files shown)


JamesNK:
If people don't want to buffer information they can read all the content from the underlying textreader themselves and then pass a StringReader to json.net

If you want to read multiple objects from a stream then set SupportMultipleContent to true.

nitrocaster:
This cannot be done because length of the string to be read from the stream is unknown. If I would read it by myself, I would need to write my own json parser, but that's the purpose of Json.NET.

PR #538, opened 2015-04-13
Mixed assembly generics fix

tystol:


JamesNK:
I think this is too edge case to include by default.

PR #542, opened 2015-04-17
JSONPath query for fields in the root document seem to not work.

jmorrill:
Given a JSON document of:
{
  "EventTypeID": 5,
  "DataTimeStamp": "2015-03-17T18:44:08.387Z"
}

A JSONPath of $[?(@.EventTypeID > 1)] would never return a match.  This is a minor change to the implementation but it has not been well tested.


JamesNK:
Is this valid? My impression from reading the spec is queries like that are only for returning items in an array, not returning properties from an object.

Does that work in other JSONPath libraries?

JamesNK:
It doesn't work here: http://www.jsonquerytool.com/

Issue #115, opened 2013-08-08
Json Schema Specification (Version 4) support


gsferreira:
Hi,

When do you expect to support the Json Schema Specification (Version 4)?

br15:
Hi,

I've downloaded the code and built solution just fine.  However I need to integrate your code with a .Net 4.0 solution which I can't (for political reasons) upgrade to .Net 4.5 which is the version you've used with your solution.

Any chance you could create a .Net 4.0 version with an associated NuGet  package for Newtonsoft.Json 7.0.1-beta1.

Cheers

JamesNK:
I'll look at adding .NET 4. Probably shouldn't be an issue.

FYI there is a limit on the number of schema validations per hour without a license. I should have a website ready for buying a license in a couple of weeks but I wouldn't recommend using it in a live environment until then.

br15:
Hi James,

Thanks for the quick response.  I'll definitely be buying a license.
What's the current limit set to for validation per hour?
Could you please notify me as soon as the .Net 4.0 support is available.

Cheers
Keith.

Great work by the way.

JamesNK:
1000 an hour, resetting each time you run an application or rebuild a website. Lots for development purposes.

I've added .NET 4.0 support and uploaded a beta release to myget here: https://www.myget.org/F/json-net/api/v2

It uses a new feature I've added to json.net that hasn't been released yet so that will be updated to a beta version as well.

br15:
That was quick. Marvellous!

 

Cheers 

Keith

br15:
Hi James,

The link to the .Net 4 code doesn't seem to work. I just get this in the browser:

This XML file does not appear to have any style information associated with it. The document tree is shown below.
<service xmlns="http://www.w3.org/2007/app" xmlns:atom="http://www.w3.org/2005/Atom" xml:base="https://www.myget.org/F/json-net/api/v2/">
<workspace>
<atom:title>Default</atom:title>
<collection href="Packages">
<atom:title>Packages</atom:title>
</collection>
<collection href="Screenshots">
<atom:title>Screenshots</atom:title>
</collection>
</workspace>
</service>

JamesNK:
It's a NuGet feed URL. You don't download from it using a browser, you need to add it as a package source to the NuGet Package Manager in Visual Studio.

http://docs.nuget.org/docs/start-here/managing-nuget-packages-using-the-dialog#Package_Sources

[...]

PR #545, opened 2015-04-20
Adding support for the required array from http://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-5.4.3

eoinmullan:
This change will support the Json Schema version 4 meaning of the "required" keyword, where required properties are specified in an array. The previous (version 3) Boolean functionality of "required" is still maintained.


JamesNK:
I'm not making anymore changes to Newtonsoft.Json.Schema in json.net - in the future I'm going to deprecate it.

https://github.com/JamesNK/Newtonsoft.Json/issues/115

PR #546, opened 2015-04-20
Constructor should respect DefaultValue for params set in Contract

Andrew-Hanlon:
If a default value and `Populate` handling flag is set for a constructor parameter in a contract's `CreateConstructorParameters` method, then the constructor parameters array should be populated with the default if no matching property is available.

This allows the constructor params to be populated with values at deserialization via the contract. It is a minor code change, with many potential use-cases. 
- All existing tests pass.
- Added test for default param value handling.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/ConstructorHandlingTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/ConstructorHandlingTests.cs
@@ -25,7 +25,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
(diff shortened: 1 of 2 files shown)


JamesNK:
Done, however I used a different way of doing it.

geniuszxy:
@JamesNK, your way ignores the default values of parameters, such as #820.

This is inconvenient, especially now that C# has introduced record and primary constructors, users have to do this when they want to specify default values:

```
public record class Foo([DefaultValue("This is what JSON.NET can see")] string Id = "This is a default ID")
{
}
```

PR #547, opened 2015-04-20
updated copyright in LICENSE

attritionorg:


JamesNK:
I'm not worrying about updating dates. The same license is in the header of every file and I don't want to update them every year.

Issue #548, opened 2015-04-21
JToken: Ability to rename property names at runtime


abibell:
# How do I rename a property name?
We have JObject objects, based on the chosen export format we want to change the property names to something else (based on a dictionary). I don't want to remove and add into a different location, the order needs to be preserved. This is not like https://github.com/JamesNK/Newtonsoft.Json/pull/487.

```c#
JObject o1 = JObject.Parse(@"{ 'FirstName': 'John', 'LastName': 'Smith', 'Enabled': false }");
//This should change it to { 'fn': 'John', 'LN': 'Smith', 'Enabled': false }
o1["FirstName"].Rename("fn");
o1["LastName"].Rename("LN");
```

We have lot of properties to change so serialisation and regular expressions is providing to be difficult as we have to deal with hierarchy. The solution listed here works, but I think we can do better if we had a rename functionality:
http://stackoverflow.com/questions/11679804/json-net-rename-properties

abibell:
See https://github.com/JamesNK/Newtonsoft.Json/pull/549 for solution

PR #549, opened 2015-04-21
Property names can now be renamed

abibell:
Property names can now be renamed using .Rename("newPropertyName") to resolve https://github.com/JamesNK/Newtonsoft.Json/issues/548
- Added unit tests to cover the functionality

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Linq/JTokenTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Linq/JTokenTests.cs
@@ -628,6 +628,35 @@ namespace Newtonsoft.Json.Tests.Linq
         }
 
         [Test]
+        public void Rename()
+        {
+            //Arrange
+            var map = new Dictionary<string, string>
+            {
+                {"FirstName", "fn"},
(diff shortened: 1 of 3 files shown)


JamesNK:
Sorry I don't like this. It doesn't make sense when the value is inside an array for example.

abibell:
So we can do it for non array objects. Any data should support CRUD
operations don't you think? At the moment we don't do update. It's a valid
feature from user point of view.

abibell:
Anyone who wishes to use this functionality can use the code below:
```c#
public static class NewtonsoftExtensions
{
    public static void Rename(this JToken token, string newName)
    {
        var parent = token.Parent;
        if (parent == null)
            throw new InvalidOperationException("The parent is missing.");
        var newToken = new JProperty(newName, token);
        parent.Replace(newToken);
    }
```

mikkoviitala:
:+1:

Maxime66410:
> 

IDK why i follow the problem #548 and i use your solution, but i get always _"Accessed JObject values with invalid key value: 0. Object property name expected."_

PR #551, opened 2015-04-23
Fix DeserializingErrorHandlingUsingEvent test when building NET20 or NET35

speps:
Just wanted to build NET20 platform and ended up with one test failing, here is very simple fix :)


speps:
I don't see what's wrong with AppVeyor, let me know if I need to fix anything.

JamesNK:
I think I've added your change in a way that works across all builds.

PR #553, opened 2015-04-28
Updated Nuget.exe to 2.8.5 to enable NuGet restore to work without askin...

andrewchaa:
The existing NuGet.exe was 2.0 version and it didn't allow NuGet restore, if the user didn't give consent by updating the option in Visual Studio. Often it's disabled by default and NuGet team fixed the issue. So NuGet 2.8 can restore the missing packages in the command line without user consent, which I think makes more sense.

So, just one file change, with the latest NuGet.exe 2.8.5


andrewchaa:
I think 'System.Collections.Immutable' 1.1.34-rc was temporarily unavailable. It restores ok now, but I can't run the build again.

JamesNK:
Can you point me towards where you downloaded the file? I don't want to accept exes in pull requests (viruses, etc)

andrewchaa:
There's a direct download link on http://docs.nuget.org/consume/installing-nuget. The link is https://nuget.org/nuget.exe

PR #575, opened 2015-05-21
Emit trace warning when unable to deserialize non-writable property

lordjeb:
When deserializing json to an object and a member property is found that matches the json property, if the property is not writable, we should emit a warning to the trace writer. This will assist developers in tracking down a deserialization problem whose cause may not be obvious.


lordjeb:
Appears to have failed in building for some other reason than my changes. The build log doesn't show anything more than "Build Started".

warnat:
maybe because "private set;"  is not available on old clr's (1.0/1.1/2.0) ?

lordjeb:
Thank you! I could not figure out why the failure when there was NO build output. I've put the test case using the private setter inside the ifdef so it only applies to CLR versions where it should.

JamesNK:
Fixing with https://github.com/JamesNK/Newtonsoft.Json/pull/1871

PR #589, opened 2015-06-15
Renamed EqualityComparer property to ReferenceLoopEqualityComparer

iSynaptic:
Minor tweak, but renaming it, I feel, better communicates how the serializer is going to use the EqualityComparer. Thoughts?


JamesNK:
I'm ok with the current name. It might be used to compare objects for other features in the future.

PR #592, opened 2015-06-17
Revert from expressions to reflection in Portable40

ddunkin:
The change to using expressions for reflection in portable builds (57ccaf0) requires the use of dynamic code generation. This adds a size overhead of ~2MB per architecture to Xamarin.iOS apps because it requires Mono.Dynamic.Interpreter to be included.

This PR reverts the change to expressions for the Portable40 build. Xamarin projects can use either the Portable or Portable40 assemblies and thereby choose whether to optimize for speed or size.


ddunkin:
I'm going to take a different approach.

PR #601, opened 2015-06-24
Support case-sensitive and type-conversions-rules JsonSerializerSettings

inadler:
Sometimes, while deserializing an object, you wish to be more strict about the type of the object's parameters.

There are two new Enums to the JsonSerializerSettings:
```csharp
// specifies whether the deserializer attempts to convert basic types
public enum MemberTypeConversionHandling
{
    AllowTypeConversions = 0,// Allows conversion between strings, integers, floats and booleans
    ProhibitTypeConversions = 1// No conversions allowed
}

// specifies the deserializer what kind of string comparison options to use when looking up a class member by name
public enum MemberNameMatchHandling
{
    CaseInsensitive = 0, // Case-insensitive lookup
    CaseSensitive = 1 // Case-sensitive lookup
}
```
For example:
```csharp
JsonConvert.DeserializeObject<int>("5"); // default behaviour - result = 5;

var strictSettings = new JsonSerializerSettings {MemberTypeConversionHandling = MemberTypeConversionHandling.ProhibitTypeConversions};
JsonConvert.DeserializeObject<int>("\"5\"", strictSettings); // throws: Error converting value "5" to type 'System.Int32'.


class TestObject  { 
   public string str; 
}
JsonConvert.DeserializeObject<TestObject  >(@"{""Str"": ""aaa""}"); // default behaviour - str = "aaa";

var strictSettings = new JsonSerializerSettings { MemberNameMatchHandling = MemberNameMatchHandling.CaseSensitive };
JsonConvert.DeserializeObject<TestObject>(@"{""Str"": ""aaa""}", strictSettings); // str = null;
```


ivi-hamiti:
Is there any news if this merge request will be approved it would be very nice to have this feature

zbourdeau:
Any updates on this pull? This would be a great feature to have.

ffje (approved):

PR #604, opened 2015-06-28
Use "Indented" for DefaultFormatting when debugging.

kb3eua:
DefaultFormatting should remain "None" for RELEASE builds, but it's easier to read "Indented" JSON for the purposes of DEBUG builds.


JamesNK:
I would prefer they were the same for unit test consistency.

Also remember that the assembly on NuGet is always RELEASE so the only people this will ever effect will be those changing Json.NET.

Issue #618, opened 2015-07-27
Specify dependencies for .NET Core


davidfowl:
JSON.NET currently has PCL folder that targets `portable-net45+wp80+win8+wpa81+dnxcore50` this works for older PCL projects and .NET Framework based projects because msbuild adds all of the facades to the compiler target. Moving forward, that's no longer the case. Instead packages are expected to specify what dependencies they use in the nuspec.

Here' an example of System.Collections.nuspec:

```xml
<?xml version="1.0"?>
<package xmlns="http://schemas.microsoft.com/packaging/2013/01/nuspec.xsd">
  <metadata minClientVersion="3.0">
    <id>System.Collections</id>
    <version>4.0.10-beta-23110</version>
    <title>System.Collections</title>
    <authors>Microsoft</authors>
    <owners>microsoft,dotnetframework</owners>
    <licenseUrl>http://go.microsoft.com/fwlink/?LinkId=329770</licenseUrl>
    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
    <requireLicenseAcceptance>true</requireLicenseAcceptance>
    <description>Provides classes that define generic collections, which allow developers to create strongly typed collections that provide better type safety and performance than non-generic strongly typed collections.

Commonly Used Types:
System.Collections.Generic.List&lt;T&gt;
System.Collections.Generic.Dictionary&lt;TKey, TValue&gt;
System.Collections.Generic.Queue&lt;T&gt;
System.Collections.Generic.Stack&lt;T&gt;
System.Collections.Generic.HashSet&lt;T&gt;
System.Collections.Generic.LinkedList&lt;T&gt;
System.Collections.Generic.EqualityComparer&lt;T&gt;
System.Collections.Generic.Comparer&lt;T&gt;
System.Collections.Generic.SortedDictionary&lt;TKey, TValue&gt;</description>
    <copyright>© Microsoft Corporation.  All rights reserved.</copyright>
    <dependencies>
      <group targetFramework="MonoAndroid1.0" />
      <group targetFramework="MonoTouch1.0" />
      <group targetFramework=".NETFramework4.6" />
      <group targetFramework="Xamarin.iOS1.0" />
      <group targetFramework="Xamarin.Mac2.0" />
      <group targetFramework=".NETPlatform5.0">
        <dependency id="System.Runtime" version="4.0.0" />
      </group>
      <group targetFramework=".NETCore5.0">
[...]

PR #620, opened 2015-07-30
Add dependencies for .NET Core

onovotny:
This PR addresses #618. The build process for the Portable project will now update the nuspec, if needed, to specify the proper dependencies.

It also removes dnxcore50 from the portable- directory and puts the same PCL lib into \lib\dotnet


Build/Newtonsoft.Json.nuspec
```diff
@@ -11,5 +11,34 @@
     <iconUrl>http://www.newtonsoft.com/content/images/nugeticon.png</iconUrl>
     <licenseUrl>https://raw.github.com/JamesNK/Newtonsoft.Json/master/LICENSE.md</licenseUrl>
     <tags>json</tags>
+    <dependencies>      
+      <group targetFramework="net2" />
+      <group targetFramework="sl5" />
+      <group targetFramework="portable-net40+sl5+wp80+win8+wpa81" />
```
JamesNK: What's the reason for these additional groups?

onovotny: I will double & triple check this later today, but this is to prevent an error @bradwilson and I saw while adding dotnet support to xUnit yesterday.

It appears that the resolution logic for the dotnet dependencies section is independent from the resolution of the \lib section.

What we saw is that an older platform got the \lib\foo version of the library but the dotnet dependencies section -- unless we added a blank dependency group.

What I want to do later this afternoon is experiment with a number of permutations to see how this all applies. I'm not 100% sure that the blank portable-net4+sl5-... group is needed but I haven't tried what happens yet if I have a PCL of that variety and I try adding a package with the newer dotnet contracts.

The underlying issue is that if the contracts/dependencies in the dotnet section are newer and thus incompatible with the target project, then the package will fail to install using NuGet v3 (v2 ignores the dotnet TFM entirely). The tests involve creating various project types in VS 2015 and adding the package to ensure it works correctly for all.

I'll report back later tonight (NY time) with my findings. I do wish this was better documented, but I'll document my findings.

onovotny: I tested this further and do not believe those groups are needed for sl5 or net2. It appears that the `dotnet` TFM dependencies are only evaluated for .NET 4.5+ projects (or more accurately, System.Runtime) based projects.

The case where the empty groups are needed is if you target .NET 4.6+ stuff (like the newer .NET Core packages that use System.Runtime 4.0.20 or higher). Then you need to block those dependencies from trying to be used by earlier platforms. Make sense?

Issue #637, opened 2015-08-28
JsonArrayContract Ignores Private Constructor With JsonConstructorAttribute


Tydude4Christ:
Consider the following
```
public class Foo : IEnumerable<int>
{
	private List<int> _bars;

	public Foo()
	{
		_bars = new List<int>();
	}

	[JsonConstructor]
	private Foo(IEnumerable<int> bars)
	{
		_bars = new List<int>(bars);
	}

	public void Add(int bar)
	{
		_bars.Add(bar);
	}

	public IEnumerator<int> GetEnumerator() => _bars.GetEnumerator();

	IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
}

[Test]
public void TestJsonConstructor()
{
	var foo = new Foo();
	foo.Add(0);
	foo.Add(1);
	foo.Add(2);
	var json = JsonConvert.SerializeObject(foo);
	var newFoo = JsonConvert.DeserializeObject<Foo>(json);
}
```

The test will fail on DeserializeObject with the following JsonSerializationException.
Message: Cannot create and populate list type Foo. Path '', line 1, position 1.
StackTrace:
  at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateNewList(JsonReader reader, JsonArrayContract contract, Boolean& createdFromNonDefaultCreator)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateList(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, Object existingValue, String id)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)
   at Newtonsoft.Json.JsonConvert.DeserializeObject(String value, Type type, JsonSerializerSettings settings)

Digging in the code I discovered JsonArrayContract to be the culprit as it only considers public constructors instead of also considering constructors with the JsonConstructorAttribute applied.

JamesNK:
[...]

PR #640, opened 2015-08-31
Potential Fix For #637

Tydude4Christ:
Potential fix fo #637. I'm unable to run the unit tests as NuGet in Visual Studio 2015 is giving me the following error

Dependency '"Newtonsoft.Json": ""' has invalid version specification.


JamesNK:
I added support here - https://github.com/JamesNK/Newtonsoft.Json/commit/ad68ea8780b2cbb63908bff68d4d6464e2967088

PR #648, opened 2015-09-12
Use yield in GetFieldsAndProperties

jamesqo:
By using yield instead of `List<T>`, we avoid the overhead of newing several new Lists and don't enumerate through the collection when we don't have to.

Associated caller has been fixed in `DefaultContractResolver`.


JamesNK:
I don't want to merge this. I know it is more efficient but GetFieldsAndProperties is only called when an object is serialized for the first time so its performance isn't important. And yield produces a lot more IL for the state machine and I don't want the DLL size to grow too large.

PR #655, opened 2015-09-19
Let json support "Dollar-Quoted String".

alanthinker:
Let json support "Dollar-Quoted String". 

Dollar-Quoted string is introduced by PostgreSQL. But it not about database. It's a very good idea to express string in source code.
It's has similar effect with xml's CDATA. But Dollar-Quoted string is more powerfull and easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ A dollar-quoted string that follows a keyword or identifier must be separated fr
       {
         "IntervalSecondsForCheck": "60",
         "Start": "2014-09-21T05:00:00+08:00",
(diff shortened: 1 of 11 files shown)


alanthinker:
More information:
https://github.com/alanthinker/Newtonsoft.Json/blob/master/README.md

JamesNK:
I don't want to add this. I don't think avoiding escaping double quotes offers much value and it isn't part of the JSON spec.

alanthinker:
If use json as a configuration file, it can let command and code block configuraton easier.

PR #657, opened 2015-09-19
- BSON files will now check for overflow when computing sizes.

KimikoMuffin:
In the [BsonBinaryWriter.CalculateSize()](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Bson/BsonBinaryWriter.cs#L238) method: the Object and Array types make no attempt to check for overflow. This _probably_ isn't going to be much of a problem, given the unlikelihood that someone will try to make a BSON file which is anywhere near 2 GB, but it's still potentially there. So here's my attempt to fix it.


JamesNK:
BSON has been split out into its own project

If it is still something you want then create a PR here https://github.com/JamesNK/Newtonsoft.Json.Bson

PR #673, opened 2015-10-05
Fixes deserializing with type name handling on assemblies loaded with Assembly.LoadFrom

Tydude4Christ:
The assemblyName should not be compared to the Assembly.FullName as Json.NET only serializes a partial name.


JamesNK:
What bug is this fixing? I don't add anything without unit tests.

Tydude4Christ:
Added issue for this #675. I'll happily add a unit test but I'm not sure you'd want it implemented.

Tydude4Christ:
Due to old code, I created a new branch and pull request for this at #712.

PR #681, opened 2015-10-09
Fixed an parsing error for mono 4.0

devgopher:


JamesNK:
I'm not doing a Mono specific build. Mono can fix itself.

Issue #675, opened 2015-10-05
Deserializing with type name handling on assemblies loaded with Assembly.LoadFrom


Tydude4Christ:
Consider this situation.

Library.dll
```
namespace LibraryNS
{
    public abstract class BaseClass
    {
    }

    public class DerivedClass : BaseClass
    {
    }
}
```

Program.exe
```
using System.Reflection;
using Newtonsoft.Json;

namespace ProgramNS
{
    class Program
    {
        static void Main()
        {
            var assembly = Assembly.LoadFrom("../../../Library/bin/Debug/Library.dll");
            var b = JsonConvert.DeserializeObject("{\"$type\":\"LibraryNS.DerivedClass, Library\"}", new JsonSerializerSettings { TypeNameHandling = TypeNameHandling.Auto });
        }
    }
}
```

DeserializeObject will fail with a JsonSerializationException with message of "Could not load assembly 'Library'." with a stacktrace of
   at Newtonsoft.Json.Serialization.DefaultSerializationBinder.GetTypeFromTypeNameKey(TypeNameKey typeNameKey) in C:\Users\brinkleyt\Documents\GitHubVisualStudio\Newtonsoft.Json\Src\Newtonsoft.Json\Serialization\DefaultSerializationBinder.cs:line 83

To fix this the partial name should be compared instead of FullName.

JamesNK:
I don't make changes without unit tests.

Tydude4Christ:
I'd like to create a unit test for this but due to the nature of the bug it requires a separate dll that is loaded using the Assembly.LoadFrom method and it can't be a system dll. I could add a project for this simple dll to the main solution, create a separate solution for the generation of the dll, or simply include a dll in the repository. Which would be most preferable?

PR #712, opened 2015-11-12
Fixes deserializing with type name handling on assemblies loaded with Assembly.LoadFrom

Tydude4Christ:
Fixes Issue #675


Tydude4Christ:
Link to old pull request, #673.

JamesNK:
I don't make changes without unit tests.

PR #737, opened 2015-12-11
* Update JContainer.ValidateToken to remove an old property with the …

TheBrooster:
…same key as the incoming property.

   - Insertion will continue as normal; resulting in the newer property overwriting the older one.
* Handle index going out of range as remove the old property will decrease the size of the Child list.


TheBrooster:
Not sure why the auto build is failing, does someone fancy having a look, please?

JamesNK:
Hasn't this already been done in the latest source?

PR #741, opened 2015-12-15
Use GetValueOrDefault() over Value when Nullable<T> known non-null

JonHanna:
`GetValueOrDefault()` is an easily-inlined access to a field as this will hold the correct value for both `HasValue` and `!HasValue` cases.

The `Value` property in comparison contains a branch on a check that the `Nullable<T>` has its internal `hasValue` flag to decide whether to throw `InvalidOperationException` or not.

Hence when a nullable type is known to have a value (not be `null`) and one wants the value, rather than to possibly trigger the exception-throw, `GetValueOrDefault()` has a slight performance benefit over `Value`.
 
Accordingly, replace calls to the `Value` property with calls to `GetValueOrDefault()` when previous checks or assignments guarantee the value is not `null`.

### DIFF
--- a/Src/Newtonsoft.Json/Bson/BsonBinaryWriter.cs
+++ b/Src/Newtonsoft.Json/Bson/BsonBinaryWriter.cs
@@ -198,7 +198,7 @@ namespace Newtonsoft.Json.Bson
         {
             if (calculatedlengthPrefix != null)
             {
-                _writer.Write(calculatedlengthPrefix.Value);
+                _writer.Write(calculatedlengthPrefix.GetValueOrDefault());
             }
 
             WriteUtf8Bytes(s, byteCount);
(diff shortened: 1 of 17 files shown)


JamesNK:
What about adding an extension method to Nullable<T> called something like GetKnownNonNullValue that internally just calls GetValueOrDefault. I'm assuming a one line method would be inlined. That makes it obvious what the intent is and still offers the perf gain.

JonHanna:
Not sure that would really make it clearer; that `GetValueOrDefault()` will return the value when the struct isn't null is known to all, and that it's faster than `Value` is known to some (and common within the framework itself), but a custom extension is known to nobody unless they look at it. Even then it'll seem strange unless a comment explains it.

If you consider this optimisation opaque, I think I'd rather just add a comment to each of the affected lines, and do the explaining right at the spot in question.

JamesNK:
Ok, I'll leave it as is. Thanks!

JonHanna:
Well, I'll happily commit a bunch of such comments if you want them.

JamesNK:
No it's ok, I'm happy with it as it.

Issue #746, opened 2015-12-18
JsonReader has protected Dispose(bool disposing) but doesn't call GC.SuppressFinalize()


JonHanna:
`JsonReader` follows the "basic disposing pattern" of having a protected `Dispose(bool)` for derived classes to call, calling `Dispose(false)` from its own `IDisposable.Dispose()` and allowing derived classes to call `Dispose(true)` in a finaliser if they implement one.

However, it doesn't suppress the finaliser in that `IDisposable.Dispose()`, so the finaliser will still be called if it has been disposed, at best delaying garbage collection and at worse leading to an error happening during that finalisation.

Reproduction:

```C#
public class UnmanagedResourceFakingJsonReader : JsonReader
{
    public static int DisposalCalls;

    public static void CreateAndDispose()
    {
        ((IDisposable)new UnmanagedResourceFakingJsonReader()).Dispose();
    }

    public UnmanagedResourceFakingJsonReader()
    {
        DisposalCalls = 0;
    }

    protected override void Dispose(bool disposing)
    {
        base.Dispose(disposing);
        ++DisposalCalls;
    }

    ~UnmanagedResourceFakingJsonReader()
    {
        Dispose(false);
    }

    /* Rest of required implementation */

}

[Test]
public void DisposeSupressesFinalization()
{
    UnmanagedResourceFakingJsonReader.CreateAndDispose();
    GC.Collect();
    GC.WaitForPendingFinalizers();
    Assert.AreEqual(1, UnmanagedResourceFakingJsonReader.DisposalCalls);
}
```
Result:

    Newtonsoft.Json.Tests.JsonTextReaderTest.DisposeSupressesFinalization:
      Expected: 1
      But was:  2

JamesNK:
Good catch. The same thing should happen in JsonWriter?

JonHanna:
Its equivalent method is private, which is presumably a mistake, but it's not clear whether it should be made protected in terms of the impact on existing code.

JonHanna:
Thinking a bit further, while I'm not a fan of this pattern (I prefer to just have a virtual Dispose and force derived classes to deal with any unmanaged resources through managed wrappers with their own finalisers), as is there's no good way to Dispose such a derived type. I'll submit a PR accordingly when I'm back at a computer (on a phone right now).

PR #750, opened 2015-12-19
Expose JsonWriter.Dispose(bool) to derived classes.

JonHanna:
Change to protected and virtual. Add GC.SuppressFinalize to IDisposable.Dispose().

See discussion on #746.

Also a micro-opt on JsonReader.Dispose(bool), changing a && to a & as the right-hand operand is fast to access and for most calls both will be true anyway, so removing the branch is a slight win on short-circuiting in that branch.


JonHanna:
Ah. @JamesNK got there ahead of me.

Issue #650, opened 2015-09-15
Async methods for JsonTextReader / JsonTextWriter


gretro:
Hello.

I wanted to know if there were any plans for Json.Net to provide async methods on the text reader and the text writer. Since those classes deal with streams, it could be good addition. I wanted to know if you would be doing it eventually soon, or if there were no plans.

I know you already made an attempt at creating those methods, but I wanted to know if you would try again without the Task.Run.

Thanks.

JamesNK:
No plans right now. It was be a significant amount of work and I'm not confident I have the knowledge to write perfect async code currently.

JonHanna:
https://github.com/JonHanna/Newtonsoft.Json/commit/0ec73f19773f6eaa4a810f5609f4c844ef7059bc is a quick start on this, pushing async through to the stream, and returning synchronously when that's not available (c.f. how `StringReader` and `StringWriter` return synchronously on their `*Async` methods) and mostly not trying to be too clever (refactoring can happen later).

I'd be happy to continue with this if it's desired.

JamesNK:
Thanks for looking at this. I've been putting it off.

Wow that is a lot of copy and paste. I know most it can't be helped due to the nature of async. Looking at that, I'm worried about how doing a lot of things twice and testing them twice will effect maintainability. Also so much duplicate code, and all the async state machines are going to grow the DLL hugely.

JonHanna:
That would be a primary focus of looking at this deeper. It's relatively easy to produce such a copy-paste implementation as this. The real work comes in with finding shared code and/or deciding to be synchronous with I/O

JonHanna:
In a way though, this assembly is iin the perfect storm position. Most are either very close to the I/O (just add a couple of async overloads) or ultimate consumers (async or not; pick one), but this assembly is nastily between the two. 4

AArnott:
+1 to async `JsonTextReader` (and possibly `JsonTextWriter`, although that's not impeding our progress so much). 
There are patterns that can be followed to avoid repetition of code to keep the test matrix small. I can explain and perhaps demonstrate those patterns if there's enough interest to respond with the update.

JonHanna:
Well, it may have been a year between my saying I'd tackle it and my submitting a PR, but PTAL.

erezvani1529:
Hi @JonHanna,

Would you please share a time-line (if it exists) as to when this change will be generally available on nuget as a stable package?

Thanks,
Elham

[...]

PR #751, opened 2015-12-19
Refactor repeated code

JonHanna:
These are a few cases of code that is repeated in several places that could be refactored into separate methods that I came across while determining how best to approach #650 (the less repeated code in the current version, the less duplicated code would be introduced by adding async versions).

I'll happily squash them down to one commit if you want, but they're probably easier to consider separately (and perhaps decide to accept some and reject some) as separate commits.

### DIFF
--- a/Src/Newtonsoft.Json/JsonReader.cs
+++ b/Src/Newtonsoft.Json/JsonReader.cs
@@ -457,67 +457,46 @@ namespace Newtonsoft.Json
         {
             _readType = ReadType.ReadAsDateTimeOffset;
 
-            JsonToken t;
+            JsonToken t = ReadToContentInternal();
 
-            do
+            switch (t)
             {
-                if (!ReadInternal())
-                {
-                    SetToken(JsonToken.None);
+                case JsonToken.None:
+                case JsonToken.Null:
+                case JsonToken.EndArray:
                     return null;
-                }
-                else
-                {
-                    t = TokenType;
-                }
-            } while (t == JsonToken.Comment);
+                case JsonToken.Date:
+                    if (Value is DateTime)
+                    {
+                        SetToken(JsonToken.Date, new DateTimeOffset((DateTime)Value), false);
+                    }
 
-            if (t == JsonToken.Date)
-            {
-                if (Value is DateTime)
-                {
-                    SetToken(JsonToken.Date, new DateTimeOffset((DateTime)Value), false);
-                }
+                    return (DateTimeOffset)Value;
+                case JsonToken.String:
+                    string s = (string)Value;
+                    if (string.IsNullOrEmpty(s))
+                    {
+                        SetToken(JsonToken.Null);
+                        return null;
+                    }
 
-                return (DateTimeOffset)Value;
+                    DateTimeOffset dt;
+                    if (DateTimeUtils.TryParseDateTimeOffset(s, _dateFormatString, Culture, out dt))
+                    {
+                        SetToken(JsonToken.Date, dt, false);
+                        return dt;
+                    }
+
+                    if (DateTimeOffset.TryParse(s, Culture, DateTimeStyles.RoundtripKind, out dt))
+                    {
+                        SetToken(JsonToken.Date, dt, false);
+                        return dt;
+                    }
+
+                    throw JsonReaderException.Create(this, "Could not convert string to DateTimeOffset: {0}.".FormatWith(CultureInfo.InvariantCulture, Value));
+                default:
+                    throw JsonReaderException.Create(this, "Error reading date. Unexpected token: {0}.".FormatWith(CultureInfo.InvariantCulture, t));
             }
-
-            if (t == JsonToken.Null)
-            {
-                return null;
-            }
-
-            if (t == JsonToken.String)
-            {
-                string s = (string)Value;
-                if (string.IsNullOrEmpty(s))
-                {
-                    SetToken(JsonToken.Null);
-                    return null;
-                }
-
-                DateTimeOffset dt;
-                if (DateTimeUtils.TryParseDateTimeOffset(s, _dateFormatString, Culture, out dt))
-                {
-                    SetToken(JsonToken.Date, dt, false);
-                    return dt;
-                }
-
-                if (DateTimeOffset.TryParse(s, Culture, DateTimeStyles.RoundtripKind, out dt))
-                {
-                    SetToken(JsonToken.Date, dt, false);
-                    return dt;
-                }
-
-                throw JsonReaderException.Create(this, "Could not convert string to DateTimeOffset: {0}.".FormatWith(CultureInfo.InvariantCulture, Value));
-            }
-
-            if (t == JsonToken.EndArray)
-            {
-                return null;
-            }
-
-            throw JsonReaderException.Create(this, "Error reading date. Unexpected token: {0}.".FormatWith(CultureInfo.InvariantCulture, t));
         }
 #endif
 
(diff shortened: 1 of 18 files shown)


JonHanna:
The one observable change introduced is that some exception messages are changed as `ReadAndAssert` consolidates a few similar methods. The exception type remains the same, and the meaning should remain understandable.

Src/Newtonsoft.Json/JsonReader.cs
```diff
@@ -1109,5 +987,50 @@ public virtual void Close()
             _tokenType = JsonToken.None;
             _value = null;
         }
+
+        internal void ReadAndAssert()
+        {
+            if (!Read())
+            {
+                throw JsonSerializationException.Create(this, "Unexpected end.");
```
JamesNK: Change the error message to "Unexpected end when reading JSON". I only had just "unexpected end" because I was lazy writing messages.

Src/Newtonsoft.Json/Converters/BinaryConverter.cs
```diff
@@ -152,7 +152,7 @@ private byte[] ReadByteArray(JsonReader reader)
         {
             List<byte> byteList = new List<byte>();
 
-            while (reader.Read())
+            while (reader.ReadAndSkipComments())
```
JamesNK: I'd prefer to keep Read() into JsonToken.Comment test in switch statements the way they are. It's slightly faster

Src/Newtonsoft.Json/JsonReader.cs
```diff
@@ -812,55 +706,39 @@ internal string ReadAsStringInternal()
...
-                }
-            } while (TokenType == JsonToken.Comment);
-
-            if (TokenType == JsonToken.Date)
-            {
-                return (DateTime)Value;
-            }
-
-            if (TokenType == JsonToken.Null)
+            switch(ReadAndSkipCommentsInternal())
```
JamesNK: Whitespace before open parenthesis.

Src/Newtonsoft.Json/JsonReader.cs
```diff
@@ -1109,5 +987,50 @@ public virtual void Close()
...
+                throw JsonSerializationException.Create(this, "Unexpected end.");
+            }
+        }
+
+        internal bool ReadAndSkipComments()
+        {
+            return Read() && SkipComments();
+        }
+
+        internal bool SkipComments()
```
JamesNK: XmlReader has a method called MoveToContent that fulfills the same purpose. I think SkipComments could be called ReadToContent.

Src/Newtonsoft.Json/JsonReader.cs
```diff
@@ -1109,5 +987,50 @@ public virtual void Close()
...
+                if (!Read())
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private JsonToken ReadAndSkipCommentsInternal()
```
JamesNK: ReadToContentInternal

Src/Newtonsoft.Json/Linq/JToken.cs
```diff
@@ -2113,60 +2113,47 @@ public static JToken ReadFrom(JsonReader reader, JsonLoadSettings settings)
 
             if (reader.TokenType == JsonToken.None)
             {
-                if (!reader.Read())
+                if (!(settings != null && settings.CommentHandling == CommentHandling.Ignore ? reader.ReadAndSkipComments() : reader.Read()))
```
JamesNK: Hmmm, I'm not a huge fan of calling methods that have a side effect inside a ternary statement, but I'm also not a fan of duplicating the throw.

Src/Newtonsoft.Json/Serialization/JsonSerializerInternalReader.cs
```diff
@@ -71,13 +71,8 @@ public void Populate(JsonReader reader, object target)
...
             JsonContract contract = Serializer._contractResolver.ResolveContract(objectType);
 
-            while (reader.TokenType == JsonToken.None || reader.TokenType == JsonToken.Comment)
-            {
-                if (!reader.Read())
-                {
-                    throw JsonSerializationException.Create(reader, "No JSON content found.");
-                }
-            }
+            if (!(reader.TokenType == JsonToken.None ? reader.ReadAndSkipComments() : reader.SkipComments()))
```
JamesNK: See previous ternary statement comment.

Also need braces around throw.

Src/Newtonsoft.Json/JsonReader.cs
JonHanna: Considering that analogy with `XmlReader`, `MoveToContent` would be a closer name, with `ReadToContent` being more like what is `ReadAndSkipComments` here.

They could be useful public methods, though to completely match expected behaviour, it would need to start with `if (TokenType == JsonToken.None && !Read()) return false;`. That would be a bit of a waste on the internal uses, but perhaps still worth it.

Src/Newtonsoft.Json/Linq/JToken.cs
JonHanna: Yeah, it was removing the duplication the throw that pushed it past the threshold for me, but also because I have a low threshold for these changes figuring it's better to discuss every such change I could see and reject some than second-guess too early.

I'm also inclined to give a bit more leeway when the side-effects of both are similar and when they're "consuming" side-effects.

Maybe line-splitting it would bring a bit more visual focus to the to the two expressions picked between, and slightly reduce the downside?

Src/Newtonsoft.Json/JsonReader.cs
JonHanna: On that basis, it may be as well to do so to the case JsonTextReader.ParseValue with the same message while I do so.

Src/Newtonsoft.Json/JsonReader.cs
```diff
@@ -548,81 +519,66 @@ internal byte[] ReadAsBytesInternal()
...
-                throw JsonReaderException.Create(this, "Unexpected end when reading bytes.");
-            }
+                    return (byte[])Value;
+                case JsonToken.StartArray:
+                    List<byte> buffer = new List<byte>();
 
-            if (t == JsonToken.EndArray)
-            {
-                return null;
+                    for(;;)
```
JamesNK: Any reason for a for over while (true)? I have always used a while loop.

JonHanna: Just that I've always used an empty `for`. I'll change it to match your style. I've a few other comments to make, but something came up just after that last push.

JonHanna:
Okay, the story so far…

The formatting mistakes are rectified, and `for(;;)` replaced with `while (true)`.

The error message is as requested, and also on `JsonTextReader.ParseValue()` for consistency.

Most of the `case JsonToken.Comment` blocks have been reverted. Those that are in response to calls to `JsonSerializerInternalReader.ReadForType()` remain, as that method already ate comments in one path, and doing so in the other can reduce calls into it, after which the case is no longer reachable.

You mention `XmlReader.MoveToContent()` in regard to what I'd called `SkipComments` and `ReadAndSkipComments` and suggested `ReadToContent()`. To follow the analogy fully then `ReadAndSkipComments` should become `ReadToContent()` while `SkipComments` should become `MoveToContent()` and call `Read()` if `TokenType == JsonToken.None`, and they should also be public.

I've done that, though I'm not sure about the making-them-public bit. Since this means that `MoveToContent()` has a check against `TokenType == JsonToken.None` that is known to not be necessary all the times it was called internally, these both work by calling into an internal `MoveToContentUnsafe()` that doesn't do that test, and which is the real replacement of `SkipComments` as it had been.

This also allows the new (checking) `MoveToContent` to replace one of the ternary cases you objected to.

The other ternary expression I haven't touched yet. It does remove the duplicate exception throw, and is the only way to do so I can think of without `goto` (I'm happier in using `goto` than most, but it's still worth avoiding). Anyway, your comment on it seemed conflicted, so I haven't touched it yet.

JamesNK:
I made a couple of minor changes but otherwise it looks good. Thanks!

PR #817, opened 2016-02-18
Updated StringEnumConverter to correctly handle integers when passed as Strings and Integers are not allowed

GeorgeVovos:
The problem was initially posted here:
http://stackoverflow.com/questions/35479669/stringenumconverter-invalid-int-values/35484440#35484440

StringEnumConverter works as expected when it tries to convert an int value like
{  "MyEnumProperty": 99999} 
but if I specify the value as string it ,
{  "MyEnumProperty": "99999"}
maybe correctly, it tries to parse it

It could have been smarter and if AllowIntegerValues is set to false could check if the value is an integer and throw an error.
I believe this have help users ,like me ,avoid strange bugs
This is a breaking change but since Integers are not valid identifiers for an enum I cannot think of a case where that affects anyone (You may can though)

I added a few lines that fix the issue and created 2 unit tests.
I seems I haven't broken anything.

Review it and if you find it useful ,and correct,accept the pull request

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Converters/StringEnumConverterTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Converters/StringEnumConverterTests.cs
@@ -413,6 +413,41 @@ namespace Newtonsoft.Json.Tests.Converters
             Assert.Fail();
         }
 
+        [Test]
+        public void DeserializeInvalidStringWhenNotAllowIntegers()
+        {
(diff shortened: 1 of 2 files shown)


JamesNK:
Hi

I fixed this for another issue. It needed to be done in a different place because EnumMember can be used to accept an integer string as the name.

https://github.com/JamesNK/Newtonsoft.Json/commit/6c272eac6ee71705fd813c90ce61b1a6d6ef2a35

Issue #841, opened 2016-03-10
Should be able to deserialize big number


challenh:
## Repro steps:
1. Create a .NET Framework 3.5 console application (not 4.0 or portable).
2. Add and run the below code:
```
        static void Main(string[] args)
        {
            string json = "{\"StandardPrice\":79228162514264337593543950335}"; // decimal.MaxValue
            using (JsonTextReader jsonTextReader = new JsonTextReader(new StringReader(json)))
            {
                jsonTextReader.Read();
                jsonTextReader.Read();
                jsonTextReader.Read(); // should read the big number
                var bigNum = jsonTextReader.Value;
            }
        }
```

## Actual result:
An unhandled exception of type 'Newtonsoft.Json.JsonReaderException' occurred in Newtonsoft.Json.dll
Additional information: JSON integer 79228162514264337593543950335 is too large or small for an Int64. Path 'StandardPrice', line 1, position 46.
   at Newtonsoft.Json.JsonTextReader.ParseNumber(ReadType readType)
   at Newtonsoft.Json.JsonTextReader.ParseValue()
   at Newtonsoft.Json.JsonTextReader.Read()
   at ConsoleApplication1.Program.Main(String[] args) in d:\users\....\ConsoleApplication1\Program.cs:line 16

## Expected result:
JsonTextReader should somehow successfully read the big number, or allow the user to customize the handling, instead of throwing exception.  (looks that some more processing is needed @ https://github.com/JamesNK/Newtonsoft.Json/blob/2491737e52fc0f3f3c36c0faa2a6b670c5c4ab54/Src/Newtonsoft.Json/JsonTextReader.cs#L1966)

JamesNK:
That's by design. BigInteger doesn't support .NET 3.5 or older.

challenh:
Well understood, just that there is need for parsing large numbers in all .net framework versions.

challenh:
Hi James, a PR has been created for this issue, could you see if that can be merged? thanks!

PR #846, opened 2016-03-15
JsonTextReader supports parsing large numeric value by using custom handler

challenh:
Should be able to deserialize big number in .net 3.5 #841.


JamesNK:
I don't want to add work arounds for older versions of .NET. If someone is using .NET 3.5 and they have very large numbers then they have the option to implement there own JsonReader.

PR #875, opened 2016-04-11
JValue.ToString() uses IFormattable if possible to produce InvariantCulture string.

### LINKED ISSUE #874: JValue.ToString() is not InvariantCulture
JValue.ToString() uses current thread culture instead of CultureInfo.InvariantCulture.
There is no way to specify culture for this call.
And documentation does not states anything regarding this call (unfortunately). Could be that this is intended behavior?

Here is unit test that shows this issue:
```
// Arrange
var json = @"{ ""Prop"": 5.6 }";
Thread.CurrentThread.CurrentCulture = new CultureInfo("en-Gb", true)
{
	NumberFormat = new NumberFormatInfo
	{
		NumberDecimalSeparator = ">>>"
	}
};

// SUT
var obj = (JToken)JsonConvert.DeserializeObject(json);

// Act
var result = obj["Prop"].ToString();

// Assert
result.Should().Be("5.6"); // Expected "5.6" but got "5>>>6"
```

I will try to look into sources & create a unit test for this case.

antonovicha:
This is fixes #874.

JValue will try to use IFormattable to produce InvariantCulture string.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Linq/JValueTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Linq/JValueTests.cs
@@ -45,6 +45,7 @@ using NUnit.Framework;
 #endif
 using Newtonsoft.Json.Linq;
 using System.Globalization;
+using System.Threading;
 #if NET20
 using Newtonsoft.Json.Utilities.LinqBridge;
 #else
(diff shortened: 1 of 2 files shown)


antonovicha:
Since issue #874 is not a real issue this pull request make no sense. Please refer to comment: https://github.com/JamesNK/Newtonsoft.Json/issues/874#issuecomment-208751470

PR #884, opened 2016-04-19
JsonTextReader auto supports decimal precision (28 significant figure…

challenh:
JsonTextReader auto supports decimal precision (28 significant figures) &
double range (10e324).


JamesNK:
Thanks for the PR but this isn't a feature that I want to add.

PR #890, opened 2016-04-22
Enables XmlElement object type support in XmlNodeConverter

jack4it:
Hey James, any particular reason why not supporting `XmlElement` **deserialization** in `XmlNodeConverter`? In one of the projects I'm working on we do have a type which has an XmlElement type field. It can be serialized perfectly by Json.NET, however, always fails during de-serialization with the error "XmlNodeConverter only supports deserializing XmlDocuments".

I'm sending this PR in advance to see if you are fine with it. I have tested it in my own project and it works. The build also passes entirely.

Thanks,
-Jack


jack4it:
Hmm, stuck at this state forever...

JamesNK:
Done (along with XmlNode, XObject, XNode & XContainer)

jack4it:
Awesome, thanks!

Issue #838, opened 2016-03-09
[8.0.2] Failure to deserialize ulong on 1.0.0-rc1-update1 coreclr x64 win


highlyunavailable:
I'm trying to deserialize output from [Consul](https://www.consul.io/)'s JSON API into an array of this class:
```csharp
    public class KVPair
    {
        public string Key { get; set; }

        [JsonProperty]
        public ulong CreateIndex { get; private set; }

        [JsonProperty]
        public ulong ModifyIndex { get; set; }

        [JsonProperty]
        public ulong LockIndex { get; private set; }

        public ulong Flags { get; set; }
        public byte[] Value { get; set; }
        public string Session { get; set; }
    }
```

The `Flags` value is `0xe0f69a2baa414de0` in this case, and it only fails for the dnxcore50 (64-bit DNXCore 5.0) target, not dnx451 (64-bit DNX 4.5.1). I can also serialize these values successfully (I'm the one setting the Flags value, which in this case is a magic number for Consul), it's only deserialization that fails.

The exception I'm getting is: `Newtonsoft.Json.JsonReaderException : JSON integer 16210313421097356768 is too large or small for an Int64. Path '[0].Flags', line 1, position 112.`

The stack trace (not including my callsite) is:
```
           at Newtonsoft.Json.JsonTextReader.ParseNumber(ReadType readType)
           at Newtonsoft.Json.JsonTextReader.ParseValue()
           at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.ReadForType(JsonReader reader, JsonContract contract, Boolean hasConverter)
           at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.ResolvePropertyAndCreatorValues(JsonObjectContract contract, JsonProperty containerProperty, JsonReader reader, Type objectType)
           at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObjectUsingCreatorWithParameters(JsonReader reader, JsonObjectContract contract, JsonProperty containerProperty, ObjectConstructor`1 creator, String id)
           at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
[...]

PR #892, opened 2016-04-24
ulong.MaxValue parsing on .NET core

ah-:
Hi, great that you're working on .NET core support!

I noticed the same issue as in https://github.com/JamesNK/Newtonsoft.Json/issues/838, that ulong parsing is broken for values greater than what fits into a signed long.

I've added a test that demonstrates the problem and hacked together something that enables BigInteger parsing on netstandard11 and up.

How do you think would this be solved best? I fear this will trigger subtle bugs when porting from framework to core if not resolved.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
@@ -2582,6 +2582,20 @@ keyword such as type of business.""
             Assert.AreEqual(3, wibbleOut.Bar[4]);
         }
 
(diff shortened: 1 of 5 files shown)


JamesNK:
I want the lowest netstandard to be 1.0

I will probably support a higher version so that BigInteger works but I want to wait for a little bit to figure out exactly which one it should be.

akatz0813:
@JamesNK Did this end up getting fixed?  I think you indicated as such on another issue.  Can you point to the commit that fixes it?

JamesNK:
Not yet. It will probably be fixed in 9.0.2

PR #922, opened 2016-05-29
Add NamingStrategy

JamesNK:
* NamingStrategy
* Camel and snake case implementations
* Hooks on DefaultContractResolver and attributes


Src/Newtonsoft.Json.Tests/Serialization/CamelCaseNamingStrategyTests.cs
```diff
@@ -0,0 +1,327 @@
...
+            Person deserializedPerson = JsonConvert.DeserializeObject<Person>(json, new JsonSerializerSettings
+            {
+                ContractResolver = contractResolver
+            });
+
+            Assert.AreEqual(person.BirthDate, deserializedPerson.BirthDate);
+            Assert.AreEqual(person.LastModified, deserializedPerson.LastModified);
+            Assert.AreEqual(person.Name, deserializedPerson.Name);
+
+            json = JsonConvert.SerializeObject(person, Formatting.Indented);
```
Simon-Campbell: Would it be correct to say that this unit test is actually two functional tests (split at this line)? One for testing the CamelCaseNamingStrategy and another for testing the "default" naming strategy?

Src/Newtonsoft.Json.Tests/Utilities/StringUtilsTests.cs
```diff
@@ -50,14 +50,45 @@ public void ToCamelCaseTest()
...
+            Assert.AreEqual("", StringUtils.ToSnakeCase(""));
+            Assert.AreEqual(null, StringUtils.ToSnakeCase(null));
+            Assert.AreEqual("person", StringUtils.ToSnakeCase("Person"));
+            Assert.AreEqual("i_phone", StringUtils.ToSnakeCase("iPhone"));
+            Assert.AreEqual("i_phone", StringUtils.ToSnakeCase("IPhone"));
+            Assert.AreEqual("i_phone", StringUtils.ToSnakeCase("I Phone"));
+            Assert.AreEqual("i_phone", StringUtils.ToSnakeCase("I  Phone"));
+            Assert.AreEqual("i_phone", StringUtils.ToSnakeCase(" IPhone"));
+            Assert.AreEqual("i_phone", StringUtils.ToSnakeCase(" IPhone "));
+            Assert.AreEqual("is_cia", StringUtils.ToSnakeCase("IsCIA"));
```
Simon-Campbell: What is the expected output of something like IsJSONProperty? (capitalized acronym in middle of string) I assume it should be similar to how "URLValue" is resolved.

Src/Newtonsoft.Json/Serialization/NamingStrategy.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <summary>
+        /// A flag indicating whether dictionary keys should be processed.
+        /// Defaults to <c>false</c>.
+        /// </summary>
+        public bool ProcessDictionaryKeys { get; set; }
+
+        /// <summary>
+        /// A flag indicating whether explicitly specified property names,
+        /// e.g. a property name customized with a <see cref="JsonPropertyAttribute"/>, should be processed.
+        /// Defaults to <c>false</c>.
```
Simon-Campbell: This seems like a sensible default. Would you think it is worth having tests for the default behaviour? If this were to change by accident it would be a breaking change

Src/Newtonsoft.Json.Tests/Serialization/CamelCaseNamingStrategyTests.cs
JamesNK: Yeah. All the tests are a mess so I don't worry about them much :)

Src/Newtonsoft.Json.Tests/Utilities/StringUtilsTests.cs
JamesNK: Should be. I'll add a test for it

Src/Newtonsoft.Json/Serialization/NamingStrategy.cs
JamesNK: Defaults are implicitly tested in tests when they aren't set but there is no harm explicitly testing them.

Src/Newtonsoft.Json.Tests/Utilities/StringUtilsTests.cs
dotnetjunkie: It would be better to change thìs into a parameterized test instead.

Src/Newtonsoft.Json/JsonContainerAttribute.cs
```diff
@@ -58,22 +59,59 @@ public abstract class JsonContainerAttribute : Attribute
...
+        /// When non-null, there must be a constructor defined in the <see cref="JsonConverter"/> that exactly matches the number,
         /// order, and type of these parameters.
         /// </summary>
         /// <example>
         /// [JsonContainer(ItemConverterType = typeof(MyContainerConverter), ItemConverterParameters = new object[] { 123, "Four" })]
         /// </example>
         public object[] ItemConverterParameters { get; set; }
 
+        /// <summary>
+        /// Gets the <see cref="Type"/> of the <see cref="NamingStrategy"/>.
```
Eilon: `Gets or sets ...` ?

Src/Newtonsoft.Json/Serialization/NamingStrategy.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            return ResolvePropertyName(key);
+        }
+
+        /// <summary>
+        /// Resolves the specified property name.
+        /// </summary>
+        /// <param name="name">The property name to resolve.</param>
+        /// <returns>The resolved property name.</returns>
+        protected abstract string ResolvePropertyName(string name);
```
Eilon: Thinking out loud: This method is called for both "properties" and "dictionary keys," but its name specifically mentions "properties." Is there a name that can be used here that more clearly encompasses both? (In JSON the distinction is pretty much non-existent, but I believe the terminology here is more about C# than JSON.)

Eilon:
Looks great to me! Very easy to use, too. So, a :shipit: from me, for whatever that's worth. ( 🍰 )

Eilon:
BTW I did this code review cruising at 539MPH (867km/h) at an altitude of 37,007 feet (11,279m) somewhere over Minnesota. 😁

Src/Newtonsoft.Json/Serialization/NamingStrategy.cs
JamesNK: Yeah I know.

The json.org spec calls them names on an object, but I thought ResolveObjectName would be even more confusing.

JamesNK:
I wrote it in my pajamas moving at 0MPH. You win.

Eilon:
I bet your pajamas have more legroom... ✈️ + 💺 = 🙎

Src/Newtonsoft.Json/Converters/IsoDateTimeConverter.cs
```diff
@@ -57,7 +57,7 @@ public DateTimeStyles DateTimeStyles
         public string DateTimeFormat
         {
             get { return _dateTimeFormat ?? string.Empty; }
-            set { _dateTimeFormat = StringUtils.NullEmptyString(value); }
+            set { _dateTimeFormat = (string.IsNullOrEmpty(value)) ? null : value; }
```
dougbu: Curious: How does this change fit? And why the extra parentheses around the condition?

JamesNK: It doesn't. I just got rid of a method that is only used once at the same time.

Wrapping boolean expressions in parenthese like they are in an if statement is an old habit of mine.

niemyjski:
I was curious, how would this handle lower casing of characters as we recently ran into an issue with this on our customer serializer with turkish characters: https://github.com/exceptionless/Exceptionless/issues/228

PR #927, opened 2016-06-03
Added support for .NET Standard 1.0+

niemyjski:
This pull request adds support for .NET Standard 1.0-1.5. What this means is that the higher .net standard framework version you target you get more features as follows:

- 1.0: Bare minimum to run.
- 1.1: Added support for Concurrent Collections, Numerics (BigInteger), and System.Data.SqlClient types
- 1.2:
- 1.3: Reflection polyfills dropped, Xml Support added
- 1.4:
- 1.5: System.AppDomain support added

I've verified that the unit tests do all pass :), there may be some tests that could be enabled to run on .net standard due to it has more api surface than portable.

Notes:
Starting with .NET Standard 1.1, a reference on `System.Data.SqlClient` is required. However, I did add a compiler directive called `NO_SQL_CLIENT` that removes that functionality. This allows people source embedding to disable that functionality.


Sjark:
Looks good

niemyjski:
Any ideas on if this will be merged?

JamesNK:
I don't want to support every netstandard but post RTM I will add a higher level one, maybe 1.3.

niemyjski:
You don't have to, but at least having these directives allow source importers to support any level also allows you use use more features per each stepping stone

JamesNK:
Every time I add a new feature I would have to add the correct pre-processor flags all over the place to make the library compile. I don't want more than I need.

niemyjski:
We could probably find a way to do this better but shouldn't need it for most code, it's the code that depends on reflection or some new api that you'd need it and how often do you do that stuff? For example all of your new renaming code didn't need anything and you already have all these compiler directives in there for DOTNET and PORTABLE already. This was a lot of work and I think a lot of people would benefit from this patch. Especially when you get 1.0 support and nuget picks the highest net standard when referencing so you get more and more features automatically.

DaveVdE:
Well if we already have pre-processor flags all over the place, what's a few more going to hurt?

JamesNK:
Thanks @DaveVdE for volunteering to keep an up to date fork of Newtonsoft.Json with all the pre-processor flags. Anyone who wants the source code with additional binding flags can get it from your repo.

niemyjski:
This is a bit ridiculous :\ especially when .net standard support isn't really there (it's using portable, which .net standard 1.0 supports a lot more than portable subset). Also in the near future you can drop all other compiler directives in favor of .net standard which this pr adds support for in a minimalist way. In the end, if this is accepted or not, you'll be adding these same compiler directives to your code base, it's just a matter of time.

DaveVdE:
@JamesNK I never claimed I wanted to fork. Why this reaction? 😦

niemyjski:
@JamesNK if you want I can simplify this and remove some of the netstandard directives if you don't want to support them all.

niemyjski:
@JamesNK looks like my compiler directives match the microsoft recommended ones which are understood by the compiler: https://docs.microsoft.com/en-us/dotnet/articles/core/tutorials/libraries

ejsmith:
@JamesNK would you be willing to accept a minimal PR for just adding .NET Standard 1.3 support where reflection comes in? We would try to keep the changes as small as possible.

JamesNK:
Where reflection comes in? What do you mean by that?

I was planning to add 1.3 and add support for Reflection.Emit but some platforms don't support it and there is no way to figure that out at runtime. I'm going to need to figure that out before adding a higher level netstandard.

niemyjski:
We are source importing and used your portable compiler directive along with the `"NETStandard.Library": "1.6.0"` meta package. And it works great, however when you get to 1.3, the `"NETStandard.Library": "1.6.0"` brings in reflection and then were unable to compile with the `PORTABLE` compiler directive because of the reflection polyfills. 

I haven't heard anything about Reflection.Emit issues and I've tested on osx, windows and linux. Do you have a link to the issues you are seeing? Could always wrap the calls and fallback to the polyfill?

JamesNK:
Reflection.Emit doesn't work in AOT scenarios: Xamarin, Unity, UWP, etc. With netstandard there is just one library so it needs to handle it.

PR #934, opened 2016-06-20
Deserialize array object[] into object[] { TypeA, TypeB, TypeC, ... }

mozi-ru:
Make Json.Net deserialize arrays object[] into object[] { TypeA, TypeB, TypeC, ... } instead of object[] { JObject, JObject... }


JamesNK:
This is too large a breaking change. There is a lot of code that expects JObject values.

PR #938, opened 2016-06-21
Unseal JsonConverterAttribute

JayBazuzi:
I want to derive from this attribute like this:

```
    class MyConverterAttribute : JsonConverterAttribute(typeof(MyConverter)) {}
```


JayBazuzi:
My read on the situation was that the .Net Framework Design Guidelines said to seal attributes, so everyone does it, without knowing the implications. (And not many programmers think of deriving from attributes like this.)

pieceofsummer:
If attribute type has `AllowMultiple = false`, it won't allow to add two attributes of the same type. 

But if you derive from another attribute type, you'll be able add derived attribute as well as original attribute, which essentially is breaking original attribute contract. 

That's probably the main reason for sealing attributes.

JamesNK:
Hi

I don't want to do this because multiple attributes becomes an issue and I don't think the benefit outweighs the complexity.

PR #944, opened 2016-06-24
Added an extension point to bson reader to allow optimization

solyutor:
I've been writing a stock exchange trading application. It uses bson-based messages for communications. I see huge string traffic, but most of the string are very short (4-20 chars) and already interned. The PR gives me opportunity to inherit from the original bson writer, and use my own cache to deserialize strings.


JamesNK:
BSON has been split out into its own project

If it is still something you want then create a PR here https://github.com/JamesNK/Newtonsoft.Json.Bson

PR #969, opened 2016-07-22
Optimized double parsing

eisber:
Avoids string allocation


JamesNK:
Oh, interesting! I've always wanted to do this but floating point math isn't my forte. Thanks. I'll look at this later tonight. Key thing will be to make sure it behaves exactly the same as double.Parse.

JamesNK:
I've been adding some more tests and I found at least one case where the result isn't consistent:

            Assert.AreEqual(1E-23, Convert.ToDouble("1E-23"));

            c = "1E-23".ToCharArray();
            result = ConvertUtils.DoubleTryParse(c, 0, c.Length, out d);
            Assert.AreEqual(ParseResult.Success, result);
            Assert.AreEqual(1E-23, d);

ekevoo:
Most fractional decimal values equate to repeating fractional binary values. E.g. 0.1d == 1/10d == 1/1010b == 0.000110011001100110011…b

I notice that there's an array for positive exponents but negative exponents are calculated. I suggest having an array for the negative exponents as well. Another option is not having pre-calculated exponents at all and use Math.Pow.

eisber:
Math.Pow performance sucks…

ekevoo:
Well, I'll take your word for it. How about the negative exponents array approach?

JamesNK:
I've made a couple of changes to make it closer to the behavior of double.TryParse and add some more tests. Main issue at the moment is the inconsistency with some negative exponents: https://github.com/JamesNK/Newtonsoft.Json/commit/38a687f68395026100d9ee3b22b10ec1b696a914#diff-1e629cf095a196fc6ef6de98ae2fbd1fR85

@eisber can you look into it?

eisber:
Changes look good. I need to talk to some other folks at Microsoft about the negative exponents. I looked at 2 other code bases to get some ideas (https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parse_primitives.h#L57 and an internal one). The internal one makes use of some funky approximation for the exponent, but nothing special is done for negative exponents. In general, you can get quite some performance improvements once you strip the array access checks. I know James doesn’t want /unsafe, but maybe a separate build for high-perf scenarios (using defines) and given you have the arrays under control.

@ some explanation why Math.Pow sucks for the use case here: it allows for xˆy both being double and we only need (double)ˆ(int -308…308)

Do you have some guidance how I can reproduce the build errors? I ran all the tests in VS. Not sure how I influence:

Utilities\ExpressionReflectionDelegateFactoryTests.cs(369,122): error CS0246: The type or namespace name 'MyStruct' could not be found (are you missing a using directive or an assembly reference?) [C:\projects\newtonsoft-json\Working\Src\Newtonsoft.Json.Tests\Newtonsoft.Json.Tests.Portable.csproj]


Markus

JamesNK:
I've fixed the build error.

JamesNK:
Added a couple more tests to my branch https://github.com/JamesNK/Newtonsoft.Json/commit/4f86f6b79ed9cf30948c2a323f75148a46c3a1e4

JamesNK:
I benchmarked deserializing a list with 5000 double values in it a few thousand times.

Old: 1450ms
New: 825ms

Definitely seems worth it to get this finished.

eisber:
I reviewed our optimized double parse implementation and it also does contain a second map for negative exponent. I added the suggested fix to my PR

Markus

JamesNK:
Thanks!

Would it be possible to build up the list of positive and negative exponents in code with a loop in a static ctor rather than hardcoding them? Why did you choose to hardcode rather than build them at runtime?

eisber:
Simply calculating 1e-23 (e.g. by 10*10*…*10) yield wrong results. Big issue?

JamesNK:
They need to be hardcoded because calculating them will lose precision?

Src/Newtonsoft.Json/Utilities/ConvertUtils.cs
```diff
@@ -930,6 +930,137 @@ public static ParseResult Int64TryParse(char[] chars, int start, int length, out
...
+            if (exponent > 0)
+            {
+                value *= DoubleExponents[exponent - 1];
+            }
+            else if (exponent < 0)
+            {
+                exponent = -exponent;
+                if (exponent > MaxExponent)
+                {
+                    // cover 4.94065645841247E-324
```
JamesNK: Dumb question but is this to cover all large negative exponents or just this one particular one?

eisber:
Yes, essentially you accumulate some numeric error.

Src/Newtonsoft.Json/Utilities/ConvertUtils.cs
eisber: ☺ it’s to cover more. I starred at this code today as well, I think beyond 324 it’s Infinity…

JamesNK:
I'm not sure if you ran the unit tests but there are now other tests that are failing, e.g.

```
Input string: 0.25e-5
Expected: 2.5000000000000002E-06d
But was:  2.4999999999999998E-06d
```

JamesNK:
Hi @eisber. When do you think you'll get a chance to look at this? I would fix it myself but I don't have much knowledge around floating point errors that I've described above.

Src/Newtonsoft.Json/Utilities/ConvertUtils.cs
```diff
@@ -930,6 +930,137 @@ public static ParseResult Int64TryParse(char[] chars, int start, int length, out
...
+                    case '1':
+                    case '2':
+                    case '3':
+                    case '4':
+                    case '5':
+                    case '6':
+                    case '7':
+                    case '8':
+                    case '9':
+                        value = (10 * value) + (c - '0');
```
TylerBrinkley: I think you might want to use an unsigned long to store the mantissa to avoid precision loss when building it up. Of course then you'll need to keep a count of how many places are stored so as to not overflow the unsigned long.

BrutalSimplicity: Wouldn't that require `unsafe` code? If it does, is that a problem?

TylerBrinkley:
In my branch, https://github.com/TylerBrinkley/Newtonsoft.Json/commit/5fd78b21f5bcb3cd573980b6e4e84665ee434c3e, I've made the mantissa be stored in a ulong instead of the double to prevent accumulating numeric error. Also, to get all of the unit tests to pass, I added a case where if the negative exponent is less than 23 to divide from the positive exponents array instead of multiply from the negative exponents array.

Src/Newtonsoft.Json/Utilities/ConvertUtils.cs
TylerBrinkley: It does not require `unsafe` code. See my branch at TylerBrinkley@5fd78b2.

BrutalSimplicity:
What about using `BitConverter.Int64BitsToDouble`. That way you can convert the complete binary representation (as a long) directly into a double. The .net core code looks like it's just a pointer cast.

```csharp
public static unsafe double Int64BitsToDouble(long value) {
    // If we're on a big endian machine, what should this method do?  You could argue for
    // either big endian or little endian, depending on whether you are writing to a file that
    // should be used by other programs on that processor, or for compatibility across multiple
    // formats.  Because this is ambiguous, we're excluding this from the Portable Library & Win8 Profile.
    // If we ever run on big endian machines, produce two versions where endianness is specified.
    Contract.Assert(IsLittleEndian, "This method is implemented assuming little endian with an ambiguous spec.");
    return *((double*)&value);
} 
```

pieceofsummer:
@BrutalSimplicity it doesn't really plays out that well.

I've made some tests, comparing the implementation from @TylerBrinkley (with one `double` multiplication) with the one manually "packing" a double (based on `NumberToDouble` native function from coreclr sources), and it turns out that two lossy `ulong` multiplications in managed code (with no appropriate inlining) are performing worse, especially when in 32-bit mode, and it involves more branching too.

My test results for 10M iterations:
```
No optimization, 64-bit
    double.TryParse: 1526ms
    TylerBrinkley's: 1030ms
    CustomPack:      1149ms

No optimization, 32-bit
    double.TryParse: 1366ms
    TylerBrinkley's: 1228ms
    CustomPack:      1552ms

Optimize, 64-bit
    double.TryParse: 1399ms
    TylerBrinkley's: 314ms
    CustomPack:      354ms

Optimize, 32-bit
    double.TryParse: 1274ms
    TylerBrinkley's: 492ms
    CustomPack:      556ms
```

I just have to point out an error in @TylerBrinkley implementation. Since `exponent > MaxExponent` is checked *before* correcting decimal points, expressions like `0.025E+309` result in overflow, while `double.Parse` produces correct result `2.5E+307`.

TylerBrinkley:
@pieceofsummer, that's very interesting. Do you have your source posted anywhere? I'd like to look at how you did it. As for the bug with numbers like `0.025E+309`, that was fixed in my second commit. See https://github.com/JamesNK/Newtonsoft.Json/compare/eisber-master...TylerBrinkley:eisber-master

Technically, valid json numbers must start with a digit but if `.1E23` would parse in the latest version of Json.NET then we should preserve that behavior.

pieceofsummer:
@TylerBrinkley sure, https://gist.github.com/pieceofsummer/a555baa83a3c6f71925dac9f2d8b2d86. 

I was too lazy to write my own parser, so I just took yours (with minimal changes).

TylerBrinkley:
@pieceofsummer, awesome work on porting that over to C#. I'd much prefer using your packing implementation as it should produce the same results as `double.TryParse` which would be critical to prevent breaking changes, per @JamesNK 

> Key thing will be to make sure it behaves exactly the same as double.Parse.

The slight performance drop would be well worth that guarantee. Additionally it has a smaller static array footprint than the 616 doubles in the current implementation. I tested your code and it's passing all of the tests. To get your changes in you could you fork my branch and insert your packing implementation and then submit a new pull request or I could insert your packing implementation in my branch and then submit a new pull request but I wouldn't want to take credit for your hard work porting that packing implementation over to C#.

It looks like Json.NET allowed any values that could be `double.TryParse`'d so `.1E23` should be allowed as well as `1.E23`. I've updated my branch to allow that.

pieceofsummer:
@TylerBrinkley it would be #1010 now

JamesNK:
> It looks like Json.NET allowed any values that could be double.TryParse'd so .1E23 should be allowed as well as 1.E23. I've updated my branch to allow that.

I want the new method to reject those as invalid. The only place that this will be used is JsonTextReader and the spec requires that doubles follow an exact format. For converting from a string to a double Json.NET can keep using double.TryParse

JamesNK:
@eisber Do you have any thoughts on the changes in the new PR #1010? I'd like to hear what you think before replacing some of your implementation.

eisber:
I looked through the PR and seems reasonable. The code is started was more concerned with speed than with BCL compatibility, thus this seems the right move.
Had another look at our optimized parse implementation and it’s also clearly optimized for speed and sacrifices BCL compatibility.

Markus

JamesNK:
Done

TylerBrinkley:
I was just poking around `System.Runtime.Numerics` and discovered they already have a `C#` `double` packing implementation in the file [NumericsHelper.cs](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Numerics/src/System/Numerics/NumericsHelpers.cs). No offense to @pieceofsummer but I'd be more comfortable with using their `double` packing implementation assuming its results are the same as what's used in `double.Parse`. Any thoughts?

TylerBrinkley:
Disregard, I think that `double` packing implementation is for values in base-2.

PR #976, opened 2016-07-28
Support for C++/CLI (.NET 4.5)

fubar-coder:
This PR adds support for C++/CLI (.NET 4.5).

An import node for the `.targets` file gets added to the project during the installation of the NuGet package, which includes all necessary references and reuses the `net45` build. IntelliSense in VS2015 Update 3 recognizes this change immediately.

The "trick" with the `_._` file is required, because NuGet needs a `lib` entry named `native` (TFM).

A `runbuild.ps1` and a test using a private NuGet feed works as expected.


JamesNK:
Hi

I'm not sure whether this is still a thing. Comment here if this is important. My perspective is Json.NET is focused on managed consumers.

PR #994, opened 2016-08-21
Support BigInteger on .NET Core (netstandard1.1+)

qmfrederik:
I'm migrating to .NET Core and I'm currently blocked because Newtonsoft.Json doesn't support BigInteger on .NET Core (#892, #927).

I've rebased @niemyjski's #927 against the latest master and verified all targets build correctly.

From the #927 discussion, I understand you don't want to maintain builds for all the individual netstandard versions.
That makes sense to me, so I'm offering to rewrite the PR to target a limited number of netstandard versions (e.g. adding support for only 1.1 would fix the BigInteger support, which is "good enough" for me).

Let me know if you're open to that & happy to help out.


Src/Newtonsoft.Json.Tests/Properties/AssemblyInfo.cs
```diff
@@ -33,6 +33,16 @@
 
 #if NETSTANDARD1_0
 [assembly: AssemblyTitle("Json.NET Tests .NET Standard 1.0")]
+#elif NETSTANDARD1_1
+[assembly: AssemblyTitle("Json.NET .NET Standard 1.1")]
```
ah-: Json.NET Tests .NET Standard 1.1?

pieceofsummer:
I'm wondering, why `BigInteger` (along with `SqlBinary` and other exotic types) was even supported on non-Core platforms to begin with. 

I can hardly imagine a scenario where raw `BigInteger`s are needed to be serialized in JSON (encryption keys aside, they're usually stored as base64 strings, and read/written just a few times during the whole application lifetime), yet `BigInteger` type checks are performed in the busy-loaded `JsonReader` on almost every read. And it also adds an extra assembly/package, while not being used by the most of the users.

I'd rather created a side package with custom serializers for these rarely used types, while keeping the main codebase small and robust.

JamesNK:
Done https://github.com/JamesNK/Newtonsoft.Json/commit/68d160f3c608cd6850b207143b2ba69151500613

PR #1002, opened 2016-09-01
Advanced JsonConverter constructor parameter binding on .NET Core & portable

pieceofsummer:
Because of using a non-standard method binder on .NET Core/Portable, the library is overly strict about constructors' signatures, and in some cases it even doesn't work as expected.

For example, take a simple converter with `System.Type` argument:
```c#
public class CustomJsonConverter : JsonConverter 
{
    public CustomJsonConverter(Type arg)
    {
    }

    ...
}
```
and try to serialize an object with it applied on .NET Core:
```c#
public class SomeClass
{
    [JsonConverter(typeof(CustomJsonConverter), typeof(int))]
    public int SomeProperty { get; set; }
}

var json = JsonConvert.SerializeObject(new SomeClass());
```

You will get a `No matching parameterized constructor found for …` exception, because on .NET Core `System.RuntimeType` (as returned by `typeof()`) and `System.Type` (from `ParameterInfo.ParameterType`) are not equal, hence `SequenceEqual()` comparison fails.

Proposed PR fixes this issue, and also attempts to fix the shortcomings of the current binder, making it behave more like standard .NET binder. 

It adds the following features:
- Base type binding for classes (for example, you can pass `string[]` value for `IEnumerable<string>` parameter).
- Type widening for primitive values (for example, you can pass `int` value for `long` or `float` parameter, no explicit casting required).
- Best constructor overload selection based on provided values' types.
- It even supports `params` arguments (but since `ReflectionDelegateFactory` doesn't support them, it is currently disabled, and `params` arrays are handled as ordinary arrays).

P.S. The new binder currently handles constructors and methods, while indexer properties still use the old one. Does anybody have an idea where indexer properties are actually used? Do we need an advanced binding for them at all?


JamesNK:
I'm looking at this now.

One thing I've noticed just from looking at the tests is an unsigned long value is being passed to an overload that takes a double. I don't want that.

JamesNK:
I wonder how big a performance impact calling Convert.ChangeType will be on every primitive type. Would checking whether the type isn't already the correct type first (whatever the op code and expression is for `is`) would make a difference.

Src/Newtonsoft.Json/Utilities/TypeExtensions.cs
```diff
@@ -273,6 +273,226 @@ public static bool IsDefined(this Type type, Type attributeType, bool inherit)
...
+
+            if (paramArrayType == null && parameters.Length != types.Count)
+            {
+                // when there's no ParamArray, number of parameters should match
+                return false;
+            }
+            
+            for (int i = 0; i < types.Count; i++)
+            {
+                var paramType = (paramArrayType != null && i >= parameters.Length - 1) ? paramArrayType : parameters[i].ParameterType;
```
JamesNK: Specify the type. Json.NET is a no-var zone.

Src/Newtonsoft.Json/Utilities/TypeExtensions.cs
```diff
@@ -273,6 +273,226 @@ public static bool IsDefined(this Type type, Type attributeType, bool inherit)
...
+                }
+            }
+            
+            return (fromMask & toMask) != 0;
+        }
+
+        private static bool FilterParameters(ParameterInfo[] parameters, IList<Type> types, bool enableParamArray)
+        {
+            if (parameters == null)
+                throw new ArgumentNullException(nameof(parameters));
```
JamesNK: Brackets on all `if`s

Src/Newtonsoft.Json/Utilities/TypeExtensions.cs
```diff
@@ -273,6 +273,226 @@ public static bool IsDefined(this Type type, Type attributeType, bool inherit)
...
+            0x0001,        0x0FE2,         0x0D54,        0x0FFA,
+            0x0D50,        0x0FE2,         0x0D40,        0x0F80,
+            0x0D00,        0x0E00,         0x0C00,        0x0800
+        };
+        
+        internal static bool CanConvertPrimitive(Type from, Type to)
+        {
+            if (from == to) return true;
+
+            int fromMask = 0, toMask = 0;
```
JamesNK: Declare individually

Src/Newtonsoft.Json/Utilities/TypeExtensions.cs
```diff
@@ -273,6 +273,226 @@ public static bool IsDefined(this Type type, Type attributeType, bool inherit)
...
+
+            for (int i = 0; i < PrimitiveTypes.Length && (fromMask == 0 || toMask == 0); i++)
```
JamesNK: Add a comment about what is going on here

Src/Newtonsoft.Json/Utilities/TypeExtensions.cs
```diff
@@ -273,6 +273,226 @@ public static bool IsDefined(this Type type, Type attributeType, bool inherit)
...
+
+                // parameterless method wins
+                if (parameters1.Length == 0) return -1;
+                if (parameters2.Length == 0) return 1;
+
+                Type paramArrayType1 = null, paramArrayType2 = null;
+
+                if (_enableParamArray)
+                {
+                    if (parameters1[parameters1.Length - 1].ParameterType.IsArray &&
```
JamesNK: Assign `parameters1[parameters1.Length - 1]` to a local variable rather than accessing the array multiple times

Src/Newtonsoft.Json/Utilities/TypeExtensions.cs
```diff
@@ -273,6 +273,226 @@ public static bool IsDefined(this Type type, Type attributeType, bool inherit)
...
+                        parameters1[parameters1.Length - 1].IsDefined(typeof(ParamArrayAttribute)))
+                    {
+                        paramArrayType1 = parameters1[parameters1.Length - 1].ParameterType.GetElementType();
+                    }
+
+                    if (parameters2[parameters2.Length - 1].ParameterType.IsArray &&
```
JamesNK: ^^

JamesNK:
> Does anybody have an idea where indexer properties are actually used?

I can't remember...

pieceofsummer:
> unsigned long value is being passed to an overload that takes a double. Is that what the default .NET binder does?

Yes, it does. The type widening rules are exactly the same as default .NET binder uses.

> I wonder what the performance impact is of calling Convert.ChangeType will be on every primitive value.

I was actually shocked when I tested it. It made method invocation like 3-4 times slower! That's why I've spent this evening rewriting them.

Current stats (release build, 10M constructors with single primitive argument):
```
DynamicReflectionDelegateFactory (.NET45):
    original (no type widening): 246ms
    proposed (exact type match): 253ms
    proposed (arg type widened): 301ms

ExpressionReflectionDelegateFactory (.NET45):
    original (no type widening): 355ms
    proposed (exact type match): 412ms
    proposed (arg type widened): 417ms

ExpressionReflectionDelegateFactory (.NET Core):
    original (no type widening): 253ms
    proposed (exact type match): 309ms
    proposed (arg type widened): 313ms
```
Much better than `Convert.ChangeType()`, which took ~ 1 second for the same  task.

JamesNK:
That looks much better.

Is there a reason the expression factory using Convert.ToInt32 and the emit factory uses IConvertable.ToInt32?

Src/Newtonsoft.Json/Utilities/MethodBinder.cs
```diff
@@ -0,0 +1,333 @@
...
+        {
+            if (from == to)
+            {
+                // same type
+                return true;
+            }
+
+            int fromMask = 0, toMask = 0;
+
+            for (int i = 0; i < PrimitiveTypes.Length && (fromMask == 0 || toMask == 0); i++)
```
JamesNK: Can you move the `(fromMask == 0 || toMask == 0)` to its own if statement inside the for loop?

Src/Newtonsoft.Json/Utilities/MethodBinder.cs
```diff
@@ -0,0 +1,333 @@
...
+        /// <returns>True if source type can be widened to target type, false otherwise</returns>
+        internal static bool CanConvertPrimitive(Type from, Type to)
+        {
+            if (from == to)
+            {
+                // same type
+                return true;
+            }
+
+            int fromMask = 0, toMask = 0;
```
JamesNK: Declare on separate lines

Src/Newtonsoft.Json/Utilities/ExpressionReflectionDelegateFactory.cs
```diff
@@ -349,17 +338,41 @@ public override Func<T> CreateDefaultConstructor<T>(Type type)
...
+                {
+                    MethodInfo toTargetTypeMethod = typeof(Convert)
+                        .GetMethod("To" + targetType.Name, new[] { typeof(object) });
+
+                    if (toTargetTypeMethod != null)
+                    {
+                        convert = Expression.Condition(
+                            Expression.TypeIs(expression, targetType),
+                            convert,
+                            Expression.Call(toTargetTypeMethod, expression));
```
JamesNK: Pass CultureInfo.Invariant to convert

Src/Newtonsoft.Json/Utilities/DynamicReflectionDelegateFactory.cs
```diff
@@ -142,29 +144,69 @@ private void GenerateCreateMethodCallIL(MethodBase method, ILGenerator generator
...
+                            generator.Emit(OpCodes.Ldloc_S, localObject);
+                            generator.Emit(OpCodes.Isinst, typeof(IConvertible));
+                            generator.Emit(OpCodes.Stloc_S, localConvertible);
+                            generator.Emit(OpCodes.Ldloc_S, localConvertible);
+                            generator.Emit(OpCodes.Brfalse_S, skipConvertible);
+
+                            // convert argument to parameter type
+                            generator.Emit(OpCodes.Ldloc_S, localConvertible);
+                            generator.Emit(OpCodes.Ldnull);
+                            generator.Emit(OpCodes.Callvirt, toParameterTypeMethod);
```
JamesNK: Pass CultureInfo.Invariant to convert

Src/Newtonsoft.Json/Utilities/MethodBinder.cs
```diff
@@ -0,0 +1,333 @@
...
+
+        /// <summary>
+        /// Retunrs a best method overload for given argument <paramref name="types"/>.
+        /// </summary>
+        /// <param name="candidates">List of method candidates</param>
+        /// <param name="types">Argument types</param>
+        /// <returns>Best method overload, or <c>null</c> if none matched</returns>
+        public static TMethod SelectMethod<TMethod>(IEnumerable<TMethod> candidates, IList<Type> types) where TMethod : MethodBase
+        {
+            if (candidates == null)
```
JamesNK: Replace with ValidationUtils.ArgumentNotNull

Src/Newtonsoft.Json/Utilities/MethodBinder.cs
```diff
@@ -0,0 +1,333 @@
...
+            {
+                throw new ArgumentNullException(nameof(candidates));
```
JamesNK: Replace with ValidationUtils.ArgumentNotNull

Src/Newtonsoft.Json/Utilities/MethodBinder.cs
```diff
@@ -0,0 +1,333 @@
...
+        /// </summary>
+        /// <param name="parameters">Method parameters</param>
+        /// <param name="types">Argument types</param>
+        /// <param name="enableParamArray">Try to pack extra arguments into ParamArray</param>
+        /// <returns>True if method can be called with given arguments, false otherwise</returns>
+        private static bool FilterParameters(ParameterInfo[] parameters, IList<Type> types, bool enableParamArray)
+        {
+            if (parameters == null)
+            {
+                throw new ArgumentNullException(nameof(parameters));
```
JamesNK: Replace with ValidationUtils.ArgumentNotNull

Src/Newtonsoft.Json/Utilities/MethodBinder.cs
```diff
@@ -0,0 +1,333 @@
...
+        private class ParametersMatchComparer : IComparer<ParameterInfo[]>
+        {
+            private readonly IList<Type> _types;
+            private readonly bool _enableParamArray;
+
+            public ParametersMatchComparer(IList<Type> types, bool enableParamArray)
+            {
+                if (types == null)
+                {
+                    throw new ArgumentNullException(nameof(types));
```
JamesNK: Replace with ValidationUtils.ArgumentNotNull

Src/Newtonsoft.Json/Utilities/MethodBinder.cs
```diff
@@ -0,0 +1,333 @@
...
+
+                _types = types;
+                _enableParamArray = enableParamArray;
+            }
+
+            public int Compare(ParameterInfo[] parameters1, ParameterInfo[] parameters2)
+            {
+                if (parameters1 == null)
+                {
+                    throw new ArgumentNullException(nameof(parameters1));
```
JamesNK: ^^

pieceofsummer:
> Is there a reason the expression factory using Convert.ToInt32 and the emit factory uses IConvertable.ToInt32?

Yes, there is. 

`Convert.ToInt32` internally casts an argument to `IConvertible` and calls `IConvertible.ToInt32`. Since we already have a non-null `IConvertible` value, calling it directly is somewhat faster (saves null check, type cast and method call).

On the other hand, expression factory should work on portable frameworks, which might not have `IConvertible` (it is part of `netstandard1.3`), so the more portable `Convert.ToInt32` is used.

Src/Newtonsoft.Json/Utilities/ExpressionReflectionDelegateFactory.cs
pieceofsummer: Culture info parameter is ignored when converting between primitive types. Do we need to waste a method call?

Src/Newtonsoft.Json/Utilities/DynamicReflectionDelegateFactory.cs
pieceofsummer: Same here

Src/Newtonsoft.Json/Utilities/DynamicReflectionDelegateFactory.cs
```diff
@@ -90,7 +90,9 @@ private void GenerateCreateMethodCallIL(MethodBase method, ILGenerator generator
                 generator.PushInstance(method.DeclaringType);
             }
 
-            int localVariableCount = 0;
+            LocalBuilder localConvertible = generator.DeclareLocal(typeof(IConvertible));
+            LocalBuilder localObject = generator.DeclareLocal(typeof(object));
+            int localVariableCount = 2;
```
pieceofsummer: Maybe remove `localVariableCount`? It is not actually needed anywhere.

Or we should throw an exception if `localVariableCount > 255` (so short instructions cannot be used). But it is unlikely to have a method taking more than 254 arguments.

JamesNK:
> which might not have IConvertible

Good point. I had forgotten about that.

JamesNK:
Looks good to me. I removed localVariableCount.

Merged!

Issue #585, opened 2015-06-11
Generic JsonConverter?


Crisfole:
It's a bit of a pain to be running casts all over for JsonConverter.  I tried searching the issues and never found a suggestion like this, but instead of copying this around my projects it seems like a good fit for 'core' Newtonsoft.Json:

``` csharp

    public abstract class JsonConverter<T> : JsonConverter {
        public sealed override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer) {
            if (!(value is T)) {
                throw new JsonSerializationException(string.Format("This converter cannot convert {1} of type {0}", value, value.GetType()));
            }
            WriteJson(writer, (T)value, serializer);
        }

        public abstract void WriteJson(JsonWriter writer, T value, JsonSerializer serializer);

        public sealed override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer) {
            if (objectType != typeof (T)) {
                throw new JsonSerializationException(string.Format("This converter cannot convert type {0}", objectType));
            }
            if (!(existingValue is T)) {
                throw new JsonSerializationException(string.Format("This converter cannot convert {1} of type {0}", existingValue, existingValue.GetType()));
            }
            return ReadJson(reader, (T) existingValue, serializer);
        }

        public abstract T ReadJson(JsonReader reader, T existingValue, JsonSerializer serializer);

        public sealed override bool CanConvert(Type objectType) {
            return typeof (T) == objectType;
        }
    }
```

By leaving the non-generic intact it makes it feasible to write however crazy a Converter you'd like, but this case covers 99% of the simple cases.

Tydude4Christ:
I really like this suggestion. I use a similar approach to a default parser system I created.

hlubovac:
Somewhat related (same/similar result, but different approach): perhaps this functionality exists and I'm not seeing it, but wouldn't it be better if the `abstract class JsonConverter` had been an interface instead? It's not adding any functionality to the derived class, except forcing it to implement method, for which interfaces are a better fit.

[...]

PR #1011, opened 2016-09-08
Added a generic JsonConverter

TylerBrinkley:
Added a generic JsonConverter and changed the base types of some of the existing converters to this new generic JsonConverter where relevant. This fulfills #585.

### DIFF
--- a/Src/Newtonsoft.Json/Converters/BsonObjectIdConverter.cs
+++ b/Src/Newtonsoft.Json/Converters/BsonObjectIdConverter.cs
@@ -33,7 +33,7 @@ namespace Newtonsoft.Json.Converters
     /// <summary>
     /// Converts a <see cref="BsonObjectId"/> to and from JSON and BSON.
     /// </summary>
-    public class BsonObjectIdConverter : JsonConverter
+    public class BsonObjectIdConverter : JsonConverter<BsonObjectId>
     {
         /// <summary>
         /// Writes the JSON representation of the object.
@@ -41,18 +41,16 @@ namespace Newtonsoft.Json.Converters
         /// <param name="writer">The <see cref="JsonWriter"/> to write to.</param>
         /// <param name="value">The value.</param>
         /// <param name="serializer">The calling serializer.</param>
-        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        public override void WriteJson(JsonWriter writer, BsonObjectId value, JsonSerializer serializer)
         {
-            BsonObjectId objectId = (BsonObjectId)value;
-
             BsonWriter bsonWriter = writer as BsonWriter;
             if (bsonWriter != null)
             {
-                bsonWriter.WriteObjectId(objectId.Value);
+                bsonWriter.WriteObjectId(value.Value);
             }
             else
             {
-                writer.WriteValue(objectId.Value);
+                writer.WriteValue(value.Value);
             }
         }
 
@@ -61,10 +59,11 @@ namespace Newtonsoft.Json.Converters
         /// </summary>
         /// <param name="reader">The <see cref="JsonReader"/> to read from.</param>
         /// <param name="objectType">Type of the object.</param>
+        /// <param name="existingHasValue">The existing value has a value.</param>
         /// <param name="existingValue">The existing value of object being read.</param>
         /// <param name="serializer">The calling serializer.</param>
         /// <returns>The object value.</returns>
-        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        public override BsonObjectId ReadJson(JsonReader reader, Type objectType, bool existingHasValue, BsonObjectId existingValue, JsonSerializer serializer)
         {
             if (reader.TokenType != JsonToken.Bytes)
             {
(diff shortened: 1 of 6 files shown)


fgreinacher:
I like this a lot!

Src/Newtonsoft.Json/JsonConverter.cs
```diff
@@ -96,4 +97,75 @@ public virtual bool CanWrite
...
+    public abstract class JsonConverter<T> : JsonConverter
+    {
+        /// <summary>
+        /// Writes the JSON representation of the object.
+        /// </summary>
+        /// <param name="writer">The <see cref="JsonWriter"/> to write to.</param>
+        /// <param name="value">The value.</param>
+        /// <param name="serializer">The calling serializer.</param>
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
```
fgreinacher: `sealed override`

Src/Newtonsoft.Json/JsonConverter.cs
```diff
@@ -96,4 +97,75 @@ public virtual bool CanWrite
...
+        /// <summary>
+        /// Reads the JSON representation of the object.
+        /// </summary>
+        /// <param name="reader">The <see cref="JsonReader"/> to read from.</param>
+        /// <param name="objectType">Type of the object.</param>
+        /// <param name="existingValue">The existing value of object being read.</param>
+        /// <param name="serializer">The calling serializer.</param>
+        /// <returns>The object value.</returns>
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
```
fgreinacher: `sealed override`

Src/Newtonsoft.Json/JsonConverter.cs
TylerBrinkley: I considered that but it would be a breaking change since I changed existing non-sealed JsonConverter classes' base class to this new generic JsonConverter.

Src/Newtonsoft.Json/JsonConverter.cs
TylerBrinkley: I considered that but it would be a breaking change since I changed existing non-sealed JsonConverter classes' base class to this new generic JsonConverter.

Src/Newtonsoft.Json/JsonConverter.cs
fgreinacher: I see, but might still be a valuable change to make (@jamesnk?)

Src/Newtonsoft.Json/JsonConverter.cs
```diff
@@ -96,4 +97,75 @@ public virtual bool CanWrite
...
+            var existingIsNull = existingValue == null;
+            if (!(existingIsNull || existingValue is T))
```
TylerBrinkley: T isn't constrained to be a reference type so I can't use the `as` operator.

Src/Newtonsoft.Json/JsonConverter.cs
JamesNK: I've never used sealed override but I expect it would be a breaking change for anyone who has inherited and overrided a sealed method from a converter whose base type was changed to the generic converter.

fgreinacher: Yes, it is a breaking change for existing inheritors. But if you decide to take this PR, I'd strongly suggest to take it (next version will be a major version bump I think?). Otherwise the `override` semantics will be pretty unclear (e.g. overriding the generic method will have no effect if the non-generic method is also overridden).

JamesNK: No breaking changes.

fgreinacher: OK :)

Src/Newtonsoft.Json/JsonConverter.cs
fgreinacher: OK

fgreinacher:
LGTM

Src/Newtonsoft.Json/Converters/BsonObjectIdConverter.cs
```diff
@@ -33,26 +33,24 @@ namespace Newtonsoft.Json.Converters
     /// <summary>
     /// Converts a <see cref="BsonObjectId"/> to and from JSON and BSON.
     /// </summary>
-    public class BsonObjectIdConverter : JsonConverter
+    public class BsonObjectIdConverter : JsonConverter<BsonObjectId>
```
JamesNK: So JsonConverter<T> inherits from JsonConverter, but is changing the base type to JsonConverter<T> a breaking change?

TylerBrinkley: According to https://stackoverflow.com/questions/1456785/a-definitive-guide-to-api-breaking-changes-in-net#1522718 changing the base class won't be a breaking change besides through reflection.

JamesNK:
I rolled back the changes to the built-in converters because I want to have converters start avoid loading types until the actual conversion.

TylerBrinkley:
That's cool. I'm just glad it's being added. Does that mean we can seal some of the methods?

JamesNK:
Hmmm, yes they could be sealed.

Issue #997, opened 2016-08-26
DeserializeXNode in .net core


wojtpl2:
In attachment is console project with very simple code using "Newtonsoft.Json": "9.0.1":
```cs
string json = @"{
    ""CreatedDate"": ""2014-01-23T00:00:00""
}";
var dxml = JsonConvert.DeserializeXNode(json, "root");
Console.WriteLine("CreatedDate: " + dxml.Root.Element("CreatedDate").Value);
Console.WriteLine(dxml.ToString());
```

If you run this application on .net net452 you see output:
```xml
CreatedDate: 2014-01-23T00:00:00
<root>
  <CreatedDate>2014-01-23T00:00:00</CreatedDate>
</root>
```

If you run this application as .netCoreApp you see:
```xml
CreatedDate: 2014-01-23T00:00:00+01:00
<root>
  <CreatedDate>2014-01-23T00:00:00+01:00</CreatedDate>
</root>
```

Function DeserializeXNode add time zone to datetime in in .net core.

I check that any test from XmlNodeConverterTest.cs not run for .net core. Function DeserializeXNode don't have test on this platform.

[DeserializeXNode.zip](https://github.com/JamesNK/Newtonsoft.Json/files/438797/DeserializeXNode.zip)

pieceofsummer:
This is because overload `XmlConvert.ToString(DateTime, XsdDateTimeFlags)` is not avaliable for `netstandard1.0`, and it ends up with implicit conversion of `DateTime` to `DateTimeOffset` (which doesn't support `DateTimeKind.Unspecified` and assumes local time zone instead). 

See https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Converters/XmlNodeConverter.cs#L1677.

Since date format in XML is fixed and culture-invariant, I'd suggest to replace `XmlConvert` with custom formatter for `DateTime` values.

wojtpl2:
You should read datetime like this:
return XmlConvert.ToString((DateTime)value, XmlDateTimeSerializationMode.RoundtripKind); 

see .net core (corefx) source
https://github.com/dotnet/corefx/blob/d0dc5fc099946adc1035b34a8b1f6042eddb0c75/src/System.Xml.XmlSerializer/src/System/Xml/Serialization/XmlCustomFormatter.cs#L202 

or see ExtendedXmlSerializer
[...]

PR #1015, opened 2016-09-09
Fix DeserializeXNode in .net core

wojtpl2:
Fix for issue [997](https://github.com/JamesNK/Newtonsoft.Json/issues/997)

Now in portable version (including the .net core) format for the DataTime in xml will be the same as the normal version of the library.

### DIFF
--- a/Src/Newtonsoft.Json/Utilities/DateTimeUtils.cs
+++ b/Src/Newtonsoft.Json/Utilities/DateTimeUtils.cs
@@ -73,6 +73,21 @@ namespace Newtonsoft.Json.Utilities
                     throw MiscellaneousUtils.CreateArgumentOutOfRangeException("kind", kind, "Unexpected DateTimeKind value.");
             }
         }
+#else
+        public static string ToDateTimeFormat(DateTimeKind kind)
+        {
+            switch (kind)
+            {
+                case DateTimeKind.Local:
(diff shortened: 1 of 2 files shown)


Src/Newtonsoft.Json/Utilities/DateTimeUtils.cs
```diff
@@ -73,6 +73,21 @@ public static XmlDateTimeSerializationMode ToSerializationMode(DateTimeKind kind
                     throw MiscellaneousUtils.CreateArgumentOutOfRangeException("kind", kind, "Unexpected DateTimeKind value.");
             }
         }
+#else
+        public static string ToDateTimeFormat(DateTimeKind kind)
+        {
+            switch (kind)
+            {
+                case DateTimeKind.Local:
+                    return "yyyy-MM-ddTHH:mm:sszzzzzzz";
```
pieceofsummer: Why so many `z`'s? https://msdn.microsoft.com/en-us/library/8kb3ddd4 says timezone specifier allows up to 3 `z`'s, having more doesn't seem to make any difference.

JamesNK:
I think he meant F's

https://github.com/JamesNK/Newtonsoft.Json/commit/39ecc359eb46ec38e50579389118a079c32e3eb3

PR #1042, opened 2016-10-11
Clean up NETFX_CORE directive usage

pyitphyoaung:


JamesNK:
Thanks! This is something I've been meaning to do but have never gotten around to doing.

pyitphyoaung:
I'm just glad that I could help, @JamesNK. :)

PR #1047, opened 2016-10-14
Clean up codes

pyitphyoaung:


JamesNK:
Thanks but I prefer the existing code. Sometimes longer is easier to read.

PR #1051, opened 2016-10-18
Stops including typename for nullable structs using TypeNameHandling.Auto

### LINKED ISSUE #1050: JSON.NET serialisation of nullable structs and TypeNameHandling.Auto
The following program:

    using System;
    using Newtonsoft.Json;
    
    namespace JsonPlayground
    {
        public class Program
        {
            private struct Message
            {
                public string Value { get; }
    
                [JsonConstructor]
                public Message(string value)
                {
                    if (value == null) throw new ArgumentNullException(nameof(value));
    
                    Value = value;
                }
            }
    
            private class ObjectWithOptionalMessage
            {
                public Message? Message { get; }
    
                public ObjectWithOptionalMessage(Message? message)
                {
                    Message = message;
                }
            }
    
            {
                var objWithMessage = new ObjectWithOptionalMessage(new Message("Hello!"));
    
                var serializerSettings = new JsonSerializerSettings
                {
                    TypeNameHandling = TypeNameHandling.Auto,
                    Formatting = Formatting.Indented
                };
    
                Console.WriteLine(JsonConvert.SerializeObject(objWithMessage, serializerSettings));
            }
        }
    }

Gives me the following output:

    {
      "Message": {
        "$type": "System.Nullable`1[[JsonPlayground.Program+Message, JsonPlayground]], System.Private.CoreLib",
        "Value": "Hello!"
      }
    }

And I can't think of a reason why `$type` should be included for what is clearly a well declared property. If I declare the `Message` property to not be nullable, the `$type` property is not included.

Why is that?

TylerBrinkley:
Fixes #1050

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/TypeNameHandlingTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/TypeNameHandlingTests.cs
@@ -1934,7 +1934,48 @@ namespace Newtonsoft.Json.Tests.Serialization
 
             Assert.IsNotNull(obtainedDictionary);
         }
+
+        [Test]
+        public void SerializeNullableStructProperty_Auto()
+        {
+            ObjectWithOptionalMessage objWithMessage = new ObjectWithOptionalMessage(new Message2("Hello!"));
+
+            JsonSerializerSettings serializerSettings = new JsonSerializerSettings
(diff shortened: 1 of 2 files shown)


JamesNK:
Perfect. Thanks!

PR #1056, opened 2016-10-22
Added possibility to ignore whitespaces during deserialization of enum values

hmvs:
For example now optionally we can deserialize such enums.  
```
{
  "Enum": "red,dark goldenrod, white"
}
```

We can often see such enums when deserializing objects from external (non .net projects). And every time we should think how to write custom converter that has all features of base one.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Converters/StringEnumConverterTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Converters/StringEnumConverterTests.cs
@@ -413,6 +413,52 @@ namespace Newtonsoft.Json.Tests.Converters
         }
 
         [Test]
+        public void DeserializeIgnoringWhitespaceEnum() 
+        {
+            string json = @"{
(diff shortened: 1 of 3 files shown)


JamesNK:
I'm not sure I agree with this. What about people who want to ignore other kinds of separating characters?

Why not place an EnumMember on the properties and specify the name with a space it in?

hmvs:
Seems EnumMember not supported in .net core version.
I haven't seen cases when we need something different than a space and even can't imagine. But space is common from what i have seen. Space symbol I met second time during two weeks working with different APIs ( I can send you the links)
But if you want I can reimplemented passing char or char array there or even array of strings. But again, I don't think it's common.

hmvs:
My fault. EnumMember works fine under .net core. I thought StringEnumConverter is enabled by default (because values without space converts just fine). 
So if you don't like this PR, but i think it's still useful (a lot of enums will be attribute free), you can close it.

PR #1064, opened 2016-10-31
Add a test for deserializing an IEnumerable with implicit conversion

hughbe:
The feature didn't seem to work on my PC, so I was going to implement the feature.

Turns out that implicit conversions are supported, and I had a typo in the product code.

Still, I'll send in some tests as I couldn;t find any else in the repo and so that I didn't entirely waste my time!


JamesNK:
Thanks!

PR #1065, opened 2016-10-31
Handle IEEE754 overflows as ±Infinity instead of raising errors

0xced:
Highlighted by @nst in [Parsing JSON is a Minefield 💣](http://seriot.ch/parsing_json.html#22):
> According to RFC 7159, "A JSON parser MUST accept all texts that conform to the JSON grammar" (section 9). However, according to the same paragraph, "An implementation may set limits on the range and precision of numbers."

Most parsers just handle overflows as ±Infinity.


0xced:
Any opinion on this @JamesNK?

JamesNK:
I ended up disabling the custom float parse logic so this change is no longer relevant.

0xced:
Actually, converting overflow to infinity could still happen by catching OverflowException. Would you consider another pull request to handle overflows as ±Infinity?

PR #1071, opened 2016-11-03
Make methods static where possible

hughbe:
Some minor gardening


hughbe:
@JamesNK friendly ping!

JonHanna:
`BinaryConverter.EnsureReflectionObject` and its memoisation field, `_reflectionObject` can also be made static, and replacing `value.GetType().AssignableToTypeName(BinaryTypeName)` with `value.GetType().FullName == BinaryTypeName` makes it clearer that it can, as well as being slightly cheaper.

JamesNK:
I'm not a huge fan of making things static just because you can but these mostly don't look too bad.

Could you double check there is nothing introduced with Async that doesn't also have static.

PR #1072, opened 2016-11-04
Fix return type for SByte and SByteNullable in ToObject

recycleddan:
Looking at the code, I noticed that SByte cast to (sbyte?) and SByteNullable cast to (sbyte). This seems backwards and in conflict with all the other cases.


JamesNK:
Good spotting. Thanks

PR #1073, opened 2016-11-05
Fix null reference error when deserializing bytes

Rantanen:
If the JSON contained empty object "{}" for a byte[] property, the `JsonReader` ended up trying to read the object content despite the fact that the content didn't exist. This caused a null reference exception.

The fix ensures the value exists before trying to read it. If the value doesn't exist, the execution ends up in "Unexpected token" exception with way more informative error message.

-------

There was a test that could have caught this. Unfortunately passing just `{` to the `JsonTextReader` causes the text reader to fail early while it's parsing the object and the execution never attempts to read the `$type` property.


JamesNK:
Looks good to me! Thanks

Issue #830, opened 2016-03-02
CamelCasePropertyNamesContractResolver uses obsolete shareCache option


TFleury:
Issue described in #321, but closed by author without being solved.

If you inherit a custom `ContractResolver` from `CamelCasePropertyNamesContractResolver`, which is intended to just change property names casing, shared cache is forced.

JamesNK:
It's not good but changing it would be a breaking change.

TFleury:
Can you at least provide a way to disable shared cache without changing default behavior ?
Changing access modifier of `_sharedCache` to `protected` would allow overriding this behavior in a derived ContractResolver constructor without introducing a breaking change.

pedrofelixsky:
Agree. The default static shared caching behavior is problematic in composition scenarios where a composite resolver is dispatching to different resolver based on type characteristics.
At least, the `CamelCasePropertyNamesContractResolver` should provide a ctor that disables shared caching.

heaths:
Just got bit by this ourselves. It took a bit to track down, and offering a constructor overload would also make it more obvious - not to mention allow us to extend `CamelCasePropertyNamesContractResolver` without caching.

AArnott:
This was closed based on the idea that it would be a breaking change, but in fact #1078 proves it can be done without a breaking change. But that PR was rejected, because it *used to be possible* another way. Now that way is also cut off. What are we supposed to do so that we don't have to define a specialized type for every single variation we might need at runtime for this?

PR #1078, opened 2016-11-11
Allow separate caches for camel case resolver

heaths:
Fixes issue #830 by allowing developers to specify whether to use cached state. The default is `true` as before so there should be no breaking change.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/CamelCasePropertyNamesContractResolverTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/CamelCasePropertyNamesContractResolverTests.cs
@@ -223,5 +223,23 @@ namespace Newtonsoft.Json.Tests.Serialization
   ""second"": ""Value2!""
 }", json);
         }
+
+        [Test]
+        public void SharedState()
+        {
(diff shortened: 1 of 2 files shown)


JamesNK:
Hi

I don't want to add this PR because you can achieve the same thing with a `DefaultContractResolver` plus `CamelCaseNamingStrategy`. The camel case resolver will probably become obsolete sometime in the future.

AArnott:
I see you've even made the `DefaultContractResolver.ctor(bool)` constructor obsolete. So what is the recommended way to do this going forward?

AArnott:
Gah! You actually *removed* that ctor in v10. So now we have no story to support this, do we?

JamesNK:
I got rid of having a shared cache inside a contract resolver because it wasn't intuitive.

I only kept it for the camel case contract resolver because tons of people new up a camel case resolver every serialization. If the shared cache was removed then a lot of apps will suddenly perform significantly worse.

If you don't want a shared cache then use DefaultContractResolver + CamelCaseNamingStrategy.

AArnott:
Oh, sweet. The cache move totally makes sense. Thanks.

PR #1081, opened 2016-11-19
Cache results when splitting fully qualified type names

### LINKED ISSUE #1079: Consider caching for ReflectionUtils.SplitFullyQualifiedTypeName
It seems that when we use fully qualified names, SplitFullyQualifiedTypeName contributes a lot of memory allocations for each object deserialization, which adds some stress to the GC.

Is there a reason why the code won't cache the mapping between the fullyQualifiedTypeName string and the TypeNameKey result? it will save quite a bit of string manipulations & memory allocation.

Normally types we deserialize will return over and over again, and are not likely to disappear during runtime.
The memory footprint of the cache should be relatively small compared to the continues allocations.. 

I ran a simple test with a small code change:
for non portable code, added a ConcurrentDictionary from fullyQualifiedTypeName string to TypeNameKey result, and made these allocations completely disappear from PerfView after running a few ms. 

I wonder how such change will be accepted if I submit a pull request for it...

i3arnon:
Solves #1079.

In high performance scenarios when deserializing json objects that use $type the `SplitFullyQualifiedTypeName` method allocates a huge amount of strings to be GCed, although the types themselves are almost always the same.

This change caches the results instead of generating them again and again for each deserialization.

### DIFF
--- a/Src/Newtonsoft.Json/Utilities/ReflectionUtils.cs
+++ b/Src/Newtonsoft.Json/Utilities/ReflectionUtils.cs
@@ -85,6 +85,8 @@ namespace Newtonsoft.Json.Utilities
     {
         public static readonly Type[] EmptyTypes;
 
+        private static readonly ThreadSafeStore<string, TypeNameKey> FullyQualifiedTypeNameKeyCache = new ThreadSafeStore<string, TypeNameKey>(CreateFullyQualifiedTypeNameKey);
+
         static ReflectionUtils()
         {
 #if !(PORTABLE40 || PORTABLE)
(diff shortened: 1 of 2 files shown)


JamesNK:
Hi

The problem with your solution is that any content in $type will be cached in a static variable that will grow forever. It would be a memory leak that someone could use to pass random values to until the application runs out of memory.

i3arnon:
@JamesNK that's very reasonable. In our product we have a construct that ages over time (and is capped by size) to solve that.

I haven't found something similar in the existing code, but I could easily set a timer to replace the dictionary every interval (10 minutes?). Is that behavior acceptable?

JamesNK:
In 10 minutes someone malicious could still easily pass enough data to exhaust a server's memory. Any caching would need to be a limit to the last N $type values. I'm not sure there is a thread safe data structure to do that that would be available in .NET 2.0 and PCL targets.

i3arnon:
@JamesNK the limit can be baked right into the `ThreadSafeStore` so it will only cache up to a max size and after that just invoke the creator.

I'll also take a look at the `SerializationBinder`. I wasn't aware of that option.

JamesNK:
I don't want to add this but I improved allocations here https://github.com/JamesNK/Newtonsoft.Json/commit/ba04246a16c66c532598df3f072f7e7ccba7d5ac

PR #1083, opened 2016-11-21
Base protected property replaces child protected property

sove001:
Look's like a bug

### DIFF
--- a/Src/Newtonsoft.Json/Utilities/ReflectionUtils.cs
+++ b/Src/Newtonsoft.Json/Utilities/ReflectionUtils.cs
@@ -1017,7 +1017,7 @@ namespace Newtonsoft.Json.Utilities
                 {
                     PropertyInfo subTypeProperty = propertyInfo;
 
-                    if (!IsPublic(subTypeProperty))
+                    if (IsPublic(subTypeProperty))
                     {
                         // have to test on name rather than reference because instances are different
                         // depending on the type that GetProperties was called on


JamesNK:
Hi

Could you give an example of the bug in action? I only merge changes that have a unit test that illustrates the change.

sove001:
Hi, James!
As you asked bug in action:

    public class BaseClass
    {
        protected virtual string MyProperty { get; set; }
    }

    public class ChildClass : BaseClass
    {
        [JsonProperty]
        protected override string MyProperty { get; set; }

        public string GetPropertyValue()
        {
            return MyProperty;
        }
    }

    public class CustomContractResolver : DefaultContractResolver
    {
        protected override IList<JsonProperty> CreateProperties(Type type, MemberSerialization memberSerialization)
        {
            var jsonProperties = base.CreateProperties(type, memberSerialization);

            Assert.AreEqual(1, jsonProperties.Count, "MyProperty should be founded");

            return jsonProperties;
        }
    }

    [TestClass]
    public class JsonTest
    {
        [TestMethod]
        public void Test1()
        {
            var propertyValue = "test";
            var testJson = @"{ 'MyProperty' : '" + propertyValue + "' }";

            var testObject = JsonConvert.DeserializeObject<ChildClass>(testJson, new JsonSerializerSettings
            {
                ContractResolver = new CustomContractResolver()
            });

            Assert.AreEqual(propertyValue, testObject.GetPropertyValue(), "MyProperty should be populated");
        }
    }

JamesNK:
Hi

I've fixed this but your PR created some failing tests and I used a different approach.

https://github.com/JamesNK/Newtonsoft.Json/commit/5269ce81c0ea4fbfa141cc886fc120e23b21625e

Issue #1097, opened 2016-12-10
Clean up XML documentation comments


stakx:
I was working with Json.NET earlier today and a typo showed up in an IntelliSense tooltip. I thought I'd take a look at the source code's XML documentation comments and see if there were any more spelling errors or irregularities.

In general, the XML documentation comments are pretty good. I'll enter a pull request (#1098) that fixes several details that I expect don't require any further discussion.

There are a few points that might require a decision by the code owners before any changes are made to the comments:

* There's inconsistency in how other code elements are referred to. Sometimes they are referenced with `<see cref="…"/>`, sometimes using code formatting `<c>…</c>`; sometimes no markup is applied at all. The proper solution would seem to use `<see cref="…"/>` everywhere where it can be applied. (In VS 2015, this triggers syntax highlighting in IntelliSense tooltips. But it makes the comments horrible in the Object Browser. No idea what difference it makes with Sandcastle help files etc.)

* Since XML documentation comments are not inherited from base classes, there are quite a few copy & paste errors. Perhaps at some point, changes in base classes' / interfaces' comments should also be applied to inherited / implementing types so that the documentations match up again.

* The results of the JSON de-/serialization process are described with different terms. I might possibly be lacking understanding of Json.NET's architecture to appreciate these differences; but it appears to me that vocabulary such as "JSON object", "JSON text", "JSON string", is used interchangeably. In order to avoid any unnecessary confusion, it might be a good idea to decide on one term and use it consistently throughout the whole code base.

* No conceptual distinction is made between `TextReader` & `TextWriter` and `Stream`. The documentation comments assume that the former are backed by `Stream`, which might be true in the general case, but it is not true by necessity (see e.g. `StringReader` & `StringWriter`). Json.NET appears to deal primarily with `TextReader` and `TextWriter`, so inappropriate references to underlying `Stream` objects (which might not actually exist at all!) should be carefully removed.

* Null references are mentioned in an inconsistent manner. I've seen `null`, `<c>null</c>`, `null reference`, and `null value`. It's a detail, but one that could potentially confuse non-C# programmers. (For example, in VB.NET, `null` doesn't mean anything; The null reference is referred to as `Nothing`.) Any confusion could be ruled out by always writing out "null reference" (but that'd often sound a little clunky).

[...]

PR #1098, opened 2016-12-10
Fix minor issues in XML documentation comments

stakx:
This fixes several minor issues in the code's documentation comments, such as spelling errors, missing punctuation and references to other code elements, and a few other small inconsistencies.

This commit deals only with fairly straight-forward stuff, but more could be done to clean up the XML documentation comments. See issue #1097 for more details.

### DIFF
--- a/Src/Newtonsoft.Json/JsonWriter.cs
+++ b/Src/Newtonsoft.Json/JsonWriter.cs
@@ -332,7 +332,7 @@ namespace Newtonsoft.Json
         }
 
         /// <summary>
-        /// Initializes a new instance of a <c>JsonWriter</c>-derived class.
+        /// Initializes a new instance of the <c>JsonWriter</c> class.
         /// </summary>
         protected JsonWriter()
         {
@@ -1424,9 +1424,6 @@ namespace Newtonsoft.Json
             InternalWriteWhitespace(ws);
         }
 
-        /// <summary>
-        /// Releases both unmanaged and managed resources.
-        /// </summary>
         void IDisposable.Dispose()
         {
             Dispose(true);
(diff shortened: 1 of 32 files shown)


stakx:
Umm, I believe that build didn't fail because of any changes I made (which was only to code comments), but because of the unit test project (which appears to contain broken references to NUnit). I hope this is a known issue.

stakx:
I'd be surprised if the CI build fails because of any of the changes in this pull request, since only code comments are affected. If there *is* something I need to do to resolve this build issue, please let me know.

Src/Newtonsoft.Json/JsonWriter.cs
```diff
@@ -1416,22 +1416,25 @@ public virtual void WriteComment(string text)
...
         /// </summary>
         /// <param name="ws">The string of white space characters.</param>
         public virtual void WriteWhitespace(string ws)
         {
             InternalWriteWhitespace(ws);
         }
 
+        /// <summary>
+        /// Releases both unmanaged and managed resources.
+        /// </summary>
```
JamesNK: Explicitly implemented interfaces don't need comments.

Src/Newtonsoft.Json/JsonWriter.cs
```diff
@@ -332,7 +332,7 @@ public CultureInfo Culture
         }
 
         /// <summary>
-        /// Creates an instance of the <c>JsonWriter</c> class. 
+        /// Initializes a new instance of a <c>JsonWriter</c>-derived class.
```
JamesNK: Hmm, is this how protected base constructors are generally documented on MSDN? I'd like to follow MS's lead in commenting style.

Src/Newtonsoft.Json/JsonWriter.cs
stakx: Will do.

stakx:
I'll leave the more substantial changes to the new few days as they'll require more time and possibly some research how they document things on MSDN.

JamesNK:
Looks good to me. Thanks for cleaning the comments up. A lot of embarrassing typos removed.

ghuntley:
Beautiful

PR #1100, opened 2016-12-13
Improve JsonConvert.ToString(object) error message

hughbe:
We had this error on one of our projects, turns out `JsonConvert.ToString()` was used instead of `JsonConvert.SerializeObject`


JamesNK:
Thanks but I don't want to mention SerializeObject whenever JsonSerializer is mentioned. Instead one of the first Json.NET documentation pages says that SerializeObject is a helper method over the top of the JsonSerializer.

PR #1106, opened 2016-12-19

stakx:
This one makes the formatting of `true`, `false`, and `null` consistent where they refer to the C# keywords.

There are a few instances of "null" where a different kind of nullity is or might be referred to (e.g. "null tokens", which apparently include those with token types `JsonToken.Null` or `JsonToken.None`). These remain as they are.

I also collapsed the rarely used phrases "null reference", "null value", and "null (Nothing in Visual Basic)" simply into `null`. (Regarding the last one, it doesn't make much sense to only explain `null` to VB.NET programmers in approx. 5 out of 100 cases.)

### DIFF
--- a/Src/Newtonsoft.Json/JsonTextReader.cs
+++ b/Src/Newtonsoft.Json/JsonTextReader.cs
@@ -445,7 +445,7 @@ namespace Newtonsoft.Json
         /// <summary>
         /// Reads the next JSON token from the stream as a <see cref="Byte"/>[].
         /// </summary>
-        /// <returns>A <see cref="Byte"/>[] or a null reference if the next JSON token is null. This method will return <c>null</c> at the end of an array.</returns>
+        /// <returns>A <see cref="Byte"/>[] or <c>null</c> if the next JSON token is null. This method will return <c>null</c> at the end of an array.</returns>
         public override byte[] ReadAsBytes()
         {
             EnsureBuffer();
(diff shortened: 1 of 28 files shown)


Src/Newtonsoft.Json/JsonTextReader.cs
```diff
@@ -445,7 +445,7 @@ public override string ReadAsString()
         /// <summary>
         /// Reads the next JSON token from the stream as a <see cref="Byte"/>[].
         /// </summary>
-        /// <returns>A <see cref="Byte"/>[] or a null reference if the next JSON token is null. This method will return <c>null</c> at the end of an array.</returns>
+        /// <returns>A <see cref="Byte"/>[] or <c>null</c> if the next JSON token is null. This method will return <c>null</c> at the end of an array.</returns>
```
stakx: Here's an instance of "null" that I chose not to markup with `<c>`. I'm not entirely sure about Json.NET's own concept of nullity, but after running a quick test, I suspect that this phrase "if the next JSON token is null" actually means "if the next JSON token has a type of either `JsonToken.Null` or `JsonToken.None`".

JamesNK: Looks good to me. I'm happy sticking with `<c>null</c>` vs "JSON token is null" to differentiate the two concepts.

PR #1109, opened 2016-12-22

stakx:
This PR is mostly about adding missing / fixing existing `<see cref="…"/>` references. See commit message for some more detail. I also smuggled in some more typo corrections and minor `<c>` reformattings.

### DIFF
--- a/Src/Newtonsoft.Json/Converters/EntityKeyMemberConverter.cs
+++ b/Src/Newtonsoft.Json/Converters/EntityKeyMemberConverter.cs
@@ -32,7 +32,7 @@ using Newtonsoft.Json.Utilities;
 namespace Newtonsoft.Json.Converters
 {
     /// <summary>
-    /// Converts an Entity Framework <c>EntityKey</c> to and from JSON.
+    /// Converts an Entity Framework <see cref="T:System.Data.EntityKeyMember"/> to and from JSON.
     /// </summary>
     public class EntityKeyMemberConverter : JsonConverter
     {
--- a/Src/Newtonsoft.Json/Converters/JavaScriptDateTimeConverter.cs
+++ b/Src/Newtonsoft.Json/Converters/JavaScriptDateTimeConverter.cs
@@ -30,7 +30,7 @@ using Newtonsoft.Json.Utilities;
 namespace Newtonsoft.Json.Converters
 {
     /// <summary>
-    /// Converts a <see cref="DateTime"/> to and from a JavaScript date constructor (e.g. new Date(52231943)).
+    /// Converts a <see cref="DateTime"/> to and from a JavaScript <c>Date</c> constructor (e.g. <c>new Date(52231943)</c>).
     /// </summary>
     public class JavaScriptDateTimeConverter : DateTimeConverterBase
     {
--- a/Src/Newtonsoft.Json/FormatterAssemblyStyle.cs
+++ b/Src/Newtonsoft.Json/FormatterAssemblyStyle.cs
@@ -9,12 +9,12 @@ namespace System.Runtime.Serialization.Formatters
     public enum FormatterAssemblyStyle
     {
         /// <summary>
-        /// In simple mode, the assembly used during deserialization need not match exactly the assembly used during serialization. Specifically, the version numbers need not match as the <c>LoadWithPartialName</c> method is used to load the assembly.
+        /// In simple mode, the assembly used during deserialization need not match exactly the assembly used during serialization. Specifically, the version numbers need not match as the <see cref="M:System.Reflection.Assembly.LoadWithPartialName(String)"/> method is used to load the assembly.
         /// </summary>
         Simple = 0,
 
         /// <summary>
-        /// In full mode, the assembly used during deserialization must match exactly the assembly used during serialization. The Load method of the Assembly class is used to load the assembly.
+        /// In full mode, the assembly used during deserialization must match exactly the assembly used during serialization. The <see cref="System.Reflection.Assembly.Load"/> is used to load the assembly.
         /// </summary>
         Full = 1
     }
--- a/Src/Newtonsoft.Json/JsonReader.cs
+++ b/Src/Newtonsoft.Json/JsonReader.cs
@@ -51,7 +51,7 @@ namespace Newtonsoft.Json
         protected internal enum State
         {
             /// <summary>
-            /// The <see cref="Read"/> method has not been called.
+            /// A <see cref="JsonReader"/> read method has not been called.
             /// </summary>
             Start,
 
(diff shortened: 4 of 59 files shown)


Src/Newtonsoft.Json/Converters/EntityKeyMemberConverter.cs
```diff
@@ -32,7 +32,7 @@
 namespace Newtonsoft.Json.Converters
 {
     /// <summary>
-    /// Converts an Entity Framework <c>EntityKey</c> to and from JSON.
+    /// Converts an Entity Framework <see cref="T:System.Data.EntityKeyMember"/> to and from JSON.
```
stakx: The member ID syntax needs to stay in this instance, since the referred type is in an assembly (System.Data.Entity.dll) that isn't referenced by default.

Src/Newtonsoft.Json/FormatterAssemblyStyle.cs
```diff
@@ -9,12 +9,12 @@ namespace System.Runtime.Serialization.Formatters
     public enum FormatterAssemblyStyle
     {
         /// <summary>
-        /// In simple mode, the assembly used during deserialization need not match exactly the assembly used during serialization. Specifically, the version numbers need not match as the <c>LoadWithPartialName</c> method is used to load the assembly.
+        /// In simple mode, the assembly used during deserialization need not match exactly the assembly used during serialization. Specifically, the version numbers need not match as the <see cref="M:System.Reflection.Assembly.LoadWithPartialName(String)"/> method is used to load the assembly.
```
stakx: Member ID syntax is also necessary here. The referenced method is not actually present in all targeted platforms. Perhaps a target-specific `#if` block is needed here to have target-specific comments?

Src/Newtonsoft.Json/JsonReader.cs
```diff
@@ -51,7 +51,7 @@ public abstract class JsonReader : IDisposable
         protected internal enum State
         {
             /// <summary>
-            /// The <see cref="Read"/> method has not been called.
+            /// A <see cref="JsonReader"/> read method has not been called.
```
stakx: This is one of three similar comments that appear on these code elements:

* `JsonReader.State.Start`
* `WriteState.Start`
* `JsonToken.None`

The thing here is that a whole group of methods is referred to. That's why I changed "the" to "a" and removed the reference to one particular method (even though that appears to be the one that always gets called internally).

Src/Newtonsoft.Json/Converters/JavaScriptDateTimeConverter.cs
```diff
@@ -30,7 +30,7 @@
 namespace Newtonsoft.Json.Converters
 {
     /// <summary>
-    /// Converts a <see cref="DateTime"/> to and from a JavaScript date constructor (e.g. new Date(52231943)).
+    /// Converts a <see cref="DateTime"/> to and from a JavaScript <c>Date</c> constructor (e.g. <c>"new Date(52231943)"</c>).
```
JamesNK: There shouldn't be quotes around the new Date. It isn't written to JSON as a string

Src/Newtonsoft.Json/JsonReader.cs
JamesNK: Back in the day there was only one Read method :)

Src/Newtonsoft.Json/Linq/JArray.cs
```diff
@@ -302,25 +302,25 @@ public int IndexOf(JToken item)
         }
 
         /// <summary>
-        /// Inserts an item to the <see cref="T:System.Collections.Generic.IList`1"/> at the specified index.
+        /// Inserts an item to the <see cref="IList{T}"/> at the specified index.
```
JamesNK: Should this be `<see cref="JArray"/>`? The original comments are inherited from `IList<T>`.

Can be repeated for other `IList<T>` crefs around here.

Src/Newtonsoft.Json/Linq/JArray.cs
```diff
@@ -340,30 +340,30 @@ public IEnumerator<JToken> GetEnumerator()
 
         #region ICollection<JToken> Members
         /// <summary>
-        /// Adds an item to the <see cref="T:System.Collections.Generic.ICollection`1"/>.
+        /// Adds an item to the <see cref="ICollection{T}"/>.
         /// </summary>
-        /// <param name="item">The object to add to the <see cref="T:System.Collections.Generic.ICollection`1"/>.</param>
-        /// <exception cref="T:System.NotSupportedException">The <see cref="T:System.Collections.Generic.ICollection`1"/> is read-only.</exception>
+        /// <param name="item">The object to add to the <see cref="ICollection{T}"/>.</param>
+        /// <exception cref="NotSupportedException">The <see cref="ICollection{T}"/> is read-only.</exception>
```
JamesNK: JArray isn't readonly so this line can be removed

Src/Newtonsoft.Json/Linq/JArray.cs
```diff
@@ -340,30 +340,30 @@ public IEnumerator<JToken> GetEnumerator()
...
         {
             Add((object)item);
         }
 
         /// <summary>
-        /// Removes all items from the <see cref="T:System.Collections.Generic.ICollection`1"/>.
+        /// Removes all items from the <see cref="ICollection{T}"/>.
         /// </summary>
-        /// <exception cref="T:System.NotSupportedException">The <see cref="T:System.Collections.Generic.ICollection`1"/> is read-only. </exception>
+        /// <exception cref="NotSupportedException">The <see cref="ICollection{T}"/> is read-only. </exception>
```
JamesNK: Never readonly.

Src/Newtonsoft.Json/Serialization/JsonObjectContract.cs
```diff
@@ -85,8 +85,8 @@ public JsonPropertyCollection CreatorParameters
 
         /// <summary>
         /// Gets or sets the override constructor used to create the object.
-        /// This is set when a constructor is marked up using the
-        /// JsonConstructor attribute.
+        /// This is set when a constructor is marked up using
+        /// [<see cref="JsonConstructorAttribute"/>].
```
JamesNK: What are the square brackets for?

stakx: Some Christmas tree decoration for Json.NET. :) Seriously, though, I wasn't quite sure how to rephrase the duplication of the word "attribute" after introducing the cref, so I figured it might be smart to have a semi-code example right inside the comment. But since that isn't immediately obvious it's probably a bad idea after all. I'll remove the brackets.

Src/Newtonsoft.Json/Linq/JArray.cs
stakx: Agreed. Same goes for `ICollection<T>` crefs.

Src/Newtonsoft.Json/Converters/JavaScriptDateTimeConverter.cs
stakx: Oops... my bad, sorry.

JamesNK:
Merged!

PR #1111, opened 2016-12-24

stakx:
As discussed in #1097, something could be done about the (often incorrect) mentions of "underlying streams" in the doc comments of `JsonReader`, `JsonWriter`, and derived classes.

In the abstract base classes `JsonReader` and `JsonWriter`, "stream" gets replaced by the generic terms "source" and "destination". The word "underlying" isn't added here since it would give a somewhat clunky sound to many of the comments, and it's not really needed here to distinguish between the type itself and the type it operates on.

In derived reader / writer types, crefs to the concrete underlying types are introduced.

Finally, the `Close` methods' documentation is more explicit about when the underlying source / destination is closed as well.

### DIFF
--- a/Src/Newtonsoft.Json/Bson/BsonReader.cs
+++ b/Src/Newtonsoft.Json/Bson/BsonReader.cs
@@ -122,7 +122,7 @@ namespace Newtonsoft.Json.Bson
         /// <summary>
         /// Initializes a new instance of the <see cref="BsonReader"/> class.
         /// </summary>
-        /// <param name="stream">The stream.</param>
+        /// <param name="stream">The <see cref="Stream"/> containing the BSON data to read.</param>
         public BsonReader(Stream stream)
             : this(stream, false, DateTimeKind.Local)
         {
@@ -131,7 +131,7 @@ namespace Newtonsoft.Json.Bson
         /// <summary>
         /// Initializes a new instance of the <see cref="BsonReader"/> class.
         /// </summary>
-        /// <param name="reader">The reader.</param>
+        /// <param name="reader">The <see cref="BinaryReader"/> containing the BSON data to read.</param>
         public BsonReader(BinaryReader reader)
             : this(reader, false, DateTimeKind.Local)
         {
@@ -140,7 +140,7 @@ namespace Newtonsoft.Json.Bson
         /// <summary>
         /// Initializes a new instance of the <see cref="BsonReader"/> class.
         /// </summary>
-        /// <param name="stream">The stream.</param>
+        /// <param name="stream">The <see cref="Stream"/> containing the BSON data to read.</param>
         /// <param name="readRootValueAsArray">if set to <c>true</c> the root object will be read as a JSON array.</param>
         /// <param name="dateTimeKindHandling">The <see cref="DateTimeKind" /> used when reading <see cref="DateTime"/> values from BSON.</param>
         public BsonReader(Stream stream, bool readRootValueAsArray, DateTimeKind dateTimeKindHandling)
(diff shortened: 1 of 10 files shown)


Src/Newtonsoft.Json/Bson/BsonWriter.cs
```diff
@@ -72,15 +72,15 @@ public BsonWriter(Stream stream)
...
+        /// <param name="writer">The <see cref="BinaryWriter"/> to write to.</param>
         public BsonWriter(BinaryWriter writer)
         {
             ValidationUtils.ArgumentNotNull(writer, nameof(writer));
             _writer = new BsonBinaryWriter(writer);
         }
 
         /// <summary>
-        /// Flushes whatever is in the buffer to the underlying streams and also flushes the underlying stream.
+        /// Flushes whatever is in the buffer to the underlying <see cref="BinaryWriter"/> or <see cref="Stream"/> and also flushes the underlying writer or stream.
```
JamesNK: A BinaryWriter can only wrap a stream so I don't think it needs to be mentioned. Repeat in other comments.

Src/Newtonsoft.Json/Bson/BsonReader.cs
```diff
@@ -175,7 +175,7 @@ private string ReadElement()
         }
 
         /// <summary>
-        /// Reads the next JSON token from the stream.
+        /// Reads the next JSON token from the underlying <see cref="BinaryReader"/> or <see cref="Stream"/>.
```
JamesNK: A BinaryReader can only wrap a stream so I don't think both need to be mentioned, just stream. Repeat in other comments that mention both.

Src/Newtonsoft.Json/Linq/JTokenWriter.cs
```diff
@@ -88,15 +88,18 @@ public JTokenWriter()
         }
 
         /// <summary>
-        /// Flushes whatever is in the buffer to the underlying streams and also flushes the underlying stream.
+        /// Flushes whatever is in the buffer to the underlying <see cref="JContainer"/>.
```
stakx: While it seemed safe to remove the statement about the underlying stream being flushed (since a `JContainer` *cannot* be flushed) I wasn't so sure about the first half of the sentence; you might introduce buffering for `JTokenWriter` in the future, so I left this as is.

Src/Newtonsoft.Json/Bson/BsonReader.cs
```diff
@@ -224,6 +224,7 @@ public override bool Read()
 
         /// <summary>
         /// Changes the reader's state to <see cref="JsonReader.State.Closed"/>.
```
stakx: Btw., working on this issue has made me wonder whether "Changes the reader's state to State.Closed" ought to be simplified to "Closes this reader" (like it is for the writer classes already).

Src/Newtonsoft.Json/Bson/BsonReader.cs
stakx: Please confirm this again. Internally, `BsonReader` always uses a `BinaryReader`. But outwardly, things look different: A user of the `BsonReader` class sees two public constructors, one accepting a `Stream`, the other accepting a `BinaryReader`. So there's a choice, and I tried to keep the comments in line with that.

But if you say so, I can change these comments to refer only to one or the other. (In that case, I'd suggest choosing `BinaryReader`.)

stakx: I see. I thought it would make sense to have the comments in line with the public interface (one constructor accepting a `Stream`, another accepting a `BinaryReader`), but it indeed makes them somewhat unwieldy. Changed.

Issue #1112, opened 2016-12-26
JsonValidatingReader does not close underlying JsonReader when CloseInput is true


stakx:
I realise that `JsonValidatingReader` is `[Obsolete]` and probably low priority, but if you're still willing to bugfix that class:

`JsonValidatingReader` operates on top of another `JsonReader`, yet it is currently lacking an override of the `Close` method that would close the underlying reader when `CloseInput` is set to `true`:

```csharp
public override void Close()
{
    base.Close();
    if (CloseInput && _reader != null)
    {
        _reader.Close();
    }
}
```

Note that such an override could be a minor breaking change if someone erroneously relied on `JsonValidatingReader` never closing its underlying reader. (This could be possibly mitigated by changing the default value of `JsonValidatingReader.CloseInput` to `false`, which would be closer to its current behavior, but this would also bear the potential for confusion.)

PR #1113, opened 2016-12-26
Make JsonValidatingReader close underlying JsonReader when CloseInput is true

stakx:
This would take care of issue #1112.


JamesNK:
Looks good to me. Thanks

PR #1119, opened 2016-12-30
Add support for asynchronous operations.

### LINKED ISSUE #650: Async methods for JsonTextReader / JsonTextWriter

JonHanna:
Closes #650 

Default base-class behaviour is to execute the associated synchronous method and return a completed task.

Similarly to `StreamReader` and `StreamWriter` in the case of classes derived from those offering asynchronous operations, the derived class itself must override the async method or it will operate synchronously, as it's not possible otherwise to match the overridden behaviour of virtual methods. An exception is `CustomCreationConverter` which explicitly checks if `ReadJson` was overridden as the common use would have `Create` (safely) overridden but not `ReadJson`.

Having separate files for all the async code was originally intended only to be a temporary method during development, but as it offers convenience in comparing methods I left them like that.

Possible future work:

Commonality between synch and asynch methods could be refactored into separate methods, but doing this immediately would lead to a more confusing PR with some of the refactoring being rather opaque if combined with this work.

Asynch support for BSON. Requires asynch support that the framework `BinaryReader` and `BinaryWriter` doesn't provide, which would impact considerably on the current implementation, so left out of this PR. (As the BSON-related classes do inherit the default synchronously-returning methods from the base, there are still tests added for those methods)

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Bson/BsonReaderAsyncTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Bson/BsonReaderAsyncTests.cs
@@ -1,0 +1,1768 @@
+﻿#region License
+// Copyright (c) 2007 James Newton-King
+//
+// Permission is hereby granted, free of charge, to any person
+// obtaining a copy of this software and associated documentation
+// files (the "Software"), to deal in the Software without
+// restriction, including without limitation the rights to use,
+// copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following
+// conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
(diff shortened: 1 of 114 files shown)


JonHanna:
> Similarly to StreamReader and StreamWriter in the case of classes derived from those offering asynchronous operations, the derived class itself must override the the async method or it will operate synchronously, as it's not possible otherwise to match the overridden behaviour of virtual methods.

Oh. There are also some `internal` derived classes that skip this checking and hit the asynchronous path directly, used wherever the project uses the class internally.

JamesNK:
Wow that's a lot of changes. Can we break it down to just the reader/writer async changes? Once those are merged then we can move onto JToken loading. Right now I'm honestly not sure whether I want JsonSerializer to be async because of all the duplication, but once the other parts of merged I'd like to look at it by itself.

JonHanna:
Sure. I'll submit a PR soon with the base `JsonReader` and `JsonWriter` having the new methods and with `JsonTextReader` and `JsonTextWriter` implementing them fully, though personally it's the ability to serialise to and deserialise from a network stream asynchronously that I miss as a user.

PR #1122, opened 2016-12-31
Asynchronous methods in readers and writers.

JonHanna:
Contributes to #650

Extracted from #1119 but only covering the async methods on `JsonReader` and `JsonWriter` and implementation for `JsonTextReader` and `JsonTextWriter`.


JamesNK:
I see you have `ConfigureAwait(false)` in the unit tests. Is that appropriate? I have seen people talk about xunit having a synchronization context.

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,2327 @@
...
+
+            OnNewLine(_charPos);
+        }
+
+        private async Task<char> ParseUnicodeAsync(CancellationToken cancellationToken)
+        {
+            char writeChar;
+            if (await EnsureCharsAsync(4, true, cancellationToken).ConfigureAwait(false))
+            {
+                char hexChar = Convert.ToChar(ConvertUtils.HexTextToInt(_chars, _charPos, _charPos + 4));
```
JamesNK: Can move this plus the next 3 lines into a method that can be used by both `ParseUnicode`s

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,2327 @@
...
+                    _quoteChar = quote;
+                    break;
+                case ReadType.ReadAsInt32:
+                case ReadType.ReadAsDecimal:
+                case ReadType.ReadAsBoolean:
+
+                    // caller will convert result
+                    break;
+                default:
+                    if (_dateParseHandling != DateParseHandling.None)
```
JamesNK: I don't see any awaits inside this default. Can its code be moved to a shared method?

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,2327 @@
...
+
+        private async Task<bool> MatchValueAsync(string value, CancellationToken cancellationToken)
+        {
+            if (!await EnsureCharsAsync(value.Length - 1, true, cancellationToken).ConfigureAwait(false))
+            {
+                _charPos = _charsUsed;
+                throw CreateUnexpectedEndException();
+            }
+
+            for (int i = 0; i < value.Length; i++)
```
JamesNK: Can the rest of the code in this method be shared between sync and async?

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,2327 @@
...
+            {
+                throw JsonReaderException.Create(this, "Unexpected content while parsing JSON.");
+            }
+        }
+
+        private async Task<object> ParseNumberNaNAsync(ReadType readType, CancellationToken cancellationToken)
+        {
+            if (await MatchValueWithTrailingSeparatorAsync(JsonConvert.NaN, cancellationToken).ConfigureAwait(false))
+            {
+                switch (readType)
```
JamesNK: Can the rest of the code in this method be shared between sync and async?

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,2327 @@
...
+            }
+
+            throw JsonReaderException.Create(this, "Error parsing NaN value.");
+        }
+
+        private async Task<object> ParseNumberPositiveInfinityAsync(ReadType readType, CancellationToken cancellationToken)
+        {
+            if (await MatchValueWithTrailingSeparatorAsync(JsonConvert.PositiveInfinity, cancellationToken).ConfigureAwait(false))
+            {
+                switch (readType)
```
JamesNK: Can the rest of the code in this method be shared between sync and async?

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,2327 @@
...
+            }
+
+            throw JsonReaderException.Create(this, "Error parsing Infinity value.");
+        }
+
+        private async Task<object> ParseNumberNegativeInfinityAsync(ReadType readType, CancellationToken cancellationToken)
+        {
+            if (await MatchValueWithTrailingSeparatorAsync(JsonConvert.NegativeInfinity, cancellationToken).ConfigureAwait(false))
+            {
+                switch (readType)
```
JamesNK: Can the rest of the code in this method be shared between sync and async?

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,2327 @@
...
+        {
+            ShiftBufferIfNeeded();
+
+            char firstChar = _chars[_charPos];
+            int initialPosition = _charPos;
+
+            await ReadNumberIntoBufferAsync(cancellationToken).ConfigureAwait(false);
+
+            // set state to PostValue now so that if there is an error parsing the number then the reader can continue
+            SetPostValueState(true);
```
JamesNK: Can the rest of the code in this method be shared between sync and async?

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,2327 @@
...
+                    case '4':
+                    case '5':
+                    case '6':
+                    case '7':
+                    case '8':
+                    case '9':
+                        charPos++;
+                        break;
+                    default:
+                        _charPos = charPos;
```
JamesNK: Move code in default into a shared method?

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,2327 @@
...
+                                throw JsonReaderException.Create(this, "Unexpected end while parsing unquoted property name.");
+                            }
+
+                            break;
+                        }
+
+                        _stringReference = new StringReference(_chars, initialPosition, _charPos - initialPosition);
+                        return;
+                    default:
+                        char currentChar = _chars[_charPos];
```
JamesNK: Move code in default into shared method?

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,2327 @@
...
+                            _charPos++;
+                            break;
+                        }
+                        else if (char.IsWhiteSpace(currentChar) || currentChar == ':')
+                        {
+                            _stringReference = new StringReference(_chars, initialPosition, _charPos - initialPosition);
+                            return;
+                        }
+
+                        throw JsonReaderException.Create(this, "Invali
```
JamesNK: Move this switched into shared method?

JamesNK: Move switch into shared method?

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,2327 @@
...
+#endif
+using System.Threading.Tasks;
+using Newtonsoft.Json.Serialization;
+using Newtonsoft.Json.Utilities;
+
+namespace Newtonsoft.Json
+{
+    public partial class JsonTextReader
+    {
+        private bool SafeAsync => GetType() == typeof(JsonTextReader);
```
JamesNK: What is the purpose of this? Should the result of this type check be cached for perf?

Src/Newtonsoft.Json/JsonTextWriter.Async.cs
```diff
@@ -0,0 +1,1196 @@
...
+using System.Numerics;
+#endif
+using System.Threading.Tasks;
+using Newtonsoft.Json.Utilities;
+
+namespace Newtonsoft.Json
+{
+    public partial class JsonTextWriter
+    {
+        private bool SafeAsync => GetType() == typeof(JsonTextWriter);
```
JamesNK: See previous SafeAsync comment

Src/Newtonsoft.Json/JsonTextWriter.Async.cs
```diff
@@ -0,0 +1,1196 @@
...
+        }
+
+        internal async Task DoCloseAsync(CancellationToken cancellationToken)
+        {
+            while (Top > 0)
+            {
+                await WriteEndAsync(cancellationToken).ConfigureAwait(false);
+            }
+
+            if (_writeBuffer != null)
```
JamesNK: Move code of rest of method into a shared method?

Src/Newtonsoft.Json/JsonTextWriter.Async.cs
```diff
@@ -0,0 +1,1196 @@
...
+
+        private Task WriteDigitsAsync(ulong uvalue, CancellationToken cancellationToken)
+        {
+            if (uvalue <= 9)
+            {
+                return _writer.WriteAsync((char)('0' + uvalue), cancellationToken);
+            }
+
+            if (cancellationToken.IsCancellationRequested)
+                return cancellationToken.CancelledAsync();
```
JamesNK: Standard is to always braces around if content

Src/Newtonsoft.Json/JsonTextWriter.Async.cs
```diff
@@ -0,0 +1,1196 @@
...
+
+            EnsureWriteBuffer();
```
JamesNK: Move 258-267 to shared method?

Src/Newtonsoft.Json/JsonTextWriter.Async.cs
```diff
@@ -0,0 +1,1196 @@
...
+            return SafeAsync ? (value == null ? WriteNullAsync(cancellationToken) : WriteValueNonNullAsync(value, cancellationToken)) : base.WriteValueAsync(value, cancellationToken);
+        }
+
+        internal async Task WriteValueNonNullAsync(byte[] value, CancellationToken cancellationToken)
+        {
+            await InternalWriteValueAsync(JsonToken.Bytes, cancellationToken).ConfigureAwait(false);
+            await _writer.WriteAsync(_quoteChar, cancellationToken).ConfigureAwait(false);
+            await Base64Encoder.EncodeAsync(value, 0, value.Length, cancellationToken).ConfigureAwait(false);
+            await Base64Encoder.FlushAsync(cancellationToken).ConfigureAwait(false);
+            await _writer.WriteAsync(_quoteChar, cancellationToken).ConfigureAwait(false);
```
JamesNK: Wow, that's a lot of awaits!

JonHanna:
It's probably okay, but it should really be `(true)` to be on the safe side. It's a mistake that NYE celebrations got in the way of fixing. Will rectify tomorrow.

Src/Newtonsoft.Json/JsonTextWriter.Async.cs
```diff
@@ -0,0 +1,1196 @@
...
+        }
+
+        internal async Task DoWriteValueAsync(DateTime value, CancellationToken cancellationToken)
+        {
+            await InternalWriteValueAsync(JsonToken.Date, cancellationToken).ConfigureAwait(false);
+            value = DateTimeUtils.EnsureDateTime(value, DateTimeZoneHandling);
+
+            if (string.IsNullOrEmpty(DateFormatString))
+            {
+                EnsureWriteBuffer();
```
JamesNK: 543-548 shared method?

Src/Newtonsoft.Json/JsonTextWriter.Async.cs
```diff
@@ -0,0 +1,1196 @@
...
+            return SafeAsync ? DoWriteValueAsync(value, cancellationToken) : base.WriteValueAsync(value, cancellationToken);
+        }
+
+        internal async Task DoWriteValueAsync(DateTimeOffset value, CancellationToken cancellationToken)
+        {
+            await InternalWriteValueAsync(JsonToken.Date, cancellationToken).ConfigureAwait(false);
+
+            if (string.IsNullOrEmpty(DateFormatString))
+            {
+                EnsureWriteBuffer();
```
JamesNK: 597-602 shared method?

Src/Newtonsoft.Json/JsonWriter.Async.cs
```diff
@@ -0,0 +1,1573 @@
...
+        /// If <see cref="JsonWriter.CloseOutput"/> is set to <c>true</c>, the destination is also closed.
+        /// </summary>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None"/>.</param>
+        /// <returns>A <see cref="Task"/> that represents the asynchronous operation.</returns>
+        /// <remarks>The default behaviour is to execute synchronously, returning an already-completed task. Derived
+        /// classes can override this behaviour for true asychronousity.</remarks>
+        public virtual Task CloseAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (cancellationToken.IsCancellationRequested)
+                return cancellationToken.CancelledAsync();
```
JamesNK: braces, etc

Src/Newtonsoft.Json/JsonWriter.Async.cs
```diff
@@ -0,0 +1,1573 @@
...
+                    throw JsonWriterException.Create(this, "Unexpected type when writing end: " + type, null);
+            }
+        }
+
+        internal async Task InternalWriteEndAsync(JsonContainerType type, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            // write closing symbol and calculate new state
+            int levelsToComplete = 0;
```
JamesNK: 227 to 251 as shared method?

Src/Newtonsoft.Json/JsonWriter.Async.cs
```diff
@@ -0,0 +1,1573 @@
...
+                    return WriteValueAsync((byte[])value, cancellationToken);
+                default:
+                    throw MiscellaneousUtils.CreateArgumentOutOfRangeException(nameof(token), token, "Unexpected token type.");
+            }
+        }
+
+        internal virtual async Task WriteTokenAsync(JsonReader reader, bool writeChildren, bool writeDateConstructorAsDate, bool writeComments, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            int initialDepth;
```
JamesNK: 645-658 shared method?

Src/Newtonsoft.Json/JsonWriter.Async.cs
```diff
@@ -0,0 +1,1573 @@
...
+        }
+
+        /// <summary>
+        /// Asynchronously writes a <see cref="Nullable{T}"/> of <see cref="short"/> value.
+        /// </summary>
+        /// <param name="value">The <see cref="Nullable{T}"/> of <see cref="short"/> value to write.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None"/>.</param>
+        /// <returns>A <see cref="Task"/> that represents the asynchronous operation.</returns>
+        /// <remarks>The default behaviour is to execute synchronously, returning an already-completed task. Derived
+        ///
```
JamesNK: Move 1551-1562 into shared method? Return bool for whether convertable and the convertedValue as an out

Src/Newtonsoft.Json/Utilities/AsyncUtils.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+        };
+
+        // Pre-allocate to avoid wasted allocations.
+        public static readonly Task<bool> False = Task.FromResult(false);
+        public static readonly Task<bool> True = Task.FromResult(true);
+
+        internal static Task<bool> ToAsync(this bool value) => value ? True : False;
+
+        public static Task CancelIfRequesedAsync(this CancellationToken cancellationToken)
```
JamesNK: Typo

Src/Newtonsoft.Json/Utilities/AsyncUtils.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            return cancellationToken.IsCancellationRequested ? CancelledAsync(cancellationToken) : null;
+        }
+
+        public static Task<T> CancelIfRequesedAsync<T>(this CancellationToken cancellationToken)
```
JamesNK: Typo

Src/Newtonsoft.Json/Utilities/AsyncUtils.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            return cancellationToken.IsCancellationRequested ? CancelledAsync<T>(cancellationToken) : null;
+        }
+
+        public static Task CancelledAsync(this CancellationToken cancellationToken)
+        {
+            Debug.Assert(cancellationToken.IsCancellationRequested);
+            return new Task(Nop, cancellationToken);
```
JamesNK: What is going on here?

Src/Newtonsoft.Json/Utilities/AsyncUtils.cs
```diff
@@ -0,0 +1,113 @@
...
+        public static Task<T> CancelledAsync<T>(this CancellationToken cancellationToken)
+        {
+            Debug.Assert(cancellationToken.IsCancellationRequested);
+            return new Task<T>(() => default(T), cancellationToken);
+        }
+
+        // Task.Delay(0) is optimised as a cached task within the framework, and indeed
+        // the same cached task that Task.CompletedTask returns as of 4.6, but we'll add
+        // our own property for previous frameworks.
+        internal static Task CompletedTask => Task.Delay(0);
```
JamesNK: Is it work caching `Task.Delay(0)` for perf?

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,2327 @@
...
+        }
+
+        private Task<int> ReadDataAsync(bool append, CancellationToken cancellationToken)
+        {
+            return ReadDataAsync(append, 0, cancellationToken);
+        }
+
+        private async Task<int> ReadDataAsync(bool append, int charsRequired, CancellationToken cancellationToken)
+        {
+            if (_isEndOfFile)
```
JamesNK: 120-177 shared method?

JonHanna:
> I see you have ConfigureAwait(false) in the unit tests.

Where. I can't find where I made that slip (but am more than willing to believe that I did and still can't fnd it...).

Src/Newtonsoft.Json/JsonTextReader.Async.cs
JonHanna: The purpose is that e.g. say someone overrides `ReadAsBoolean`. Now `ReadAsBooleanAsync` which is meant to be an asych equivalent to `ReadAsBoolean` is not, because what `ReadAsBoolean` does has changed. Therefore our implementation of `ReadAsBooleanAsync` asynchronously does the wrong thing. We could use reflection to check on whether either `ReadAsBoolean` and any virtual method it calls into was overridden but just assuming the async methods here aren't viable in a derived class is simpler and faster.
Unsure about the pros and cons of 1) checking once on construction 2) caching on first access and 3) not caching. 2 seems a reasonable middle-ground so I will change to that.

JamesNK:
> Where

Whoops, I see `ConfigureAwait` and I automatically assume it is for false. I thought `ConfigureAwait(true)` was the default? Why state it explicitly?

JonHanna:
I decided to state it explicitly precisely because it would be a bad idea to have `ConfigureAwait(false)` and hence adding in `ConfigureAwait(true)` marks "I know I want any synchronization to flow here, so don't think you're being clever adding a `ConfigureAwait(false)` at this point". Whether to be explicit or implicit is as such largely a style matter as to whether to express that explicitly or not. I'll happily follow any style guid decision without argument.

Src/Newtonsoft.Json/Utilities/AsyncUtils.cs
JonHanna: I'm honestly not sure. It's pretty quickly obtained going through `Task.Delay(0)`, but caching would be faster again. I think I'll change to caching as it only costs one static field after all.

Src/Newtonsoft.Json/Utilities/AsyncUtils.cs
JonHanna: This creates a task with a no-op action with a `CancellationToken` known to be cancelled, and hence will be cancelled when received. It's more or less the same as `Task.FromCanceled` in .NET 4.6 but we can't use that as we have to support 4.5

Src/Newtonsoft.Json/JsonTextReader.Async.cs
JonHanna: I was leaving consolidating shared code as a later step to make the changes of this PR more obvious. If you'd rather it was included I'll get onto moving it out tomorrow and push to this PR soon. (But do keep pointing out shared cases that you see, as the more eyeballs the merrier as far as that goes).

benaadams:
Hurrah! This makes me very happy 😊

Src/Newtonsoft.Json/Utilities/AsyncUtils.cs
ayende: Yes, it should be. That avoid allocation for a pretty common case

Jetski5822:
Same, good stuff man!!

Src/Newtonsoft.Json/Utilities/AsyncUtils.cs
JonHanna: @ayende the caching within `Task.Delay()` already avoids allocation for the case of `0`, so on that count it's not necessary, but it will be a faster call as `Task.Delay(0)` hits a range check and a check on cancellation being needed that caching will skip.

benaadams: As comparable example of completed Task, aspnet's use is as follows https://github.com/aspnet/Common/blob/dev/shared/Microsoft.Extensions.TaskCache.Sources/TaskCache.cs#L8-L22

Src/Newtonsoft.Json/Utilities/AsyncUtils.cs
JonHanna: I'm adding a comment and changing the name to `FromCanceled` to make this clearer.

Src/Newtonsoft.Json/JsonTextReader.Async.cs
JonHanna: I shared from slightly after this point, as leaving this check-and-return duplicated kept the shared method simpler, and this remaining bit isn't a very large duplication.

JonHanna:
Responded to feedback, and a few further changes. I did so as a bunch of small commits to make it easier to perhaps revert one or two on the basis of further feedback, they can be squashed if accepted.

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,1778 @@
...
+            switch (_currentState)
+            {
+                case State.Start:
+                case State.Property:
+                case State.Array:
+                case State.ArrayStart:
+                case State.Constructor:
+                case State.ConstructorStart:
+                case State.PostValue:
+                    for (;;)
```
JamesNK: while

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,1778 @@
...
+                                {
+                                    SetToken(JsonToken.None, null, false);
+                                    return null;
+                                }
+
+                                break;
+                            case '"':
+                            case '\'':
+                                await ParseStringAsync(currentChar, readType, cancellationToken).ConfigureAwait(false);
+                                switch (readType)
```
JamesNK: 1162-1184 shared method?

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,1778 @@
...
+            switch (_currentState)
+            {
+                case State.Start:
+                case State.Property:
+                case State.Array:
+                case State.ArrayStart:
+                case State.Constructor:
+                case State.ConstructorStart:
+                case State.PostValue:
+                    for (;;)
```
JamesNK: while

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,1778 @@
...
+                                {
+                                    SetToken(JsonToken.None, null, false);
+                                    return null;
+                                }
+
+                                break;
+                            case '"':
+                            case '\'':
+                                await ParseStringAsync(currentChar, readType, cancellationToken).ConfigureAwait(false);
+                                switch (readType)
```
JamesNK: switch shared method?

Src/Newtonsoft.Json/JsonWriter.Async.cs
```diff
@@ -0,0 +1,1633 @@
...
+                default:
+                    throw JsonWriterException.Create(this, "Unexpected type when writing end: " + type, null);
+            }
+        }
+
+        internal async Task InternalWriteEndAsync(JsonContainerType type, CancellationToken cancellationToken)
+        {
+            int levelsToComplete = CalculateLevelsToComplete(type);
+
+            while(levelsToComplete-- > 0)
```
JamesNK: Space between while and brace

Src/Newtonsoft.Json/JsonWriter.Async.cs
```diff
@@ -0,0 +1,1633 @@
...
+        {
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return cancellationToken.FromCanceled();
+            }
+
+            switch (token)
+            {
+                case JsonToken.None:
+
```
JamesNK: Remove blank line

Src/Newtonsoft.Json/JsonTextReader.Async.cs
JamesNK: Makes sense!

JamesNK:
What does MS do for async in unit tests and `ConfigureAwait`?

benaadams:
> What does MS do for async in unit tests and ConfigureAwait?

Apparently leave them out (in tests) unless it causes issue
https://github.com/dotnet/corefx/pull/14771#discussion_r94254124

JamesNK:
I looked around the corefx a bit and couldn't find any with ConfigureAwait, e.g. https://github.com/dotnet/corefx/blob/c9b1b944763aa8aa83c3bad29c12c4845ef38f21/src/System.IO.FileSystem/tests/FileStream/CopyToAsync.cs

I'd like a find/replace to get rid of the `ConfigureAwait(true)`s

JamesNK:
Looking at those tests, something to worry about now will be tests where the cancellation token is true.

Src/Newtonsoft.Json/Utilities/JavaScriptUtils.cs
```diff
@@ -314,5 +313,175 @@ public static string ToEscapedJavaScriptString(string value, char delimiter, boo
...
+    bool[] charEscapeFlags, StringEscapeHandling stringEscapeHandling, IArrayPool<char> bufferPool, char[] writeBuffer, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            // leading delimiter
+            if (appendDelimiters)
+            {
+                await writer.WriteAsync(delimiter, cancellationToken).ConfigureAwait(false);
+            }
+            else
+            {
+                cancellationToken.ThrowIfCancellationRequested();
```
JonHanna: In the other branch there's a check hit within the call to `writer.WriteAsync`, and it seemed a good idea to have one near the beginning of the call in the case that that branch wasn't taken.

Co-operative cancellation is one of those things it's impossible to be perfect with. I think this one is reasonable, but there are a few others in this method that seem excessive, so I'll cull them.

Src/Newtonsoft.Json/Utilities/JavaScriptUtils.cs
```diff
@@ -314,5 +313,175 @@ public static string ToEscapedJavaScriptString(string value, char delimiter, boo
...
+                            }
+
+                            BufferUtils.ReturnBuffer(bufferPool, writeBuffer);
+
+                            writeBuffer = newBuffer;
+                        }
+
+                        s.CopyTo(lastWritePosition, writeBuffer, start, length - start);
+
+                        cancellationToken.ThrowIfCancellationRequested();
```
JonHanna: It's subject to debate, but I think reasonable guidelines are:

1. Avoid doing anything with side-effects before at least one check.
2. On the other hand, avoid doing a check just before a call that will itself check.
3. Avoid large allocations or anything "weighty" before at least one check. (Though in one of the revision commits I removed a few I was doing before `EnsureWriteBuffer` on the basis that since that allocation is cached and comes from a shared store of buffers it's not quite as "heavy" as if it was a one-off allocation of a similarly-sized array)
4. Avoid having loops with no checks within them (but again, if you're calling a method that will itself check in that loop, that's fine.
5. It's better to check before an operation than after it.

Throwing within an `async` method results in the `Task` returned by that method having `IsCanceled` as true. Conversely, `await`ing a cancelled task results in a throw of a `TaskCanceledException` and calling `Wait()` or `Result` on it in an `AggregateException` wrapping one.

As such, throw when the method is `async` return a cancelled task when it isn't.

JonHanna:
I've actually started on precisely that cancellation test matter, but some chores have gotten in the way, so I won't be pushing until tomorrow.

Src/Newtonsoft.Json/JsonReader.Async.cs
```diff
@@ -105,6 +107,27 @@ public virtual Task<byte[]> ReadAsBytesAsync(CancellationToken cancellationToken
             return cancellationToken.CancelIfRequestedAsync<byte[]>() ?? Task.FromResult(ReadAsBytes());
         }
 
+
+        internal async Task<byte[]> ReadArrayIntoByteArrayAsync(CancellationToken cancellationToken)
+        {
+            List<byte> buffer = new List<byte>();
+
+            for (;;)
```
JonHanna: Yeah, they're exactly the same. To me `for (;;)` reads as "forever" and `while (true)` reads as "while the basic principles of Aristotelian logic and Boolean logic still hold", and hence a strange way to express "forever", but I think that might be a very personal take on which to use 😄

Src/Newtonsoft.Json/Utilities/JavaScriptUtils.cs
```diff
@@ -314,5 +313,175 @@ public static string ToEscapedJavaScriptString(string value, char delimiter, boo
...
+                        case '\u2028': // Line Separator
+                            escapedValue = @"\u2028";
+                            break;
+                        case '\u2029': // Paragraph Separator
+                            escapedValue = @"\u2029";
+                            break;
+                        default:
+                            if (c < charEscapeFlags.Length || stringEscapeHandling == StringEscapeHandling.EscapeNonAscii)
+                            {
+                                if (c == '\'' && stringEscapeHandling != StringEscapeHandling.EscapeHtml)
```
JonHanna: Indeed, we can share as far as 338-405.

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,1778 @@
...
+
+                                        // potential start of a surrogate pair
+                                        if (EnsureChars(2, true) && _chars[_charPos] == '\\' && _chars[_charPos + 1] == 'u')
+                                        {
+                                            char highSurrogate = writeChar;
+
+                                            _charPos += 2;
+                                            writeChar = await ParseUnicodeAsync(cancellationToken).ConfigureAwait(false);
+
+                                            if (StringUtils.IsLowSurrogate(writeChar))
```
JonHanna: There was a missed point within this that should have been calling into an async method within that. What's left is hence smaller and also hits a lot of locals used elsewhere, so I don't think it's worth it here.

Src/Newtonsoft.Json/JsonTextWriter.Async.cs
JonHanna: Not really compared to those that loop (or indeed other such code elsewhere like `Stream.CopyToAsync()`) when one considers that the loop could cycle through many times.

Src/Newtonsoft.Json/JsonReader.Async.cs
JamesNK: Meanwhile I just think `for (;;)` looks ugly 😸

JonHanna: Good enough for Kernighan & Ritchie, good enough for me. (That's a lie, I used to use their indent style once, but I dislike it now). In any case, those have all been changed to `while (true)`.

Src/Newtonsoft.Json/Utilities/JavaScriptUtils.cs
JamesNK: This for loop is used heavily and is performance important. I worry about placing everything in a method.

JonHanna: Yeah, I'm reading back over today's changes and wondering the same thing. This is pretty much certain not to be inlined, though I think 379-398 would be pushing our luck too, and that's the most common path for most data AFAICT.
We could use `MethodImplOptions.AggressiveInlining` but that's only available from 4.5 (earlier? Certainly not all the framework versions supported).
I think I will revert this one and just have the duplication here given the tight loop it's in.

JonHanna:
As discussed above, I took out the commit that changes `WriteEscapedJavaScriptString` and am leaving all the duplicate code in that and the new `Async` version as it's a tight loop.

Src/Newtonsoft.Json/Utilities/JavaScriptUtils.cs
JamesNK: Most ASCII code will be caught by the `continue` on 342 but non-ASCII code will hit here. Probably best to leave it as is for perf

Src/Newtonsoft.Json/JsonTextReader.Async.cs
JamesNK: ok!

Src/Newtonsoft.Json/Utilities/JavaScriptUtils.cs
```diff
@@ -314,5 +313,175 @@ public static string ToEscapedJavaScriptString(string value, char delimiter, boo
...
+                    if (escapedValue == null)
+                    {
+                        continue;
+                    }
+
+                    bool isEscapedUnicodeText = string.Equals(escapedValue, EscapedUnicodeText);
+
+                    if (i > lastWritePosition)
+                    {
+                        int length = i - lastWritePosition + ((isEscapedUnicodeText) ? UnicodeTextLength : 0);
```
JamesNK: I think this could be shared. It is only called once per string

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,1713 @@
...
+                                if (await ReadNullCharAsync(cancellationToken).ConfigureAwait(false))
+                                {
+                                    SetToken(JsonToken.None, null, false);
+                                    return null;
+                                }
+
+                                break;
+                            case '"':
+                            case '\'':
+                                await ParseStringAsync(currentChar, ReadType.ReadAsBytes, cancella
```
JonHanna: Sorry, I don't follow you here.

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,1713 @@
...
+using Newtonsoft.Json.Serialization;
+using Newtonsoft.Json.Utilities;
+
+namespace Newtonsoft.Json
+{
+    public partial class JsonTextReader
+    {
+        // It's not safe to perform the async methods here in a derived class as if the synchronous equivalent
+        // has been overriden then the asychronous method will no longer be doing the same operation.
+        private readonly bool _safeAsync;
```
JonHanna: Seems unlikely without also doing something uncompilable, but it's not going to do any harm either, so sure.

Src/Newtonsoft.Json/Utilities/JavaScriptUtils.cs
JonHanna: Yes, it'd be fine for us English-speakers with our 99% ASCII range script, but I'll have Russian and Japanese coders processing text-heavy JSON cursing my name 😉

Src/Newtonsoft.Json/Utilities/JavaScriptUtils.cs
JonHanna: You're right. I went back a bit to far in unsharing it. Will have it back in a few minutes.

Src/Newtonsoft.Json/JsonTextReader.Async.cs
```diff
@@ -0,0 +1,1778 @@
...
+                        charPos = _charPos;
+                        break;
+                    case StringUtils.LineFeed:
+                        _charPos = charPos - 1;
+                        ProcessLineFeed();
+                        charPos = _charPos;
+                        break;
+                    case '"':
+                    case '\'':
+                        if (_chars[charPos - 1] == quote)
```
JonHanna: I shared the contents of this branch but not the branch. It means that no input would result in it being hit more than once.

JamesNK:
FYI I made a small change to the build script to download nuget.exe during build rather than have it under source control. You'll need to remove it from the PR

JonHanna:
Yeah, I noticed that. Good stuff, it was a nuisance there.

Src/Newtonsoft.Json/JsonTextReader.Async.cs
JamesNK: I was thinking that strings used in both sync/async should be shared, but it isn't important.

JamesNK:
I can't think of anything else to share or change. Thanks!

I've done some benchmarking and profiling. Both async reading and writing are slower over a StringReader/StringWriter which is expected. Reading is about half as fast but writing is significantly slower. I'm not an async expert but `WriteEscapedJavaScriptStringAsync` allocations are very high.

image

JamesNK:
Perf tests https://github.com/JamesNK/Newtonsoft.Json/commit/13a31e3a4d0234dc9d9adcc62af7e18c3b7ebec7

Apologies for the awful perf test class. It is about 10 years old and I've never gotten around to cleaning it up.

Src/Newtonsoft.Json/JsonTextReader.Async.cs
JonHanna: Oh, I get you. The comment just ended up a bit away from the string in question.
I'm inclined to think it's not worth doing, as it will compile to a shared literal, and it's not something that seems likely to be slipped up on. Indeed, adding the const would increase assembly size, albeit very minutely.

JonHanna:
Being slower and heavier isn't much of a surprise since we sacrifice speed and memory use (though for the benefit of the better thread use in the sort of cases async benefits). Being **that** much heavier though, is a surprise, and it is also seemingly particularly with `WriteEscapedJavaScriptStringAsync` which doesn't seem to offer much in the way of ways to reduce allocations.

I wonder if changing the way it interacts with the buffering would help, though on first glance I imagine it would just mean we've changed from having 1,250,000 `Task<char[]>` to having as many `Task`.

JamesNK:
Isn't the issue that a new task is being created for each call to `WriteEscapedJavaScriptStringAsync` to return the buffer? Would placing the buffer inside a wrapper class, and passing that wrapper into `WriteEscapedJavaScriptStringAsync`, and changing the array ref in the wrapper when needed fix that?

JonHanna:
That's what I'm wondering in my last sentence above. There's still going to be tasks created though. Too late at night here for me to try right now though.

JamesNK:
This quick experiment eliminated the `Task<char[]>` allocations - https://github.com/JamesNK/Newtonsoft.Json/commit/2d981720b8d78ebab4ba593894b65a252c109361 - although I'm sure there must be a better way of optimizing this sort of thing.

JamesNK:
https://github.com/JamesNK/Newtonsoft.Json/commit/2d981720b8d78ebab4ba593894b65a252c109361

This quick experiment eliminated the `Task<char[]>` allocations and reduce traffic from 190megs to 140megs. I'm sure there must be a better way of optimizing this sort of thing. Need advice from an expert.

@dfowler @benaadams ? 😸

JamesNK:
https://github.com/JamesNK/Newtonsoft.Json/commit/2d981720b8d78ebab4ba593894b65a252c109361

This quick experiment eliminated the `Task<char[]>` allocations and reduce traffic from 190megs to 140megs. I'm sure there must be a better way of optimizing this sort of thing. Need advice from an expert.

@davidfowl @benaadams ? 😸

benaadams:
The return value is to change the instance buffer, in case of expansion/allocation? So would be mostly same array anyway? Cached task would be nasty especially with static method.

If you're ok with exposing the buffer as a property, change to generic constrained extension method? 

Other than test is only called in one place? Change from static to instance method? Would also cut down on params passed.

Ensuring `writeBuffer != null && writeBuffer.Length >= UnicodeTextLength (const 6)` before the call would get rid of one of the possible allocations. Adding an extra write to flush the buffer `if > length` would get rid of another.

Remaining one is `WriteEscapedStringThroughBuffer` but could always rent and return as part of that if oversized; rather than changing and holding on to the buffer? Or is it because it doesn't always use ArrayPool and may be allocating the buffer so better to hold?

benaadams:
Could also drop the instance buffer cache and internalize the pool to the method? Or that no go?

JonHanna:
Okay. The last four commits have had some impact on allocations from `WriteEscapedJavaScriptStringAsync`.
fba5260 changes the allocation approach to one that calls back to the writer to ask that the buffer be large enough. This tightens the coupling, but if there's ever another client for the method it could be loosened again with an internal interface. It reduced total bytes allocated by about 11% and objects by about 13%.

7d56e1e splits up the method a bit for different branches. That slightly worsened allocations, but as well as enabling the next bits it would likely improve things for non-delimiter using branches that could hit it.

45d87e7 checks if escaping is necessary at all and returns just the `TextWriter.WriteAsync()` task when it can. As well as being more thorough than `ShouldEscapeJavaScriptString` it also returns the point at the first character needing escaping is found so escaped writing can start by writing the first bit of the string and searching from that point on. This lowered allocations a bit further and could be even more impactful in cases with lots of strings that don't have characters that need escaping.

d97451e finally just tightened up a few unnecessary checks (the buffer can no longer be null so don't test for it) and avoids calling `WriteAsync` with zero chars. It's effect was marginal, but it's still worth doing.

Src/Newtonsoft.Json/JsonTextReader.Async.cs
JonHanna: Just saw this question now. I'm not familiar with the implementation of those classes (though I keep meaning to become so), but StreamReader/Writer takes this approach. It was seeing it do so that gave me the eureka moment to come back to this, as when I first took a look at doing so a year I'd gotten into much more complicated knots trying to manage the issue of overrides.

JamesNK:
> loosened again with an internal interface

Sounds good.

FYI there is a change to an exception inside JsonTextReader that might need to be replicated in the async version https://github.com/JamesNK/Newtonsoft.Json/pull/1127/files#diff-2e13f50a9db15cfb3f0139531b953d32R1238

JonHanna:
Yeah, had to update the test to correspond to the new error message.

JamesNK:
Memory profiling shows the Task issue is solved which is great. Writing performance is still rather bad. AFAICT it is overhead from async state machines.

image

Src/Newtonsoft.Json/JsonWriter.Async.cs
```diff
@@ -0,0 +1,1690 @@
...
+        {
+            Debug.Assert(_formatting == Formatting.Indented);
+            if ((oldState == State.Object || oldState == State.Array || oldState == State.Constructor) && tokenBeingWritten != JsonToken.Comment)
+            {
+                await WriteValueDelimiterAsync(cancellationToken).ConfigureAwait(false);
+            }
+
+            if (oldState == State.Property)
+            {
+                await WriteIndentSpaceAsync(cancellationToken).ConfigureAwait(false);
```
JonHanna: I think this path isn't reachable, but I'm not 100% sure. Cutting it would be a slight improvement, but only slight since once you've one `await` having even a lot more is just adding states to the `MoveNext()` and so not a big further impact.

JonHanna:
Okay, some more changes that have some improvements in performance.

8ecd5fd is one I'm a bit uneasy about, but it definitely has an impact. It leads to some duplication, though the duplicate cases are visually close to each other which helps in terms of maintenance, but since the change to `AutoCompleteAsync` means it often returns synchronously and since it, or `InternalWriteValueAsync` which tail-calls it, followed by a single `await` we can branch on `IsCompleted` and just call that task instead of awaiting.

The downside in terms of duplication is not to be sniffed at, though.

Src/Newtonsoft.Json/JsonWriter.Async.cs
JamesNK: I don't know. It seems like an odd check. It might be a situation that can come up in a JTokenWriter or BsonWriter.

JamesNK:
Definitely faster. It has gone from about 240ms to 180ms 👍

Src/Newtonsoft.Json/JsonTextWriter.cs
```diff
@@ -317,27 +319,28 @@ private void UpdateCharEscapeFlags()
...
+                _writer.Write(_indentChars, newLineLen, Math.Min(currentIndentCount, 12));
+            }
+        }
 
-                    currentIndentCount -= writeCount;
-                }
+        private int SetIndentChars()
+        {
+            string writerNewLine = _writer.NewLine;
+            if (_newLine != writerNewLine)
```
JamesNK: Could you add a comment explaining what is going on here? It isn't immediately obvious

Src/Newtonsoft.Json/JsonTextWriter.cs
```diff
@@ -317,27 +319,28 @@ private void UpdateCharEscapeFlags()
...
-                if (_indentChars == null)
-                {
-                    _indentChars = new string(_indentChar, 10).ToCharArray();
-                }
+            int newLineLen = SetIndentChars();
 
-                while (currentIndentCount > 0)
-                {
-                    int writeCount = Math.Min(currentIndentCount, 10);
+            _writer.Write(_indentChars, 0, newLineLen + Math.Min(currentIndentCount, 12));
```
JamesNK: Is there a reason you've increased the indentChars from length 10 to 12? I pick that number randomly when I wrote this so I'm just curious.

Src/Newtonsoft.Json/JsonTextWriter.cs
```diff
@@ -317,27 +319,28 @@ private void UpdateCharEscapeFlags()
...
+        }
 
-                    currentIndentCount -= writeCount;
-                }
+        private int SetIndentChars()
+        {
+            string writerNewLine = _writer.NewLine;
+            if (_newLine != writerNewLine)
+            {
+                _indentChars = ((_newLine = writerNewLine) + new string(_indentChar, 12)).ToCharArray();
```
JamesNK: 12 is scattered everywhere in both the sync and async methods related to this. I think it should be a const on JsonTextWriter.

Src/Newtonsoft.Json/JsonTextWriter.cs
JonHanna: 4 spaces is a common indent size, so for 2 extra chars we get a 50% reduction in loops for large indenting at that size.

Src/Newtonsoft.Json/JsonTextWriter.cs
JonHanna: Agreed.

JamesNK: Hmmm, will _indentChars now get reallocated depending on whether a new line is required? To avoid that issue what about always having a new line at the start of the char array, and then change the start position of where you write from the _indentChars array to `TextWriter.WriteAsync(char[], int, int)`?

JamesNK:
I think after these last few comments this monster can be merged.

Src/Newtonsoft.Json/JsonTextWriter.cs
JamesNK: Ok. Json.NET defaults to indenting 2 spaces but making the size a multiple of 4 seems like a good idea for users who override the default indent size to 4.

Src/Newtonsoft.Json/JsonTextWriter.cs
JamesNK: Hmmm, will _indentChars now get reallocated depending on whether a new line is required? To avoid that issue what about always having a new line at the start of the char array, and then change the start position of where you write from the _indentChars array to `TextWriter.WriteAsync(char[], int, int)`?

JonHanna: Previously the only methods using it were always calling WriteNewLine/Async. Now they write from _indentChars and if necessary to repeat skip past the newline at the start.

It does get reallocated if someone changes the `NewLine` property of the TextWriter (or the IndentChar property of the writer itself, as before).

Src/Newtonsoft.Json/JsonTextWriter.cs
JonHanna: Yep, that was my thinking.

JonHanna:
Here we go…

Src/Newtonsoft.Json/JsonTextWriter.cs
JamesNK: Gotcha. One thing I'm trying to do is minimize the number of fields on JsonTextWriter/JsonTextReader because they increase memory size of objects that hang around until serialization is complete. Instead of a new field, what about comparing the `_writer.NewLine` chars to the beginning of the `_indentChars` array?

JonHanna: Ah. I see. It'd be a more expensive check, especially since 99% of the time the exact same strings would be in place and string `==` is optimised to short-circuit to true on same objects, but it's probably worth it on balance. Done.

JamesNK:
Ready to go I think. Squish it and I'll merge 😸

JonHanna:
Squashed.

(I gather there's a way to squash while merging in one handy step in github now, but I haven't been on that side of a github project since it was brought in).

JamesNK:
So there is. I'd forgotten about that feature.

Merged :shipit:

JonHanna:
It's a good feature, as lots of little commits is great in a PR, but can lead to very chatty logs if merged above about 2 or 3.

Right, now the fun begins on tokens and schemata.

JamesNK:
Don't worry about async and JsonValidatingReader. It is obsolete and can use the base async methods from JsonReader. Thing I'd most like to see async are JTokenReader/JTokenWriter

JonHanna:
I'm not going to touch the obsolete stuff, but the Schema project can gain from it once it picks it up in the nuget.

The token readers and writers don't have much to gain, since they're mostly doing in-memory stuff, but the serialisers will gain from the tokens having async methods, so it's a sensible next step.

PR #1125, opened 2017-01-03
Use null-propogation where applicable and useful.

JonHanna:
Primarily where it means one fewer field-access or simplifies event delegate invocation, though one or two cases have just the lesser advantage of allowing an IL stack `dup` rather than local access.

(Cases that involve introducing a nullable, e.g. where `obj != null ? obj.Value : 0` can be turned into `(obj?.Value) ?? 0` are less clearly wins, and so not done).

### DIFF
--- a/Src/Newtonsoft.Json/Serialization/JsonSerializerInternalWriter.cs
+++ b/Src/Newtonsoft.Json/Serialization/JsonSerializerInternalWriter.cs
@@ -483,39 +483,36 @@ namespace Newtonsoft.Json.Serialization
                 }
             }
 
-            if (contract.ExtensionDataGetter != null)
+            IEnumerable<KeyValuePair<object, object>> extensionData = contract.ExtensionDataGetter?.Invoke(value);
+            if (extensionData != null)
             {
-                IEnumerable<KeyValuePair<object, object>> extensionData = contract.ExtensionDataGetter(value);
-                if (extensionData != null)
+                foreach (KeyValuePair<object, object> e in extensionData)
                 {
-                    foreach (KeyValuePair<object, object> e in extensionData)
+                    JsonContract keyContract = GetContractSafe(e.Key);
+                    JsonContract valueContract = GetContractSafe(e.Value);
+
+                    bool escape;
+                    string propertyName = GetPropertyName(writer, e.Key, keyContract, out escape);
+
+                    propertyName = (contract.ExtensionDataNameResolver != null)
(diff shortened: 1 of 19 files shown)


TylerBrinkley:
Looks good to me.

Src/Newtonsoft.Json/Utilities/ReflectionUtils.cs
```diff
@@ -123,13 +123,7 @@ public static MethodInfo GetBaseDefinition(this PropertyInfo propertyInfo)
                 return m.GetBaseDefinition();
             }
 
-            m = propertyInfo.GetSetMethod(true);
```
stakx: The main part of the `GetBaseDefinition` static method could probably be reduced to:

```c#
return propertyInfo.GetGetMethod(true)?.GetBaseDefinition()
    ?? propertyInfo.GetSetMethod(true)?.GetBaseDefinition()
    ?? null;
```

And in the same vein, for the `IsVirtual` static method (just above it):

```c#
return propertyInfo.GetGetMethod(true)?.IsVirtual
    ?? propertyInfo.GetSetMethod(true)?.IsVirtual
    ?? false;
```

Src/Newtonsoft.Json/Serialization/JsonSerializerInternalWriter.cs
```diff
@@ -483,39 +483,36 @@ private void SerializeObject(JsonWriter writer, object value, JsonObjectContract
...
 
-                        bool escape;
-                        string propertyName = GetPropertyName(writer, e.Key, keyContract, out escape);
+                    bool escape;
+                    string propertyName = GetPropertyName(writer, e.Key, keyContract, out escape);
 
-                        propertyName = (contract.ExtensionDataNameResolver != null)
-                            ? contract.ExtensionDataNameResolver(propertyName)
-                            : propertyName;
+                    propertyName = (contract.ExtensionDataNameResolver != null)
```
stakx: If I'm not mistaken, this could be further reduced to:

```c#
propertyName = contract.ExtensionDataNameResolver?.Invoke(propertyName) ?? propertyName;
```

But since `propertyName` doesn't always change, perhaps the extra assignment could be avoided with a regular `if` block:

```c#
if (contract.ExtensionDataNameResolver != null)
{
    propertyName = contract.ExtensionDataNameResolver(propertyName);
}
```

JamesNK:
Looks good to me. Merged

PR #1130, opened 2017-01-04
Linq bridge optimise

JonHanna:
Optimise a few spots in the linq bridge for 2.0

Only cases with relatively high bang for buck, and only ignoring anywhere that doesn't get hit by tests as irrelevant.

### DIFF
--- a/Src/Newtonsoft.Json/Utilities/LinqBridge.cs
+++ b/Src/Newtonsoft.Json/Utilities/LinqBridge.cs
@@ -72,7 +72,14 @@ namespace Newtonsoft.Json.Utilities.LinqBridge
     {
       CheckNotNull(source, "source");
 
-      return CastYield<TResult>(source);
+        var servesItself = source as IEnumerable<TResult>;
+        if (servesItself != null
+            && (!(servesItself is TResult[]) || servesItself.GetType().GetElementType() == typeof(TResult)))
+        {
+            return servesItself;
+        }
+
+        return CastYield<TResult>(source);
     }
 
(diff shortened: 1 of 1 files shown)


Src/Newtonsoft.Json/Utilities/LinqBridge.cs
```diff
@@ -72,7 +72,14 @@ public static IEnumerable<TResult> Empty<TResult>()
     {
       CheckNotNull(source, "source");
 
-      return CastYield<TResult>(source);
+        var servesItself = source as IEnumerable<TResult>;
+        if (servesItself != null
+            && (!(servesItself is TResult[]) || servesItself.GetType().GetElementType() == typeof(TResult)))
```
JonHanna: This test to avoid array variance causing an issue isn't in the framework linq, which is an in-progress bug. Guarding against it here just in case some 2.0 code in some way hits it and would hence be broken by this change.

Src/Newtonsoft.Json/Utilities/LinqBridge.cs
```diff
@@ -151,11 +158,20 @@ private static IEnumerable<TResult> RepeatYield<TResult>(TResult element, int co
       this IEnumerable<TSource> source,
       Func<TSource, bool> predicate)
     {
+      CheckNotNull(source, "source");
```
JonHanna: Since the use is internal it's tempting to make CheckNotNull debug-only, but there could be some path that delegates its own validation to it.

PR #1131, opened 2017-01-05
Optimise Extensions.Values

JonHanna:
Branch on key being null or not once, rather than once per loop.

Remove up-cast of token to `JToken` to allow jitter to either elide `JValue` path if impossible or turn into null-test when inevitable.

Replace is-and-cast with as-and-null-test for quicker cast in cases where cast is still done.


JamesNK:
LGTM

PR #1132, opened 2017-01-05
Replace is-and-cast with as-and-nullcheck

JonHanna:
Gives slightly better performance, so applying across the board adds up.

### DIFF
--- a/Src/Newtonsoft.Json/JsonWriter.cs
+++ b/Src/Newtonsoft.Json/JsonWriter.cs
@@ -1576,28 +1576,25 @@ namespace Newtonsoft.Json
 #endif
                 default:
 #if !PORTABLE
-                    if (value is IConvertible)
+                    IConvertible convertible = value as IConvertible;
+                    if (convertible != null)
                     {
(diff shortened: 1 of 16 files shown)


Src/Newtonsoft.Json/JsonWriter.cs
```diff
@@ -1576,28 +1576,25 @@ internal static void WriteValue(JsonWriter writer, PrimitiveTypeCode typeCode, o
...
-                    if (value is IConvertible)
+                    IConvertible convertible = value as IConvertible;
+                    if (convertible != null)
                     {
                         // the value is a non-standard IConvertible
                         // convert to the underlying value and retry
-                        IConvertible convertable = (IConvertible)value;
 
-                        TypeInformation typeInformation = ConvertUtils.GetTypeInformation(convertable);
+                        TypeInformation typeInformation = ConvertUtils.GetTypeInformation(convertible);
```
JonHanna: Might as well change the typo on the variable name while changing how it's initialised.

PR #1137, opened 2017-01-06
Async methods on tokens.

JonHanna:


Src/Newtonsoft.Json/Linq/JArray.Async.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// </summary>
+        /// <param name="writer">A <see cref="JsonWriter"/> into which this method will write.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+        /// <param name="converters">A collection of <see cref="JsonConverter"/> which will be used when writing the token.</param>
+        /// <returns>A <see cref="Task"/> that represents the asynchronous write operation.</returns>
+        /// <remarks>Derived classes must override this method to get asynchronous behaviour. Otherwise it will
+        /// execute synchronously, returning an already-completed task.</remarks>
+        public override Task WriteToAsync(JsonWriter writer, CancellationToken cancellationToken, params JsonConverter[] converters)
+        {
+            return GetType() == typeof(JArray) ? DoWriteToAsync(writer, cancellationToken, converters) : base.WriteToAsync(writer, cancellationToken, converters);
```
JonHanna: With tokens it's much more likely that they'll be written only once, AFAICT, so not worth caching the `GetType()` test.

Src/Newtonsoft.Json/JsonReader.Async.cs
```diff
@@ -204,6 +204,31 @@ public virtual Task<string> ReadAsStringAsync(CancellationToken cancellationToke
...
         }
+
+        internal async Task<bool> ReadAndMoveToContentAsync(CancellationToken cancellationToken)
+        {
+            return await ReadAsync(cancellationToken).ConfigureAwait(false) && await MoveToContentAsync(cancellationToken).ConfigureAwait(false);
+        }
+
+        internal async Task<bool> MoveToContentAsync(CancellationToken cancellationToken)
+        {
+            for (;;)
```
JamesNK: while true

Src/Newtonsoft.Json/JsonTextWriterInternal.cs
```diff
@@ -0,0 +1,356 @@
...
+using System.Numerics;
+#endif
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Utilities;
+#endif
+
+namespace Newtonsoft.Json
+{
+    // A sealed class derived from JsonTextWriter that, as it's internal and sealed, we know isn't
```
JamesNK: Does this have much effect? Is it saving more than an if check per operation?

Src/Newtonsoft.Json/JsonTextWriterInternal.cs
```diff
@@ -0,0 +1,356 @@
...
+        {
+            return DoWriteValueDelimiterAsync(cancellationToken);
+        }
+
+        protected override Task WriteEndAsync(JsonToken token, CancellationToken cancellationToken)
+        {
+            return DoWriteEndAsync(token, cancellationToken);
+        }
+
+
```
JamesNK: remove line

Src/Newtonsoft.Json/Linq/JConstructor.Async.cs
```diff
@@ -0,0 +1,113 @@
...
+        public override Task WriteToAsync(JsonWriter writer, CancellationToken cancellationToken, params JsonConverter[] converters)
+        {
+            return GetType() == typeof(JConstructor) ? DoWriteToAsync(writer, cancellationToken, converters) : base.WriteToAsync(writer, cancellationToken, converters);
+        }
+
+        private async Task DoWriteToAsync(JsonWriter writer, CancellationToken cancellationToken, params JsonConverter[] converters)
+        {
+            await writer.WriteStartConstructorAsync(_name, cancellationToken).ConfigureAwait(false);
+
+            foreach (JToken token in Children())
```
JamesNK: Can you change this to a for loop over the _values collection? Also do the same thing in the sync Write. This was an optimization done in JArray that wasn't repeated in JConstructor.

Src/Newtonsoft.Json/Linq/JContainer.Async.cs
```diff
@@ -0,0 +1,175 @@
...
+                {
+                    if (parent == this)
+                    {
+                        return;
+                    }
+
+                    parent = parent.Parent;
+                }
+
+                switch (reader.TokenType)
```
JamesNK: Could this switch be moved to a shared method?

JamesNK: Hmmm, although I guess it is rather important for perf when loading tokens.

Src/Newtonsoft.Json/Linq/JValue.Async.cs
```diff
@@ -0,0 +1,145 @@
...
+                    return writer.WriteValueAsync((byte[])_value, cancellationToken);
+                case JTokenType.Guid:
+                    return writer.WriteValueAsync(_value != null ? (Guid?)_value : null, cancellationToken);
+                case JTokenType.TimeSpan:
+                    return writer.WriteValueAsync(_value != null ? (TimeSpan?)_value : null, cancellationToken);
+                case JTokenType.Uri:
+                    return writer.WriteValueAsync((Uri)_value, cancellationToken);
+            }
+
+            throw MiscellaneousUtils.CreateArgumentOutOfRangeException("TokenType", _valueType, "Unexpected token type.");
```
JamesNK: Could `"TokenType"` be changed to `nameof(TokenType)`? If so also repeat in sync version.

Src/Newtonsoft.Json/Newtonsoft.Json.Portable.csproj
```diff
@@ -136,11 +143,14 @@
     <Compile Include="Linq\JsonPath\ScanFilter.cs" />
     <Compile Include="Linq\JsonPath\ScanMultipleFilter.cs" />
     <Compile Include="Linq\JToken.cs" />
+    <Compile Include="Linq\JToken.Async.cs" />
     <Compile Include="Linq\JTokenEqualityComparer.cs" />
     <Compile Include="Linq\JTokenReader.cs" />
```
JamesNK: No JTokenReader.Async.cs? Is it not needed?

JonHanna: Unless I missed something, there's nothing that can ever hit on async I/O since its all in-memory and hence nothing that could benefit from being done async. A few methods could be optimised such as returning CompletedTask directly from FlushAsync and having WriteValueAsync inline their work, but the improvement over the base is likely marginal, especially since we still have to guard against a subclass that did something else in Flush, etc.

Src/Newtonsoft.Json/JsonTextWriterInternal.cs
JonHanna: Right now it'll be marginal, but I introduced it in my first stab at this since async serialisation would hit it many times.
OTOH, the caching of _safeAsync has made that check a lot cheaper. Once I've no longer got a toddler asleep on top of me I'll take it out. The benefits can be examined again if async serialisation means there's a path that does more hits on it again.

JonHanna: Actually, I see now that where I'd been using this isn't really appropriate for it anyway, so definitely gone.

Src/Newtonsoft.Json/Linq/JConstructor.Async.cs
JonHanna: I think that wasn't a valid optimisation, since both `Children()` and `ChildrenTokens` are virtual and not sealed it's not certain that `Children()` can be replaced with `_values` in that call.

Since it hasn't broken anyone (no screaming in issues?) it's probably safe enough, but I think it would be better to go through `Children()` but have a means for going to `IList<T>` if possible.

I'll do so. As compensation for the performance hit, at least we can cache `Count` locally rather than hit it `Count` times.

JamesNK: JObject/JArray/JValue and friends really aren't designed to be inherited from

JonHanna: Yes, inheriting doesn't seem very profitable. I wonder if we could seal them. If nobody cried when you made that optimisation…

Src/Newtonsoft.Json/Linq/JContainer.Async.cs
JonHanna: It could be shared and made `MethodImplOptions.AggressiveInlining` though that would only be available with 4.5 and higher so other frameworks would take a performance hit.

Src/Newtonsoft.Json/Linq/JValue.Async.cs
JonHanna: There's no such argument, and as such this probably shouldn't have been an `ArgumentOutOfRangeException` but `InvalidOperationException` or something similar. I'm leaving it as-is.

JonHanna:
Responded to feedback.

Src/Newtonsoft.Json/Linq/JConstructor.Async.cs
JonHanna: Failing that. We could remove the subclass checks and change the documentation to say that derived classes should override them if they override the equivalent.
It seems reasonable to put the burden on the implementors when there aren't expected to be any, and the result won't be as bizarre as it could be with JsonTextReader/Writer if someone did do such a subclass. 
I think I shall do that when I get back to my computer.

JonHanna:
On the basis that inheriting from `JToken` outside of the project isn't very profitable, I'm removing the subclassing tests. Do you think that might be too risky?

Src/Newtonsoft.Json/Linq/JValue.Async.cs
```diff
@@ -0,0 +1,138 @@
...
+                case JTokenType.Comment:
+                    return writer.WriteCommentAsync(_value != null ? _value.ToString() : null, cancellationToken);
+                case JTokenType.Raw:
+                    return writer.WriteRawValueAsync(_value != null ? _value.ToString() : null, cancellationToken);
+                case JTokenType.Null:
+                    return writer.WriteNullAsync(cancellationToken);
+                case JTokenType.Undefined:
+                    return writer.WriteUndefinedAsync(cancellationToken);
+                case JTokenType.Integer:
+                    if (_value is int)
```
JamesNK: As optimizations?

Src/Newtonsoft.Json/Linq/JConstructor.cs
```diff
@@ -164,9 +164,10 @@ public override void WriteTo(JsonWriter writer, params JsonConverter[] converter
         {
             writer.WriteStartConstructor(_name);
 
-            foreach (JToken token in Children())
+            int count = _values.Count;
```
JonHanna: I didn't examine this case, but having measured similar cases before, as a rule, it's not faster if `_values` is an array and the compiler can see that (not the case here, and the `foreach` would be just as fast if that was the case anyway), it's not much faster in an `async` or `yield` iterator (but is still a bit, at a cost of another field to the state machine structure), and it's faster otherwise.

Src/Newtonsoft.Json/Linq/JValue.Async.cs
JonHanna: Do you mean doing as-and-null-check instead of is-and-cast? All the types involved here are value types, so that can't be done.

JamesNK: True, true

Src/Newtonsoft.Json/Linq/JConstructor.cs
JamesNK: I did a quick benchmark and the difference is very minor. I'd prefer to use `_values.Count` in the for loop and save this sort of optimization for hot path code.

JonHanna: I tend to go for one extreme or the other; either just use `foreach` or do every little tweak available. In any case though, done as requested.

JamesNK:
I've done some perf testing. For writing I think there is the opportunity to optimize `JProperty.WriteToAsync` with the same pattern used in JsonTextWriter to avoid the async state machine overhead.

image

`JavaScriptUtils.WriteEscapedJavaScriptStringWithDelimitersAsync` and `JsonTextWriter.DoWritePropertyNameAsync` are also pretty high in time taken. Is there a way to have a sync fast path for methods with 3 awaits?

JamesNK:
Perf tests https://github.com/JamesNK/Newtonsoft.Json/commit/6559f59087a1b86aa9f96108721210fb1b1e9f6c

JamesNK:
I made the perf changes myself. More big async write perf gains https://github.com/JamesNK/Newtonsoft.Json/compare/JonHanna-tokens_async

image

JonHanna:
There should be a `.ConfigureAwait(false)` at https://github.com/JamesNK/Newtonsoft.Json/compare/JonHanna-tokens_async#diff-1ea2446866c30c4a0d050e6c79a699a6R59 but apart from that, looks good.

JamesNK:
What will happen if an exception is thrown in these optimized methods? Should it be being caught and `Task.FromException` be used to wrap the exception?

JamesNK:
Hmm, I thought about this some more and it makes sense to me that as long as there is no chance a method that returns a `Task` without `async` can throw then there is no need to try/catch. Methods that it calls that return tasks will take care of themselves and return a faulted task if need be.

JamesNK:
More thinking! If a task has a faulted state, shouldn't it be returned immediately? We're testing if it a task is completed or not, but if it is faulted then then the rest of the method will continue to execute.

JonHanna:
Oh, that last point is a good one. The exception it could pass up immediately (depending on where in the called task it throws, but it could also mean we eat an exception, and that's very bad.

Right, everything else on my open-source to-do list is on hold!

JonHanna:
Incidentally, I've also been doing some thinking, and something occured to me that I wish had occurred a month ago. What do you think of the way that XmlReader does it's async methods.

```
// Any async call will throw NotImplementedException() (NotSupported would be better IMO);
var xr = new XmlTextReader(streamReader); 

// Create was introduced with 2.0. Any async call call InvalidOperationException
var xr = XmlReader.Create(streamReader); 

// Async proprety introduced with 4.5. Async methods work.
var xr = XmlReader.Create(streamReader, new XmlReaderSettings{Async = false}); 
```

The last two return `XmlTextReaderImpl` and `XmlAsyncCheckReaderWithLineInfoNS` respectively, with other subclasses existing. It gets around the complications of dealing with subclasses. It wouldn't be as suitable for `StreamReader` with a higher expectation that third-party subclasses would be produced, but `JsonReader` and `JsonWriter` are perhaps closer to `XmlReader/Writer` in this regard, and it wouldn't hurt any existing users if we did it now before the next nuget release.

PR #1139, opened 2017-01-06
Call nullable WriteValue overloads in TraceJsonWriter

### LINKED ISSUE #1138: Serialization doesn't call overriden nullable writes when tracing.
If a `WriteValue` method that takes a nullable value is overridden in a custom writer, then serialization will call it when appropriate if not tracing, but will not call it if a trace-writer is set. 

(It won't be called for null values either way, and whether that's a bug or not is less obvious, but it should certainly have the same behaviour regardless of whether a trace-writer is in use).

JonHanna:
Fixes #1138


JamesNK:
LGTM

PR #1035, opened 2016-10-03
Added string conversion virtual methods to JsonConverter

TylerBrinkley:
Added IJsonStringConverter interface to enable configuring dictionary key serialization and deserialization. Address #915.

Issue #915, opened 2016-05-23
Add Dictionary Key (De)serialization Settings


Tydude4Christ:
I've written a high performance Enum utility library that I'd like to use for serializing and deserializing enum values in Json.NET as System.Enum is woefully slow due to its use of reflection on each call. Through the use of a JsonConverter I'm able to use my library for most purposes but I noticed that if an enum is used as a dictionary key there is no way to make Json.NET use my library to serialize or deserialize the json property name. Please consider adding support for configuring dictionary key serialization and deserialization. Thanks.

tangdunfeng:
Use custom JsonConverter
[http://stackoverflow.com/questions/6845364/json-net-specify-converter-for-dictionary-keys](url)

Tydude4Christ:
Thanks @tangdunfeng, but that really doesn't seem like a great solution as instead of relying on Json.NET to handle the dictionary object creation you'd now have to do that yourself. It seems like it would be very difficult to make it as robust as Json.NET's dictionary handling.

Tydude4Christ:
@JamesNK, could you look at my attempt to achieve this in my fork. The commit is https://github.com/Tydude4Christ/Newtonsoft.Json/commit/372b62646e574c66a450c33977e0ddceda46a875

PR #1141, opened 2017-01-06
Added string conversion virtual methods to JsonConverter

TylerBrinkley:
Added string conversion virtual methods to JsonConverter to support dictionary key serialization and deserialization which addresses #915.

Fixed merge conflict from previous PR #1035.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Converters/StringEnumConverterTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Converters/StringEnumConverterTests.cs
@@ -545,6 +545,41 @@ namespace Newtonsoft.Json.Tests.Converters
         }
 
         [Test]
+        public void EnumDictionaryKeySerializationTest()
+        {
+            Dictionary<Bar, int> value = new Dictionary<Bar, int>
+            {
+                { Bar.FooBar, 9 },
+                { Bar.Bat, 8 },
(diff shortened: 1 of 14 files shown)


fgreinacher:
I like this a lot!

TylerBrinkley:
Fixed merge conflict related to the change to compilation constants by feature.

JamesNK:
I've been considering this for a while and I think it puts too much on JsonConverter. I would prefer to stick with TypeConverter as the way for converting strings in dictionary types. I just recently added support for the netstandard1.0 build for TypeConverter so it is available everywhere, and I want to invest a little more in support for specifying a TypeConverter via a contract resolver, and caching it to improve the performance.

TylerBrinkley:
Maybe it does put too much on `JsonConverter` but it does make things more intuitive. One would expect if using the `StringEnumConverter` that the dictionary key would be serialized the same as a value would without having to recreate the same logic with the same settings in a `TypeConverter` class. And currently for primitive types a `TypeConverter` is not even looked for during serialization or deserialization so as an example enums are serialized with a [call](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Serialization/JsonSerializerInternalWriter.cs#L1169) to `Convert.ToString`.

PR #1143, opened 2017-01-07
Don't throw on BsonWriter.WriteValue for null Uri or byte[]

### LINKED ISSUE #1142: BsonWriter throws for null Uri or byte[]
Attempting to call `WriteValue` with a null `Uri` throw a `NullReferenceException`.

Attempting to call `WriteValue` with a null `byte[]` succeeds, but the final `WriteEnd()` that finishes the document throws a `NullReferenceException`.

JonHanna:
Fixes #1142


JamesNK:
Good catch. Merged!

PR #1144, opened 2017-01-07
Implement async writing in BsonWriter

JonHanna:


JamesNK:
I'm considering splitting BSON serialization out into its own package. If I did that then any async changes would be only applied to the new package.

JonHanna:
No point adding to the size and maintenance burden in that case, I'll park this and the partially-done bsonreader, so.

JamesNK:
Hi

BSON has been split out into its own repo/package - https://github.com/JamesNK/Newtonsoft.Json.Bson

It is ready for PRs if you want to submit async for it.

JonHanna:
Hmm. There's about 10 places above where something internal to Newtonsoft.Json is used, and the WIP I had for the reader had more, since reading isn't as limited by the length-prefixing as writing is. I dislike using `InternalsVisibleTo` but I think it's going to be necessary here.

JamesNK:
What internals are being used? I'd like to avoid it if possible both by either copying and pasting for BSON specific logic, or making things public/protected if they're generally needed. Also InternalsVisibleTo is a real pain in the butt when mixing signed/non-signed builds.

JamesNK:
For example, some changes I made to the BSON impl to make it use protected instead of internal methods and making generally useful internal methods public - https://github.com/JamesNK/Newtonsoft.Json/commit/b65807bbd2be51e8fa5bfebb50af5786b3eeb9ed

JonHanna:
I'll see if I can reduce to that. I really don't like making internals visible

JamesNK:
It is too bad that I didn't make the initial commit a straight copy from Newtonsoft.Json, that way you could easily see the changes I made to the non-async BSON code to eliminate internal. A lot of things you'll call are available copied in Newtonsoft.Json.Bson.Utilities.

JonHanna:
I just needed to do everything completely differently. Simple 😄

PR #1145, opened 2017-01-07
Use singleons for Null, Undefined, true and false BsonTokens.

JonHanna:
With `Null` and `Undefined` `BsonValue`s, one such value is indistinguishable from another, so singletons can be used for them, reducing allocations.

With `BsonValue`s containing booleans as well as there being only two values, so there's a similar advantage, there's the extra benefit that if there is only one true value then a reference-equals check can be used instead of unboxing to determine the value to write.


JamesNK:
LGTM

PR #1147, opened 2017-01-07
Merge ifs into following switch on same expression in ReadAsBytes

JonHanna:


Src/Newtonsoft.Json/JsonReader.cs
```diff
@@ -510,29 +510,25 @@ public virtual byte[] ReadAsBytes()
         {
             JsonToken t = GetContentToken();
 
-            if (t == JsonToken.None)
+            switch (t)
             {
-                return null;
-            }
+                case JsonToken.None:
```
JamesNK: Merge this case with cases at 559 that also return null?

JonHanna: Oops. I should have spotted that. Done.

PR #1151, opened 2017-01-08
Use singleton for zero-element arrays.

JonHanna:
Create an easily-inlinable method that accesses the array used by `Enumerable.Empty` (and hence not just shared within this project but likely with the client assembly or other libraries it uses) and use it anywhere a zero-element array is used, reducing allocations.

### DIFF
--- a/Src/Newtonsoft.Json/Utilities/CollectionUtils.cs
+++ b/Src/Newtonsoft.Json/Utilities/CollectionUtils.cs
@@ -29,12 +29,16 @@ using System.Collections.ObjectModel;
 using System.Reflection;
 using System.Text;
 using System.Collections;
+using System.Diagnostics;
 #if NET20
 using Newtonsoft.Json.Utilities.LinqBridge;
 #else
 using System.Linq;
 #endif
 using System.Globalization;
+#if !(NET20 || NET35 || NET40 || PORTABLE40)
+using System.Runtime.CompilerServices;
+#endif
 using Newtonsoft.Json.Serialization;
 
 namespace Newtonsoft.Json.Utilities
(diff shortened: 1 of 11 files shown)


Src/Newtonsoft.Json/Utilities/DynamicProxyMetaObject.cs
```diff
@@ -176,7 +176,7 @@ public override DynamicMetaObject BindDeleteIndex(DeleteIndexBinder binder, Dyna
 
         private delegate DynamicMetaObject Fallback(DynamicMetaObject errorSuggestion);
 
-        private static readonly Expression[] NoArgs = new Expression[0];
+        private static Expression[] NoArgs => CollectionUtils.ArrayEmpty<Expression>();
```
JonHanna: Because this is a generic class this allocated an identical field for every `T` the class was used with. Changing to an inlinable property reduces this.

Src/Newtonsoft.Json/Utilities/ExpressionReflectionDelegateFactory.cs
```diff
@@ -87,40 +87,45 @@ private Expression BuildMethodCall(MethodBase method, Type type, ParameterExpres
...
+            else
             {
-                ParameterInfo parameter = parametersInfo[i];
-                Type parameterType = parameter.ParameterType;
-                bool isByRef = false;
-                if (parameterType.IsByRef)
+                argsExpression = new Expression[parametersInfo.Length];
+                refParameterMap = new List<ByRefParameter>();
+
+                for (int i = 0; i < parametersInfo.Length; i++)
```
JonHanna: If branching on `parametersInfo.Length == 0` anyway, we might as well skip this when it will hit zero loops.

Src/Newtonsoft.Json/Utilities/CollectionUtils.cs
```diff
@@ -340,9 +344,27 @@ public static Array ToMultidimensionalArray(IList values, Type type, int rank)
...
         }
+
+        // 4.6 has Array.Empty<T> to return a cached empty array. Lacking that in other
+        // frameworks, Enumerable.Empty<T> happens to be implemented as a cached empty
+        // array in all versions (in .NET Core the same instance as Array.Empty<T>).
+        // This includes the internal Linq bridge for 2.0.
+        // Since this method is simple and only 11 bytes long in a release build it's
+        // pretty much guaranteed to be inlined, giving us fast access of that cached
+        // array. With 4.5 and up we use AggressiveInlining just to be sure, so it's
+        // effectively the same as calling Array.Empty<T> even when not available.
```
JamesNK: The thing I worry about this is what happens if an implementation of `Enumerable.Empty<T>()` by someone at some point in time doesn't return an array? ArrayEmpty will start returning null.

We could put this code in a cctor and check if `Enumerable.Empty<T>()` is null or not, creating an empty array if it is, but that would require a new generic class to store the variable on.

Can you think of other options?

JamesNK: What about just creating a new array, `new T[0]`, if the result of Empty is null? That adds safety, although there is a small perf hit when fetching the array.

JonHanna: I flip-flopped on that a few times when doing this. Having contributed to System.Linq in .Net Core I'd say it would be extremely unlikely in Core and even less likely in a successor to 4.6. (We do even have an enumerable type that in some ways would be better to return in Core, but it's not going to happen).

But there's always the possibility of a completely new version of .NET apart from Core, Mono, the MS closed-source frameworks and it could do something different. Trying a more defensive version I found:

1. Good: It still gets inlined even without the `MethodImplOptions.AggressiveInlining`. (It ends up being 21 bytes of IL with one branch, so still very small and simple).
2. Bad: The null-check isn't optimised away by the jitter.

An extra null-check is probably worth it for the safety though, but I'm leaving the assertion in the debug builds as if such a case did happen this would still be sub-optimal.

JamesNK:
LGTM. Merged

PR #1152, opened 2017-01-08
Optimise DynamicProxyMetaObject

JonHanna:
Remove unnecessary `ToArray` call.

Remove `Value` property.

- One use compares with null, so cast to T is a waste, base property serves better.

- This leaves just one other use, so it can be inlined.

Remove `_dontFallbackFirst`

- Is always `true`. Remove it and branches for it being `false`.

Don't use `Expression.Convert` to cast reference types to object.

- The end-result is a no-op, so don't waste the allocation unless the type is a value type.


JamesNK:
All seem good changes. Merged

PR #1153, opened 2017-01-08
Short-circuit on reference equality in ValueEquals and JValue.Compare

JonHanna:
Covers the both-null case previously checked for, so no extra cost, and skips all further work.

Also make two `is` checks a null-check after `as`.

### DIFF
--- a/Src/Newtonsoft.Json/Linq/JValue.cs
+++ b/Src/Newtonsoft.Json/Linq/JValue.cs
@@ -25,6 +25,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using Newtonsoft.Json.Utilities;
 using System.Globalization;
 #if !(NET35 || NET20 || PORTABLE40)
@@ -248,15 +249,15 @@ namespace Newtonsoft.Json.Linq
 
         internal static int Compare(JTokenType valueType, object objA, object objB)
         {
-            if (objA == null && objB == null)
+            if (objA == objB)
             {
                 return 0;
             }
-            if (objA != null && objB == null)
+            if (objB == null)
             {
                 return 1;
             }
-            if (objA == null && objB != null)
+            if (objA == null)
             {
                 return -1;
             }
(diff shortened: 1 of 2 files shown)


Src/Newtonsoft.Json/Linq/JValue.cs
```diff
@@ -350,21 +350,17 @@ internal static int Compare(JTokenType valueType, object objA, object objB)
...
-                    if (!(objB is byte[]))
+                    byte[] bytes2 = objB as byte[];
+                    if (bytes2 == null)
                     {
                         throw new ArgumentException("Object must be of type byte[].");
                     }
 
                     byte[] bytes1 = objA as byte[];
-                    byte[] bytes2 = objB as byte[];
                     if (bytes1 == null)
```
JonHanna: Should this being true be an exception?

JamesNK: Hrm, I'm not sure. I ran all the tests with debugging and neither if statement for byte[] being null was hit.

Src/Newtonsoft.Json/Utilities/MiscellaneousUtils.cs
```diff
@@ -39,15 +39,11 @@ internal static class MiscellaneousUtils
...
+            if (objA == objB)
             {
                 return true;
             }
-            if (objA != null && objB == null)
-            {
-                return false;
-            }
-            if (objA == null && objB != null)
+            if (objA == null | objB == null)
```
JamesNK: Why `|`?

Src/Newtonsoft.Json/Linq/JValue.cs
JonHanna: Yeah, I think it should probably just be asserted, but I'm not going to risk changing behaviour when I'm not 100% sure.

Src/Newtonsoft.Json/Utilities/MiscellaneousUtils.cs
JonHanna: Null checks against locals are fast, so the cost of an extra check when objA is null is worth it to have one less branch.

Its something that puts the micro in micro-opt, but if I have to choose between two options, might as well be the slightly quicker.

Src/Newtonsoft.Json/Linq/JValue.cs
JamesNK: I had a look at the code and I don't see any way to create a JValue with `TokenType.Bytes` that doesn't have a value of 'byte[]'.

Src/Newtonsoft.Json/Utilities/MiscellaneousUtils.cs
JamesNK: I'd like to keep micro-optimizations to critical areas.

Src/Newtonsoft.Json/Linq/JValue.cs
JonHanna: I checked back on the calls earlier too and couldn't find any either. Let's replace that check with an assertion.

Src/Newtonsoft.Json/Utilities/MiscellaneousUtils.cs
JonHanna: Sure. I've gotten into the habit of considering if the RHS is very small whenever I do a `||` or `&&` but that habit may not be widely shared.

PR #1155, opened 2017-01-09
Reduce enumeration of rightResults in BooleanQueryExpression.IsMatch

JonHanna:
If `Operator` is `Exists` or `leftResults` is empty don't obtain `rightResults`, as not relevant.

Otherwise if `rightResults` is not an `ICollection<JToken>` load it into a `List` so it won't be recomputed on each enumeration of `leftResults`.

### DIFF
--- a/Src/Newtonsoft.Json/Linq/JsonPath/QueryExpression.cs
+++ b/Src/Newtonsoft.Json/Linq/JsonPath/QueryExpression.cs
@@ -2,6 +2,11 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+#if NET20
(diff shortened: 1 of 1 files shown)


JamesNK:
I thought about a number of options for optimizing this when I wrote it - what is currently in master doesn't have anything.

Usually the right side of the boolean query expression will be a constant value, e.g. `@.name == 'james'` and the enumerable will be an array with the JValue of 'james' in it. This change should throw that array into a local var, correct?

JonHanna:
Yeah, arrays implement `ICollection<T>` for whatever their element type is so it'll end up going straight into `rightResults` without `ToList()` being called.
The case that's now worse is where the left has one element, and the right is a path expression, but it's not a lot worse so this seems sensible on balance.

PR #1156, opened 2017-01-09
Optimise WriteEscapedJavaScriptString

JonHanna:
Skip attempting write when string is empty.

`FirstCharToEscape` was introduced for the async version and gives a faster means to identify the first block before any escaping, so may as well use it here too.

Conversely, `WriteEscapedStringThroughBuffer` was introduced to share code between the sync and async versions, but later optimisations meant it was no longer used by the async. Inline the code back again.


JamesNK:
LGTM. Merged

PR #1159, opened 2017-01-09
Don't eat exceptions from faulted Tasks.

JonHanna:
If explicitly testing for `IsCompleted` then one must also test for `IsFaulted` and if not handling the exception, return the faulted `Task`.

### DIFF
--- a/Src/Newtonsoft.Json/JsonTextWriter.Async.cs
+++ b/Src/Newtonsoft.Json/JsonTextWriter.Async.cs
@@ -186,9 +186,13 @@ namespace Newtonsoft.Json
         private Task WriteValueInternalAsync(JsonToken token, string value, CancellationToken cancellationToken)
         {
             Task task = InternalWriteValueAsync(token, cancellationToken);
-            if (task.IsCompleted)
+            switch (task.Status)
             {
-                return _writer.WriteAsync(value, cancellationToken);
+                case TaskStatus.Canceled:
+                case TaskStatus.Faulted:
+                    return task;
+                case TaskStatus.RanToCompletion:
+                    return _writer.WriteAsync(value, cancellationToken);
             }
 
             return WriteValueInternalAsync(task, value, cancellationToken);
@@ -266,9 +270,13 @@ namespace Newtonsoft.Json
         private Task WriteIntegerValueAsync(ulong uvalue, bool negative, CancellationToken cancellationToken)
         {
             Task task = InternalWriteValueAsync(JsonToken.Integer, cancellationToken);
-            if (task.IsCompleted)
+            switch (task.Status)
             {
-                return WriteDigitsAsync(uvalue, negative, cancellationToken);
+                case TaskStatus.Canceled:
+                case TaskStatus.Faulted:
+                    return task;
+                case TaskStatus.RanToCompletion:
+                    return WriteDigitsAsync(uvalue, negative, cancellationToken);
             }
 
             return WriteIntegerValueAsync(task, uvalue, negative, cancellationToken);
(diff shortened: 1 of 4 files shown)


JamesNK:
Does it also make sense to return the task if it is cancelled?

JamesNK:
Are you sure the test is in the right place? When the `Task` isn't completed then it will be awaited, and the state machine should handle returning a faulted task.

It is when the Task IsCompleted is true that we need to check whether it is faulted, no?

JonHanna:
Oh darn it. Some of these are not right in other ways, too.

JonHanna:
No, it's the only flaw with it.

More haste, less speed.

PTAL, but I want a test hitting at least a couple of these paths before merge. Coming up soon.

JamesNK:
Should the task also be returned if it is cancelled?

JonHanna:
At the time we first added some of these I was thinking that since they were in a hot path and if cancellation had been requested it was likely to happen in a subsequent task anyway, so I deliberately left that out. If we're having a check anyway, maybe it's as well to check for that too. I wonder how light that check is.

JonHanna:
We can switch on `task.Status` and have faulted, canceled completed all in one test.

JonHanna:
Right. That should do it, I think.

Src/Newtonsoft.Json/JsonTextWriter.Async.cs
```diff
@@ -317,18 +325,28 @@ public override Task WritePropertyNameAsync(string name, CancellationToken cance
...
             Task task = InternalWritePropertyNameAsync(name, cancellationToken);
-            if (!task.IsCompleted)
+            switch (task.Status)
             {
-                return DoWritePropertyNameAsync(task, name, cancellationToken);
+                case TaskStatus.Canceled:
+                case TaskStatus.Faulted:
+                    return task;
+                case TaskStatus.RanToCompletion:
+                    break;
```
JamesNK: Add a comment that the task was completed in sync and to continue executing on sync path

Src/Newtonsoft.Json/Utilities/JavaScriptUtils.cs
```diff
@@ -381,21 +381,32 @@ public static Task WriteEscapedJavaScriptStringAsync(TextWriter writer, string s
...
             Task task = writer.WriteAsync(delimiter, cancellationToken);
-            if (!task.IsCompleted)
+            switch (task.Status)
             {
-                return WriteEscapedJavaScriptStringWithDelimitersAsync(task, writer, s, delimiter, charEscapeFlags, stringEscapeHandling, client, writeBuffer, cancellationToken);
+                case TaskStatus.Canceled:
+                case TaskStatus.Faulted:
+                    return task;
+                case TaskStatus.RanToCompletion:
+                    break;
```
JamesNK: ^^

Src/Newtonsoft.Json.Tests/JsonTextReaderTests/ExceptionHandlingAsyncTests.cs
```diff
@@ -1000,6 +1002,82 @@ public class ExceptionHandlingAsyncTests : TestFixtureBase
...
+            // allergic to certain octets, this null-stream writer throws on any attempt to write them.
+
+            private char[] _singleCharBuffer = new char[1];
+
+            public ThrowingWriter(params char[] throwChars)
+            {
+                ThrowChars = throwChars;
+            }
+
+            public bool ThrowSynchronously { get; set; }
```
JamesNK: Is this used?

JamesNK:
Just a couple of minor things. Looks good!

JonHanna:
I can't see why the exception isn't happening in the test in .NET core, and as such whether the bug is in the code or the test. It's late here, so I won't be looking at it again for some time, but maybe you can see what the issue is.

JonHanna:
I'd actually fixed that problem and introduced something silly. This is good to go.

It occurs to me that now we're explicitly looking at the status we can in fact do anything in the `switch` for failed tasks; as is returns the failure immediately, continuing to the method that would `await` it would result in that method failing. The latter would perhaps have a slightly better call-stack, in terms of the method names seen, but the former (as currently is in this PR) would be marginally faster. It's much of a muchness, I suppose.

JamesNK:
LGTM

PR #1171, opened 2017-01-13
Some optimisations to async reading.

JonHanna:
`MoveToContentAsync` is often called in states where it will immediately return true without any reading. In this case return `AsyncUtils.True` avoiding allocation or async state-machine.

`DoReadAsync` is a very heavily hit method, and can often be fulfilled without creating another state machine, so do this.

The remaining looping async case is the only use of `ParsePostValueAsync`, so inline it into the calling method, so there is one less task allocation.

Return value of `EatWhitespaceAsync` is never used, returning void reduces the number of locals and hence the size of the state-machine. Do the same for `EatWhitespace` though the impact is less.

### DIFF
--- a/Src/Newtonsoft.Json/JsonReader.Async.cs
+++ b/Src/Newtonsoft.Json/JsonReader.Async.cs
@@ -27,7 +27,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.Threading;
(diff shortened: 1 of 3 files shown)


JamesNK:
I was thinking about doing something similar to `DoReadAsync` last night. Looks good to me.

PR #1176, opened 2017-01-14
Optimise XmlNodeConverter

JonHanna:
Short-circuit `SerializeGroupedNodes` on no child nodes: Avoids creating unused `Dictionary`.

Don't create `Dictionary` in `ReadAttributeElements` if it'll be empty.

Reduce calls to `JsonReader.TokenType` virtual property.

### DIFF
--- a/Src/Newtonsoft.Json/Converters/XmlNodeConverter.cs
+++ b/Src/Newtonsoft.Json/Converters/XmlNodeConverter.cs
@@ -1177,6 +1177,11 @@ namespace Newtonsoft.Json.Converters
 
         private void SerializeGroupedNodes(JsonWriter writer, IXmlNode node, XmlNamespaceManager manager, bool writePropertyName)
         {
+            if (node.ChildNodes.Count == 0)
+            {
+                return;
+            }
+
             // group nodes together by name
             Dictionary<string, List<IXmlNode>> nodesGroupedByName = new Dictionary<string, List<IXmlNode>>();
 
--- a/Src/Newtonsoft.Json/Converters/XmlNodeConverter.cs
+++ b/Src/Newtonsoft.Json/Converters/XmlNodeConverter.cs
@@ -1605,33 +1605,34 @@ namespace Newtonsoft.Json.Converters
                 }
             }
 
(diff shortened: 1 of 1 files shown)


Src/Newtonsoft.Json/Converters/XmlNodeConverter.cs
```diff
@@ -1177,6 +1177,11 @@ private bool IsArray(IXmlNode node)
 
         private void SerializeGroupedNodes(JsonWriter writer, IXmlNode node, XmlNamespaceManager manager, bool writePropertyName)
         {
+            if (node.ChildNodes.Count == 0)
```
JamesNK: I ran this through a memory profiler and a collection is still allocated even though it will always be empty. Using XmlNode.HasChildNodes avoids allocating the ChildNodes collection. There will be other places this could be optimized.

To add this to the wrapper implementation for XNode you can test whether LastNode is null.

JamesNK: Same thing with XmlElement.HasAttributes.

JonHanna: Good idea. I'm not sure it's worth doing for attributes though, as it being not-null but empty requires hitting the same collection anyway so it's a bit swing and roundabouts, and it being null is already tested for.

JonHanna: Yes, I tried doing a `HasAttributes` but as well as it being guarded against for the null case already it doesn't offer as much of an improvement. The `HasChildNodes` test is in now, though.

JamesNK:
I thought about it some more after the review and I wrote this - https://github.com/JamesNK/Newtonsoft.Json/commit/96fc838858096a7be4167b8d0deeaf8803ffc0d3

It keeps all the tests for HasChildNodes and HasAttributes internal to the wrappers. XmlNodeConverter will happily use the collections.

If you remove the last commit then I'll squash the others then add mine.

JonHanna:
That's out. I added two comments on your commit (one minor but does have an impact, one trivial).

PR #1177, opened 2017-01-14
Optimise BinaryConverter's reflection object creation.

JonHanna:
Make the test for a type being S.D.L.Binary a single test as `Binary` is a sealed class and therefore there's no need to test bases.

Make the stored `ReflectionObject` static, reducing size and repeated allocations if `BinaryConverter` is explicitly created by the user.


JamesNK:
LGTM

PR #1178, opened 2017-01-14
Replace some null checks with conditionl operator, reducing field access

JonHanna:
Just three that were missed as they were being added on another branch at the same time that their sync equivalents had this done on them.


JamesNK:
:shipit:

PR #1179, opened 2017-01-15
Don't explicitly check for Failed or Cancelled Statuses

JonHanna:
Between discussion at #1137 and a few rounds of revision at #1159 we arrived at an optimisation where tasks that are both heavily hit and very likely to return in a completed state are checked for a status of either `RanToCompletion`, `Canceled` or `Faulted` and a path is taken where no `await` is needed for any of those.

The `Canceled` and `Faulted` tasks are returned immediately, allowing that state to be passed to the caller.

However, these are exceptional cases, and if we just continued the subsequent `await` would raise the appropriate exception. Checking for all three cases instead of just `RanToCompletion` is optimising for those exceptional conditions at cost (in the form of a larger comparison segment of code) to the more common cases of the task either being completed or still running.

Simplify this back to a check just for `RanToCompletion`.

### DIFF
--- a/Src/Newtonsoft.Json/JsonTextReader.Async.cs
+++ b/Src/Newtonsoft.Json/JsonTextReader.Async.cs
@@ -495,14 +495,10 @@ namespace Newtonsoft.Json
             _charPos++;
 
             Task<bool> task = EnsureCharsAsync(1, append, cancellationToken);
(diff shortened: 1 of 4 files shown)


JamesNK:
Hmm, I'm not fused either way. I like that there is slightly less code.

PR #1181, opened 2017-01-15
Handle F# lists that implement IReadOnlyCollection<T> (F# 4.1 and above)

### LINKED ISSUE #1180: Support upcoming F# 4.1
There is a [breaking change](https://github.com/Microsoft/visualfsharp/issues/2257) in F# 4.1 concerning Json.NET.
It causes an exception when deserializing to FSharpList type

_"Newtonsoft.Json.JsonSerializationException: Cannot create and populate list type Microsoft.FSharp.Collections.FSharpList`1[System.Int32]. Path '', line 1, position 1."_

Because in F# 4.1 FSharp list type now [implements IReadOnlyCollection](https://github.com/Microsoft/visualfsharp/commit/13b0ba050d450ca09a3e461d6355e490a273fca9)

this conditional block in [JsonArrayContract.cs](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Serialization/JsonArrayContract.cs#L182)
```csharp
#if HAVE_READ_ONLY_COLLECTIONS
else if (ReflectionUtils.ImplementsGenericDefinition(underlyingType, typeof(IReadOnlyCollection<>), out tempCollectionType))
```
is executed instead of [the following](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Serialization/JsonArrayContract.cs#L210)
```csharp
#if HAVE_FSHARP_TYPES
if (!HasParameterizedCreatorInternal && underlyingType.Name == FSharpUtils.FSharpListTypeName)
```

Basically as for now the logic in JsonArrayContract assumes FSharpList type doesn't implement IReadOnlyCollection. This is no longer true in the upcoming F# release.

fgreinacher:
Fixes #1180.

### DIFF
--- a/Src/Newtonsoft.Json/Serialization/JsonArrayContract.cs
+++ b/Src/Newtonsoft.Json/Serialization/JsonArrayContract.cs
@@ -191,6 +191,11 @@ namespace Newtonsoft.Json.Serialization
 
                 _genericCollectionDefinitionType = typeof(List<>).MakeGenericType(CollectionItemType);
                 _parameterizedConstructor = CollectionUtils.ResolveEnumerableCollectionConstructor(CreatedType, CollectionItemType);
+
+#if HAVE_FSHARP_TYPES
+                StoreFSharpListCreatorIfNecessary(underlyingType);
+#endif
+
                 IsReadOnlyOrFixedSize = true;
                 canDeserialize = HasParameterizedCreatorInternal;
             }
@@ -207,11 +212,7 @@ namespace Newtonsoft.Json.Serialization
                 _parameterizedConstructor = CollectionUtils.ResolveEnumerableCollectionConstructor(underlyingType, CollectionItemType);
 
 #if HAVE_FSHARP_TYPES
-                if (!HasParameterizedCreatorInternal && underlyingType.Name == FSharpUtils.FSharpListTypeName)
-                {
-                    FSharpUtils.EnsureInitialized(underlyingType.Assembly());
-                    _parameterizedCreator = FSharpUtils.CreateSeq(CollectionItemType);
-                }
+                StoreFSharpListCreatorIfNecessary(underlyingType);
 #endif
 
                 if (underlyingType.IsGenericType() && underlyingType.GetGenericTypeDefinition() == typeof(IEnumerable<>))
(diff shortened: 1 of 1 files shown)


fgreinacher:
@JamesNK Is there any way to get a nupkg from a PR so that @majocha can verify?

majocha:
I've built Newtonsoft.Json.dll from your PR. Works fine!

JamesNK:
Hrm, it is unfortunate that there will be no way to unit test that both old and new FSharp library users will both work.

Src/Newtonsoft.Json/Serialization/JsonArrayContract.cs
```diff
@@ -191,6 +191,15 @@ public JsonArrayContract(Type underlyingType)
 
                 _genericCollectionDefinitionType = typeof(List<>).MakeGenericType(CollectionItemType);
                 _parameterizedConstructor = CollectionUtils.ResolveEnumerableCollectionConstructor(CreatedType, CollectionItemType);
+
+#if HAVE_FSHARP_TYPES
+                if (!HasParameterizedCreatorInternal && underlyingType.Name == FSharpUtils.FSharpListTypeName)
```
JamesNK: Move this into a method that is called here and below on line 219.

fgreinacher:
Updated commit, AppVeyor seems to be hung :/

0x53A:
> Hrm, it is unfortunate that after this there will be no way to unit test that users of both old and new FSharp library will work.

Hm I think it should be possible to check-in the binaries of both FSharp.Core.dlls, and then reference both with a different alias (https://msdn.microsoft.com/de-de/library/ms173212.aspx). You would need to rename the filename of one of them though ...

Src/Newtonsoft.Json/Serialization/JsonArrayContract.cs
```diff
@@ -207,11 +212,7 @@ public JsonArrayContract(Type underlyingType)
                 _parameterizedConstructor = CollectionUtils.ResolveEnumerableCollectionConstructor(underlyingType, CollectionItemType);
 
 #if HAVE_FSHARP_TYPES
-                if (!HasParameterizedCreatorInternal && underlyingType.Name == FSharpUtils.FSharpListTypeName)
```
dsyme: You will still need this when using FSharp.Core 4.4.1.0 and before.  FSharp.Core only impemented IEnumerable up to version 4.4.1.0 inclusive

fgreinacher: Yes, I extracted the logic in a new method and call it from here and from the other branch.

JamesNK:
LGTM

I'll probably do a new pre-release version on NuGet sometime soon so this is available for people who use the latest version of FSharp.Core

dsyme:
@JamesNK Thank you

blumu:
This bug also affects FSharpLu.Json (https://github.com/Microsoft/fsharplu/issues/14). Is the new Nuget available for testing now? @JamesNK

majocha:
it's fixed in beta
[https://www.nuget.org/packages/Newtonsoft.Json/10.0.1-beta1](https://www.nuget.org/packages/Newtonsoft.Json/10.0.1-beta1)

PR #1183, opened 2017-01-16
Throw for CheckAdditionalContent when extra content is preceded by co…

### LINKED ISSUE #1182: JsonSerializerSettings.CheckAdditionalContent is fooled by comments followed by further content.
If `JsonSerializerSettings.CheckAdditionalContent` is `true` then any real content (that is, not a comment or insignificant whitespace) in the JSON read after an object that was deserialised should throw an exception.

However, if there is a comment between the JSON for the object deserialised and such erroneous content, the check does not find the content and the exception isn't thrown.

JonHanna:
…mments.

Fixes #1182

### DIFF
--- a/Src/Newtonsoft.Json/Serialization/JsonSerializerInternalReader.cs
+++ b/Src/Newtonsoft.Json/Serialization/JsonSerializerInternalReader.cs
@@ -169,9 +169,12 @@ namespace Newtonsoft.Json.Serialization
 
                 if (checkAdditionalContent)
                 {
-                    if (reader.Read() && reader.TokenType != JsonToken.Comment)
+                    while (reader.Read())
                     {
-                        throw new JsonSerializationException("Additional text found in JSON string after finishing deserializing object.");
+                        if (reader.TokenType != JsonToken.Comment)
+                        {
+                            throw new JsonSerializationException("Additional text found in JSON string after finishing deserializing object.");
+                        }
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


JamesNK:
LGTM

JamesNK:
The same check needs to be fixed in some other places, e.g. https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Linq/JArray.cs#L179

JonHanna:
Yes, that's already underway.

Src/Newtonsoft.Json/Serialization/JsonSerializerInternalReader.cs
```diff
@@ -169,9 +169,12 @@ public object Deserialize(JsonReader reader, Type objectType, bool checkAddition
 
                 if (checkAdditionalContent)
                 {
-                    if (reader.Read() && reader.TokenType != JsonToken.Comment)
+                    while (reader.Read())
                     {
-                        throw new JsonSerializationException("Additional text found in JSON string after finishing deserializing object.");
+                        if (reader.TokenType != JsonToken.Comment)
+                        {
+                            throw new JsonSerializationException("Additional text found in JSON string after finishing deserializing object.");
```
JamesNK: I wonder if this thrown exception should be changed to include information from the reader via https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/JsonSerializationException.cs#L83

JonHanna: It's a tricky balance. There have certainly been times over the last month or so when I wished Json.NET was less informative in its error messages, as it would allow for some internal simplification, but there have certainly also been times when the line/column info has been useful to me as a user.
In this case there seems to be little to gain in not having the more informative message. The reader has to be there anyway, and once it's there the line/column info can be extracted.
I think I'll include that with the work I'm currently on that catches the relatives of this issue.

PR #1184, opened 2017-01-16
Fix issues that are variants of those #1182

JonHanna:
Fix issues similar to #1182 in `JToken` and derived types.

Include line information in exception.

### DIFF
--- a/Src/Newtonsoft.Json/JsonConvert.cs
+++ b/Src/Newtonsoft.Json/JsonConvert.cs
@@ -962,9 +962,15 @@ namespace Newtonsoft.Json
             {
                 jsonSerializer.Populate(jsonReader, target);
 
-                if (jsonReader.Read() && jsonReader.TokenType != JsonToken.Comment)
+                if (settings.CheckAdditionalContent)
                 {
-                    throw new JsonSerializationException("Additional text found in JSON string after finishing deserializing object.");
+                    while (jsonReader.Read())
+                    {
+                        if (jsonReader.TokenType != JsonToken.Comment)
+                        {
+                            throw new JsonSerializationException("Additional text found in JSON string after finishing deserializing object.");
+                        }
+                    }
                 }
             }
         }
--- a/Src/Newtonsoft.Json/JsonConvert.cs
+++ b/Src/Newtonsoft.Json/JsonConvert.cs
@@ -962,13 +962,13 @@ namespace Newtonsoft.Json
             {
                 jsonSerializer.Populate(jsonReader, target);
 
-                if (settings.CheckAdditionalContent)
+                if (settings != null && settings.CheckAdditionalContent)
                 {
                     while (jsonReader.Read())
                     {
                         if (jsonReader.TokenType != JsonToken.Comment)
                         {
-                            throw new JsonSerializationException("Additional text found in JSON string after finishing deserializing object.");
+                            throw JsonSerializationException.Create(jsonReader, "Additional text found in JSON string after finishing deserializing object.");
                         }
                     }
                 }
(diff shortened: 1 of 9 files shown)


Src/Newtonsoft.Json/JsonConvert.cs
```diff
@@ -962,9 +962,15 @@ public static void PopulateObject(string value, object target, JsonSerializerSet
             {
                 jsonSerializer.Populate(jsonReader, target);
 
-                if (jsonReader.Read() && jsonReader.TokenType != JsonToken.Comment)
+                if (settings != null && settings.CheckAdditionalContent)
```
JonHanna: Adding this check because it seems that the current behaviour of throwing when there isn't a setting requesting it is, given that such a setting exists and is available to the user, a bug in itself. Maybe I'm misinterpreting the intent here.

Src/Newtonsoft.Json/Linq/JArray.cs
```diff
@@ -176,9 +176,12 @@ public new static JArray Parse(string json, JsonLoadSettings settings)
             {
                 JArray a = Load(reader, settings);
 
-                if (reader.Read() && reader.TokenType != JsonToken.Comment)
+                while (reader.Read())
                 {
-                    throw JsonReaderException.Create(reader, "Additional text found in JSON string after parsing content.");
+                    if (reader.TokenType != JsonToken.Comment)
```
JonHanna: Should this also throw if `settings.CommentHandling == CommentHandling.Load`?

Src/Newtonsoft.Json/JsonConvert.cs
```diff
@@ -962,9 +962,15 @@ public static void PopulateObject(string value, object target, JsonSerializerSet
...
                 {
-                    throw new JsonSerializationException("Additional text found in JSON string after finishing deserializing object.");
+                    while (jsonReader.Read())
+                    {
+                        if (jsonReader.TokenType != JsonToken.Comment)
+                        {
+                            throw new JsonSerializationException("Additional text found in JSON string after finishing deserializing object.");
```
JamesNK: Change this exception to use `JsonSerializationException.Create(reader, "blah blah")`

Src/Newtonsoft.Json/Linq/JArray.cs
JamesNK: No. That is specific to comments in arrays. It really just exists as a backwards compatibility toggle for older bad behavior.

Src/Newtonsoft.Json/JsonConvert.cs
JamesNK: Sure, that makes sense.

Src/Newtonsoft.Json/Linq/JArray.cs
JonHanna: Ah. That makes that whole setting make more sense now :)

JamesNK:
Changes look good.

I'd like tests of the new behavior - exceptions with more detail and checking for additional content - before merging.

Src/Newtonsoft.Json/Linq/JArray.cs
```diff
@@ -176,9 +176,9 @@ public new static JArray Parse(string json, JsonLoadSettings settings)
             {
                 JArray a = Load(reader, settings);
 
-                if (reader.Read() && reader.TokenType != JsonToken.Comment)
+                while (reader.Read())
                 {
-                    throw JsonReaderException.Create(reader, "Additional text found in JSON string after parsing content.");
+                    // Any content encountered here other than a comment will throw in the reader.
```
JonHanna: As became apparent. There's no way to actually detect `reader.TokenType != JsonToken.Comment` without `catch`ing the exception from the reader itself, and little point since the exception thrown is applicable.

JamesNK: Ok! Looks good to me.

PR #1191, opened 2017-01-24
Replace ThreadSafeStore's lock with Interlocked.CompareExchange

TylerBrinkley:
Using `Interlocked.CompareExchange` instead of C#'s `lock` keyword should slightly improve performance when populating the `ThreadSafeStore`. I'm not sure how to go about performance testing it though.


JamesNK:
I'm not sure how much it would effect performance. The only thing that happens inside the lock is creating a new dictionary and adding the new item to it. And the lock isn't hit that often. ThreadSafeStore is only used to store some reflection data which is finite based on how many types an app is serializing. I'd need to see some perf numbers.

Also some unit tests showing it as thread safe really need to be added. The current implementation doesn't have tests but it has been proven in the real world not to have thread safety problems which is why I'm hesitant to change it.

JamesNK:
I should add I'm also familiar with using `lock` to do locking like this. I have not seen Interlocked.CompareExchange used like this.

TylerBrinkley:
This pattern, a compare and swap using a do while loop and `Interlocked.CompareExchange`, is actually generated by `C#` 4 and up to make declared events thread-safe and lockless. You can see it if you decompile a class with an event.

Here's a [link](https://blogs.msdn.microsoft.com/cburrows/2010/03/04/events-get-a-little-overhaul-in-c-4-part-i-locks/) to a post describing the changes to events with `C#` 4. It doesn't actually show you the implementation but this is how events are declared now.

I just looked through `ThreadSafeStore`s uses and most are static and like you said are used to store reflection data so this shouldn't have much effect so I'm closing this PR.

JonHanna:
CAS loops are generally heavier than locking under low contention, but become a win with heavy contention. But copying a dictionary would be quite a lot of work to do within one. If contention was heavy enough for it to be worth trying CAS loops, it would probably be better to have a concurrent dictionary in there instead.

TylerBrinkley:
@JonHanna Isn't it the opposite. Aren't CAS loops generally faster than locking under low contention but slower in heavy contention due to wasted work within the loop?

JonHanna:
Yeah, actually very low (so that looping almost never happens) tends to be a win for CAS too. The higher-contention cases I was thinking about when I wrote that above is when you've built something out of several individually light CASes. E.g. you can build concurrent collections that depend entirely on just CASes (and maybe a few memory barriers if CASes don't entail barriers, but they do in .NET) and the sum of e.g. a few CASes to add an element tends to be heavier than just a single lock wrapping it all for low-contention use but in high-contention use the fact that the CASes may be quite separate and that each will release when only part-way through the overall add method can mean better use when contention is heavy. Especially when the CASes can help each other out ("oh, that's been set now, my work is done"). In my comment above I was led astray by tending only to ever loop on CASes in such a case.

PR #1215, opened 2017-02-15
xml serialization with arrays when using JsonConvert.DeserializeXmlNode(String, String, Boolean)

reinaldoarrosi:
Given this class
```csharp
public class TestInfo
{
    public TestInfo()
    {
        Key = "Key1";
        Name = "Name1";
        Values = new string[] { "Value1" };
        Properties = new Dictionary<string, string>() { { "Key", "Value" } };
    }

    public string Key { get; set; }
    public string Name { get; set; }
    public string[] Values { get; set; }
    public Dictionary<string, string> Properties { get; set; }
}
```
JsonConvert.DeserializeXmlNode generates the following xml when writeArrayAttribute is set to true:
```xml
<TestInfo>
  <Key>Key1</Key>
  <Name>Name1</Name>
  <Values json:Array="true" xmlns:json="http://james.newtonking.com/projects/json">Value1</Values>
  <Properties>
    <Key>Value</Key>
  </Properties>
</TestInfo>
```

If we save this xml to a file, load it back using XmlDocument.Load and then try to serialize it to json using JsonConvert.SerializeXmlNode, the resulting json looks like this:
```json
{
  "Key": "Key1",
  "Name": "Name1",
  "Values": [
    {
      "#text": "Value1"
    }
  ],
  "Properties": {
    "Key": "Value"
  }
}
```
As you can see the "Values" array is not correctly serialized, which means that if we try to pass this json to JsonConvert.DeserializeObject<TestInfo> an exception is thrown.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Converters/XmlNodeConverterTest.cs
+++ b/Src/Newtonsoft.Json.Tests/Converters/XmlNodeConverterTest.cs
@@ -1162,6 +1162,32 @@ namespace Newtonsoft.Json.Tests.Converters
   }
 }";
             StringAssert.AreEqual(expected, arrayJsonText);
+
+            arrayXml = @"<root>
+			  <person id=""1"">
+				  <name>Alan</name>
+				  <url>http://www.google.com</url>
+				  <role json:Array=""true"" xmlns:json=""http://james.newtonking.com/projects/json"">Admin</role>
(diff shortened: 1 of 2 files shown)


JamesNK:
I rewrote the fix slightly to remove the `#ifdef`s. Thanks for the fix + test.

PR #1221, opened 2017-02-19
Make InternalsVisibleTo Newtonsoft.Json.Bson

JonHanna:
See https://github.com/JamesNK/Newtonsoft.Json/pull/1144#issuecomment-280927785


JonHanna:
I'm going to give a second go at doing without this, and see if I can reduce it to just making a small amount of stuff public. Will reopen if necessary.

PR #1250, opened 2017-03-21
Fixing JToken line info set

### LINKED ISSUE #1249: JToken and JObject Load doesn't store line information
I have a simple code that reads a JSON file and stores it to a JObject:
```
using (StreamReader file = File.OpenText(filePath))
using (JsonTextReader reader = new JsonTextReader(file))
{
    JsonLoadSettings loadSettings = new JsonLoadSettings()
    {
        CommentHandling = CommentHandling.Ignore,
        LineInfoHandling = LineInfoHandling.Load
    };
    
    JObject token = JObject.Load(reader, loadSettings);
}
```
From what I observed, the JsonTextReader has the correct line numbers, but after loading, the JObject and it's children do not have any line information. Same thing happens with JToken.Load.

davilimap:
The `SetLineInfo` method of JToken was doing nothing on `LineInfoHandling.Load`, which in turn caused problems in JToken as well as JObject parsing methods.

This explains and closes #1249 

Edit: After the CI build failed, I discovered unit tests that depended on this behavior, but without explicitly using the property. Since it seems to be consolidated behavior, this PR would constitute a breaking change. For now, I'll just leave it for consideration.

### DIFF
--- a/Src/Newtonsoft.Json/Linq/JToken.cs
+++ b/Src/Newtonsoft.Json/Linq/JToken.cs
@@ -2229,7 +2229,7 @@ namespace Newtonsoft.Json.Linq
 
         internal void SetLineInfo(IJsonLineInfo lineInfo, JsonLoadSettings settings)
         {
-            if (settings != null && settings.LineInfoHandling == LineInfoHandling.Load)
+            if (settings != null && settings.LineInfoHandling != LineInfoHandling.Load)
             {
                 return;
             }


JamesNK:
Thanks. I cleaned up the remaining issues https://github.com/JamesNK/Newtonsoft.Json/commit/68533cdc19095b44479b3a925f61662ca964066d

PR #1280, opened 2017-04-12
Use spec-compliant readOnly property

LordJZ:
http://json-schema.org/latest/json-schema-hypermedia.html#rfc.section.4.4


JamesNK:
Hi. JsonSchema and related classes have been deprecated. If it is still an issue add it to the https://github.com/JamesNK/Newtonsoft.Json.Schema repo

Issue #998, opened 2016-08-26
SelectToken with JSONPath fails with unexpected character $


saiphcbk:
I'm trying to test for the value of a property which is not in an array but i'm getting an error:
Unexpected character while parsing path query: $
here's my fiddle: [https://dotnetfiddle.net/QpkilW](https://dotnetfiddle.net/QpkilW)

my JSONPath filter expression is $.[?($.store.bicycle.price < 20)]
This expression works in the javascript and jayway's java implementation of JSONPath (Goessner 0.8)
[http://jsonpath.herokuapp.com/?path=$.[?($.store.bicycle.price%20%3C%2020)]](http://jsonpath.herokuapp.com/?path=$.[?($.store.bicycle.price%20%3C%2020)])

Can we not use the root object as part of the filter expression?

```
var json = @"
			{ ""store"": 
			  {
				""book"": [ 
				  { ""category"": ""reference"",
					""author"": ""Nigel Rees"",
					""title"": ""Sayings of the Century"",
					""price"": 8.95
				  },
				  { ""category"": ""fiction"",
					""author"": ""Evelyn Waugh"",
					""title"": ""Sword of Honour"",
					""price"": 12.99
				  },
				  { ""category"": ""fiction"",
					""author"": ""Herman Melville"",
					""title"": ""Moby Dick"",
					""isbn"": ""0-553-21311-3"",
					""price"": 8.99
				  },
				  { ""category"": ""fiction"",
					""author"": ""J. R. R. Tolkien"",
					""title"": ""The Lord of the Rings"",
					""isbn"": ""0-395-19395-8"",
					""price"": 22.99
				  }
				],
				""bicycle"": {
				  ""color"": ""red"",
				  ""price"": 19.95
				}
			  },
			  ""expensive"": 10
			}";
		var jobj = JObject.Parse(json);

		var coursesTokens = jobj.SelectTokens("$.[?($.store.bicycle.price < 20)]");
```

JamesNK:
Fixed https://github.com/JamesNK/Newtonsoft.Json/commit/ac2d61947428a34eb6e48d26bfd22f7141c757ce

IlyaDeskov:
Hello.
I've got the same error with '?' character:
code:
....
from jsonpath_ng import jsonpath, parse
....
json_label_synchronize = parse("labels[?(@.name == 'Sync')].id")
....
and part of the stacktrace:
   File "/app/app.py", line 19, in <module>
[...]

PR #1304, opened 2017-05-03
Fix RootFilter to proper return an IEnumerable containing the root

asiffermann:
This PR is related to #998 and its resolution with https://github.com/JamesNK/Newtonsoft.Json/commit/ac2d61947428a34eb6e48d26bfd22f7141c757ce

I was trying to achieve what is described by the unit test `RootInFilterWithInitializers` in my project using the latest version of Newtonsoft.Json (10.0.2), but `SelectTokens` always returned an empty Enumerable.

Suspecting the problem came from DateTime JValue, I tried what is described by the unit test `RootInFilterWithRootObject`, which came from an example in https://github.com/json-path/JsonPath and that you can test in [Jayway JsonPath Evaluator](http://jsonpath.herokuapp.com/?path=$..book[?(@.price%20%3C=%20$[%27expensive%27])]).

Then I tried to understand the use case described by @saiphcbk in #998, but I couldn't... I honestly do not know what he is trying to achieve with those data and this JsonPath.
BTW, if you look carefully at his [Jayway JsonPath Evaluator example](http://jsonpath.herokuapp.com/?path=$.%5B?%28$.store.bicycle.price%20%3C%2020%29%5D), you would see that in Jayway implementation the result is wrapped in an array (couldn't figure why) and that in Goessner it just return all objects, wrapped in an array too.
I guess you couldn't understand this use case too, and that's why in your unit test `RootInFilter` you wrapped his root object in an array. But then the unit test itself seems to be bad because the JsonPath `$.store.bicycle.price` does not match anymore...

Anyway, it seems obvious to me that the `RootFilter` should return an IEnumerable containing the root JToken rather than the root JToken itself. If not, you cannot access top level properties, and the unit tests I added would fail.

I changed your unit test `RootInFilter` the less intrusive way I found, but I'm not sure if it is necessary to keep it.

I tried to be the most exhaustive I could about this issue, just let me know what do you think about it 😃

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Linq/JsonPath/JPathExecuteTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Linq/JsonPath/JPathExecuteTests.cs
@@ -1271,11 +1271,85 @@ namespace Newtonsoft.Json.Tests.Linq.JsonPath
 
             JArray a = JArray.Parse(json);
 
-            List<JToken> result = a.SelectTokens("$.[?($.store.bicycle.price < 20)]").ToList();
+            List<JToken> result = a.SelectTokens("$.[?($.[0].store.bicycle.price < 20)]").ToList();
             Assert.AreEqual(1, result.Count);
 
-            result = a.SelectTokens("$.[?($.store.bicycle.price < 10)]").ToList();
+            result = a.SelectTokens("$.[?($.[0].store.bicycle.price < 10)]").ToList();
             Assert.AreEqual(0, result.Count);
+        }
+
+        [Test]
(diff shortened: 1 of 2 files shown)


JamesNK:
Thanks for the detailed investigation. I made a small change to avoid the yield (it generates a lot of code) but otherwise looks good! https://github.com/JamesNK/Newtonsoft.Json/commit/572661d826721690b3d8ef780c0e119da691e23f

asiffermann:
You're right about the yield, looks better with a simple array.
Thank you! 😃

PR #1310, opened 2017-05-12
Update Compare method@JValue.cs

yadoer:
For JTokenType.Float type, we should add the comparison for decimal type.


JamesNK:
How is this different than the same code here: https://github.com/JamesNK/Newtonsoft.Json/pull/1310/files#diff-0abb278d0397d70dd0f5f7714841e269R277

Add a unit test that this fixes.

JamesNK:
Merged! Thanks

PR #1333, opened 2017-06-01
Added support for instantiating a generic JsonConverter from a JsonConverterAttribute

### LINKED ISSUE #1332: Add support for generic JsonConverter instantiation
Consider the following value wrapper struct.

```csharp
[JsonConverter(typeof(ValueConverter<>))]
struct Value<T>
{
    public static implicit operator T(Value<T> value) => value._value;

    public static implicit operator Value<T>(T value) => new Value<T>(value);

    private readonly T _value;

    public Value(T value)
    {
        _value = value;
    }
}

class ValueConverter<T> : JsonConverter
{
    public override bool CanConvert(Type objectType) => objectType == typeof(Value<T>);

    public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer) => new Value<T>(serializer.Deserialize<T>(reader));

    public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer) => serializer.Serialize(writer, (T)(Value<T>)value, typeof(T));
}
```

Running the following code I currently get an `ArgumentException` of "Invalid type owner for DynamicMethod."

`JsonConvert.SerializeObject(new Value<bool>(true));`

Obviously the problem is that the converter type provided with the `JsonConverterAttribute` is an open generic type and Json.NET doesn't know how to instantiate it. When an open generic converter type is provided I would like Json.NET to implicitly use the generic type arguments of the type the attribute is applied to in order to instantiate the json converter.

TylerBrinkley:
Added support for instantiating a generic JsonConverter from a JsonConverterAttribute.

Fixes #1332

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/ContractResolverTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/ContractResolverTests.cs
@@ -701,11 +701,14 @@ namespace Newtonsoft.Json.Tests.Serialization
             public string Prop3 { get; set; }
             [JsonProperty]
             public string Prop4 { get; set; }
+            [JsonProperty]
+            public string Prop5 { get; set; }
 
             public bool Prop1Specified;
             public bool Prop2Specified { get; set; }
             public static bool Prop3Specified { get; set; }
             public event System.Func<bool> Prop4Specified;
+            public static bool Prop5Specified;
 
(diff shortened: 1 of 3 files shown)


TylerBrinkley:
Tried to rollback commit to fix merge conflict issue. Submitted a PR #1359 as a replacement.

Issue #1344, opened 2017-06-15
Slow ResolveContract with large class


heldersepu:
ResolveContract is very slow with large classes, here is an example:
https://github.com/heldersepu/csharp-proj/tree/master/Newtonsoft_ResolveContract
On that project I have 
```c#
    public class HugeClass
    {
        public bool var10000;
        public bool var10001;
        public bool var10002;
        public bool var10003;
        public bool var10004;
//      ...
        public bool var30997;
        public bool var30998;
        public bool var30999;
    }

```
There are more than 20K of those ```bool var10000```
and it takes more than a minute to execute the ResolveContract

WildCard65:
Resolve contract for objects iterates throw all the property and fields of the type, that iteration is probably where the slowdown is (especially if it's 20k items).

heldersepu:
@WildCard65 
Did you look at my example?
My example also iterates through all the properties using reflection... 
and it is significantly faster (milliseconds)

WildCard65:
@heldersepu My statement still stands as non-value types are serialized/deserialized as JSON objects, I forgot the iteration also includes creating instances of Newtonsoft.Json.Serialization.JsonProperty

heldersepu:
@WildCard65 are you saying that is OK to take more than a minute to do such a task?

heldersepu:
Just for kicks I added a ```new JsonProperty``` to my loop: 
https://github.com/heldersepu/csharp-proj/commit/14e351ee4378556b0d1120f1a0cb30970b0aa50b?diff=unified

...it makes almost no difference, the reflection loop runs in 6 miliseconds

```
11:52:03.703
Reflection GetProperties
11:52:03.709
ResolveContract
11:52:55.360

```

WildCard65:
ResolveContract detects if a Contract for the type already exists, if not, it calls CreateContract.

CreateContract deduces the type of the contract, for your example, it's JsonObjectContract and calls CreateObjectContract

CreateObjectContract creates an instance of JsonObjectContract, eventually it gets to CreateProperties and calls it.

CreateProperties calls a private utils method of Json.Net to get an enumerable containing both Fields and Properties of the type.

CreateProperties iterates the obtained enumerable (for your example, it will have 20k public fields) and creates a JsonProperty instance to handle the field.

Your slowdown is in CreateProperties at that iteration moment.

WildCard65:
Here's a screenshot of me running your example: http://prntscr.com/fk5mkx

[...]

PR #1346, opened 2017-06-16
Added ConcurrentDictionary support for better performance. 

TylerBrinkley:
Added `ConcurrentDictionary` support for better performance. Should probably add .NET Standard 1.1 target as that's where `ConcurrentDictionary` is first available.

Partially addresses #1344.


JamesNK:
This is better performance when resolving a contract but 99.9% of the time the performance of reading from a dictionary is more important. What effect does this have on getting items?

TylerBrinkley:
You're absolutely correct and I should have done more research into that before submitting the PR. Fortunately, `ConcurrentDictionary`'s reading performance is essentially identical to `Dictionary`'s. See [here](https://github.com/dotnet/corefx/blob/master/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L478) for the retrieval logic.

From testing the changes in my own library [Enums.NET](https://github.com/TylerBrinkley/Enums.NET/commit/51e6050a1e6d7dd667483126dfe0442d277c82d4) I've noticed essentially no performance difference in reading with `ConcurrentDictionary` vs `Dictionary`.

JamesNK:
Ok! Merged.

I made a couple of small changes like moving ConcurrentDictionary back behind ThreadSafeStore. I don't want to have a type with the same name but missing a lot of methods.

PR #1350, opened 2017-06-19
ModelMetadataType support is added for dotnetcore1.1

tursoft:


tursoft:
Can you please help to solve the AppVeyor build error?
Actually it seems ok on local.

PR #1352, opened 2017-06-20
Adding support for ModelMetadataTypeAttribute (issue1349)

tursoft:
Added ModelMetadataTypeAttribute  support for dotnetcore

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Issues/Issue1349.cs
+++ b/Src/Newtonsoft.Json.Tests/Issues/Issue1349.cs
@@ -1,0 +1,125 @@
+﻿#region License
+// Copyright (c) 2007 James Newton-King
+//
+// Permission is hereby granted, free of charge, to any person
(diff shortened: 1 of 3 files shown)


JamesNK:
Hi

Thanks but I don't want to have dependencies on features in individual libraries. If MetadataTypeAttribute is added back to .NET Core then I will add support for it.

tursoft:
Hi James,
Since current netcore version does not support **ModelMetadataType** (previously it was **MetadataClassType** and was supported in Newtonsoft.Json), Newtonsoft.Json library could not preserve backward compatibility. If you still do not want to add this support in dotnetcore version, at least you should add a mechanism to resolve associated metadata type (like it is in the name of the method within the library - **GetAssociatedMetadataType**). By this way, you still keep the library independent of external libraries (like MVC). For example, maybe we can register "**AssociatedMetadataTypeRevolvers**" to solve this issue?

What do you think about this solution?

Best regards..

PR #1353, opened 2017-06-21
Fixed deserializing ObservableCollection in .NET Core 2.0

justinvp:
Workaround for https://github.com/dotnet/corefx/issues/21304

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
@@ -1085,6 +1085,20 @@ namespace Newtonsoft.Json.Tests.Serialization
             Assert.AreEqual("1", s[0]);
             Assert.AreEqual("2", s[1]);
         }
+
+        [Test]
+        public void SerializeObservableCollection()
+        {
+            ObservableCollection<string> c1 = new ObservableCollection<string> { "1", "2" };
+
+            string output = JsonConvert.SerializeObject(c1);
+            Assert.AreEqual("[\"1\",\"2\"]", output);
+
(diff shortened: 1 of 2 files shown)


JamesNK:
A test please!

justinvp:
There's already a test for deserializing `ObservableCollection<T>` here: https://github.com/JamesNK/Newtonsoft.Json/blob/928eea83404b12916a85e47f61df6f4e1af2e121/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs#L1081-L1087

What else should I add?

Note: Without this change, the test is going to start failing when run on the next preview of .NET Core 2.0 that contains https://github.com/dotnet/corefx/commit/acb4fbd7fed04ea74035091542cf57f67687db8f

justinvp:
I updated the existing test to both serialize and deserialize.

JamesNK:
I didn't realize there was already a test for deserializing but having serializing as well is useful. Thanks

justintoth:
Was this issue ever fixed in ConcurrentDictionary or are we waiting on a new .Net Core release? I'm using the latest Newtonsoft.Json package (11.0.1) and am running into this error:

{System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Collections.Concurrent.ConcurrentDictionary`2.InitializeFromCollection(IEnumerable`1 collection)
   at System.Collections.Concurrent.ConcurrentDictionary`2.OnDeserialized(StreamingContext context)

My scenario is that I'm writing an object that contains a ConcurrentDictionary to ElasticSearch, which tries to serialize it as json. Full stack trace [here](https://gist.github.com/justintoth/efb3d32d42385a2c3b8f18cd86b382ed).

JamesNK:
This is not Json.NET: `Nest.Json.Serialization`

PR #1374, opened 2017-07-13
Add HAVE_ICONVERTIBLE to NETSTANDARD1_3

### LINKED ISSUE #1372: JValue Doesn't Implement IConvertable in .NET Core
### Source/destination types

```csharp
class Test
{
    public double MyDouble { get; set; }
}
```

### Source/destination JSON

```javascript
{"myDouble":1.0}
```

### Expected behavior

Running on .NET Core, when using the above json I first create a 'JObject'. I then get the myDouble key and try and convert it to a double using `Convert`:

```csharp
var jObj = JOtject.Parse(json);
var d = Convert.ToDouble(jObj["myDouble"]);
```
I would expect the above code to work.

### Actual behavior

The above code throws an exception complaining that JValue does not implement IConvertable.

### Steps to reproduce

Use .NET Core and run the above code.

I noticed that JValue does implement IConvertable but only when the `HAVE_ICONVERTIBLE` compiler directive is set. This was added in #1146 and cording to that PR it should be set for all versions except portable. Is that correct or is there a bug? As far as I can see .NET Core has IConvertable so this should work right?

shortspider:
Also added unit test.

Fixes #1372.


JamesNK:
I needed to fix up the tests a little (no idea why Assert.IsInstanceOf fails) but otherwise it looks good. Thanks.

https://github.com/JamesNK/Newtonsoft.Json/commit/c33b257f18ef36501fd379dd7bbb7adb01f1ff2b

Issue #1379, opened 2017-07-20
JSONPATH : Filter expression can't compare number properly


For2017Dev:
Giving a JSON string :
{
  "persons": [
    {
      "name"  : "John",
      "age": "26"
    },
    {
      "name"  : "Jane",
      "age": "2"
    }
  ]
}

And a JSONPATH script :
$.persons[?(@.age > 3)]

The SelectTokens function can't find a record.

I know that age field should be a number but in practical we should be able to compare string with number like that. (Tested on http://jsonpath.com/)

Since I tried to write $.persons[?(3 < @.age)] and it doesn't work, could we just switch the comparison like 

                    case QueryOperator.GreaterThan:
                        if (rightValue.CompareTo(leftValue) < 0)
                        {
                            return true;
                        }
                        break;
 
It will have expected behavior 

Unit Test :

        [Test]
        public void BooleanExpressionTest_StringOnLeftHand_CompareWith_NumberOnRightHand()
        {
            BooleanQueryExpression e1 = new BooleanQueryExpression
            {
                Operator = QueryOperator.GreaterThan,
                Right = new JValue(3),
                Left = new List<PathFilter>
                {
                    new ArrayIndexFilter()
                }
            };

            Assert.IsTrue(e1.IsMatch(null, new JArray("2", "26")));
        }

JamesNK:
Fixed https://github.com/JamesNK/Newtonsoft.Json/commit/ad1a13728a2f155d6b4f2ca39bcac4b9873e344a

PR #1380, opened 2017-07-20
fix issue with JSONPath query expression when compare string with a number

MinhNguyenDev:
To fix issue #1379


JamesNK:
A problem I see is your change will break the comparison in the opposite direction.

This test will fail after the PR:

```
        [Test]
        public void GreaterThan_ComparingStringParameterWithIntegerValues()
        {
            string json = @"{
  ""persons"": [
    {
      ""name""  : ""John"",
      ""age"": 26
    },
    {
      ""name""  : ""Jane"",
      ""age"": 2
    }
  ]
}";

            JObject models = JObject.Parse(json);

            var results = models.SelectTokens("$.persons[?(@.age > '3')]").ToList();

            Assert.AreEqual(1, results.Count);
        }

```

MinhNguyenDev:
Thanks for your reply, James! But in pratice the right hand value is the one we have control, we can choose > 3 or > '3', the left hand is the field in JSON object that we can't change. So I think this can be a quick work around to support most use cases.

JamesNK:
I don't like quick fixes. The proper fix is to make `leftValue.CompareTo(rightValue)` and `rightValue.CompareTo(leftValue)` return equivalent results.

JamesNK:
I made this change instead - https://github.com/JamesNK/Newtonsoft.Json/commit/ad1a13728a2f155d6b4f2ca39bcac4b9873e344a

Issue #858, opened 2016-03-26
using regularExpression in JsonPath filter criteria


binitjha:
I am trying to do filter based on regular expression ( or simple contains ) in filter. for example 

`{
  "firstName": "John",
  "lastName" : "doe",
  "age"      : 26,
  "address"  : {
    "streetAddress": "naist street",
    "city"         : "Nara",
    "postalCode"   : "630-0192"
  },
  "phoneNumbers": [
    {
      "type"  : "iPhone",
      "number": "0123-4567-8888"
    },
    {
      "type"  : "home",
      "number": "0123-4567-8910"
    }
  ]
}`
For the Json above
$..phoneNumbers[?(@.type == 'home')].type  ==> works fine.

but I want to do something like @.type contains a given string or better yet matches with a regular expression. This doesn't seem to be supported.  Something like
$..phoneNumbers[?(@.type startswith 'home')].type  OR
$..phoneNumbers[?(@.type =~ /*ho*/)].type

Thanks.

JamesNK:
You can't do that. If in the future there is a better standard, or a popular defacto standard then I'll add support.

binitjha:
Thanks James. 
Really appreciate your response. There are few operator supported already. Instead of full fledge regex , one can also support simple string comparison like string.startswith or endswith or contains. 

I am not sure who updates the standard but I think the JSONPath is similar as Xpath and it supports regex.

Lets see if there are more demands for it.

MinhNguyenDev:
We really need the regex compare in filter to make JSONPATH useful in real application. In Javascript JSONPATH library and other (i.e Jayway Jsonpath) they all can use regex. It would be very helpful to have "=~" syntax like Jayway (https://github.com/json-path/JsonPath)

PR #1382, opened 2017-07-20
feature: implement new regex equals operator for JsonPath

MinhNguyenDev:
Example: Blah[?( @.name =~'/hi/i' )].
Feature request from #858

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Linq/JsonPath/JPathParseTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Linq/JsonPath/JPathParseTests.cs
@@ -315,6 +315,17 @@ namespace Newtonsoft.Json.Tests.Linq.JsonPath
         }
 
         [Test]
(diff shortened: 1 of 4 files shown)


JamesNK:
Thanks for looking into this. Some thoughts:

* If this is going to be added then it should match http://jsonpath.herokuapp.com/?path=$..book[?(@.author%20=~%20/.*REES/i)] and not have quotes surrounding the regex.
* The left/right arguments should probably be changed from JToken to object so that a Regex can be set to one. Re-parsing the regex for every comparison will hurt performance. This could be a lot of work
* Handle all the various Regex options - https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Converters/RegexConverter.cs#L144-L173

JamesNK:
I'm going to close this. If you make further progress on the points above you are welcome to submit a new PR.

PR #1390, opened 2017-07-27
Feature : JsonPath-based projection converter

magicmoux:
First of all thanks a lot for the tremendous work you did on this library.

My PR proposes a ProjectionConverter Class that takes the root instance being serialized as a base context for dictionary-defined propertyPaths (the idea is to work the same as X/Json-Path template-matches or EF eager-loading Includes)

As much as I searched, I could not find a solution to the problem I had : my first use-case was to automatically define an optimal/terse ElasticSearch document based on the indices' field-mappings for documents **without resorting to hard-coded C#/Linq projections**.

I feel this proposal could be useful in any situation where we need to serialize ViewModels of an object depending on the target system/framework configuration (instead of code) whereas opt-ins/out annotations are not suited because non-contextual. 

Currently it should works with .NET 3.5 to 4.x; if the PR is accepted, help would be most welcome to expand support to .NetCore or netstandard environments if possible

Thx for your consideration and your feedback.
Max.

### DIFF
--- a/Src/Newtonsoft.Json/Converters/ProjectionConverter.cs
+++ b/Src/Newtonsoft.Json/Converters/ProjectionConverter.cs
@@ -59,6 +59,10 @@ namespace Newtonsoft.Json.Converters
         private ReferenceLoopHandling baseReferenceLoopHandling;
         private HashSet<String> propertyPaths = null;
 
+        /// <summary>
+        /// Creates a new ProjectionConverter instance
+        /// </summary>
+        /// <param name="projections"></param>
         // TODO implement constructors from KeyValuePairs and replace the HashSet by a IEnumerable
         public ProjectionConverter(Dictionary<Type, HashSet<string>> projections = null)
             : base()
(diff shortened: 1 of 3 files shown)


Src/Newtonsoft.Json/Converters/ProjectionConverter.cs
```diff
@@ -0,0 +1,187 @@
...
+                _rootType = value.GetType();
+                // NOTE ? is it valuable to handle inheritance and interfaces ?
+                while (propertyPaths == null)
+                {
+                    if (!PropertyPaths.TryGetValue(value.GetType(), out propertyPaths))
+                    {
+                        _rootType = _rootType.BaseType;
+                        if (_rootType == null)
+                        {
+                            // TODO ? revert to basic native serialization based on the current
```
magicmoux: ? is it valuable to handle inheritance and interfaces ?

JamesNK:
Hi

I'm trying hard to keep the size of Json.NET down by focusing on essentials. For example I spent a lot of time moving the BSON functionality into its own package. I would prefer that additional JsonConverters that people write like this one are placed in their own NuGet packages.

magicmoux:
Hi James, 
that's totally OK, I understand.
Max.

PR #1393, opened 2017-07-28
Added ReferenceResolverProvider to JsonSerializer to allow safe caching

### LINKED ISSUE #870: Concurrency issue in DefaultReferenceResolver/JsonSerializer
Just saw the exception below with JsonSerializer being used in a multi-threaded environment, which I believe is due to the _referenceCount++ in DefaultReferenceResolver.GetReference() not being thread-safe (should be using Interlocked.Increment).
Also the internal dictionaries in BiDirectionalDictionary are not thread-safe either, and there's no synchronization mechanism around them.

System.ArgumentException: A different value already has the Id '181186'.
   at Newtonsoft.Json.Utilities.BidirectionalDictionary`2.Set(TFirst first, TSecond second)
   at Newtonsoft.Json.Serialization.DefaultReferenceResolver.GetReference(Object context, Object value)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.GetReference(JsonWriter writer, Object value)
   --- End of inner exception stack trace ---
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.GetReference(JsonWriter writer, Object value)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.WriteReferenceIdProperty(JsonWriter writer, Type type, Object value)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.WriteObjectStart(JsonWriter writer, Object value, JsonContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeObject(JsonWriter writer, Object value, JsonObjectContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeList(JsonWriter writer, IEnumerable values, JsonArrayContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeObject(JsonWriter writer, Object value, JsonObjectContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeList(JsonWriter writer, IEnumerable values, JsonArrayContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeObject(JsonWriter writer, Object value, JsonObjectContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeObject(JsonWriter writer, Object value, JsonObjectContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeObject(JsonWriter writer, Object value, JsonObjectContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.Serialize(JsonWriter jsonWriter, Object value, Type objectType)
   at Newtonsoft.Json.JsonSerializer.SerializeInternal(JsonWriter jsonWriter, Object value, Type objectType)

### LINKED ISSUE #1391: Add `ReferenceResolverProvider` to `JsonSerializer` to allow safe caching
Since `JsonSerializer`'s default `ReferenceResolver` is stateful I'd like a `ReferenceResolverProvider` added to `JsonSerializer` as was done for `JsonSerializationSettings`. This will prevent reference ids continuing to increment with each serialization call instead of starting at 1 for each.

### LINKED ISSUE #1452: Possible thread-unsafe serialization - DefaultReferenceResolver
Application I am developing has been throwing for some time intermittent exceptions:
> Newtonsoft.Json.JsonSerializationException: Error writing object reference for '<TYPE>'. Path '$values[34]'. ---> System.ArgumentException: A different value already has the Id '<ID>'.
I tried reproducing in debug sessions, but I couldn't get any results. After that, I started looking at the code based on the stack trace, which lead to method [https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Serialization/JsonSerializerInternalWriter.cs#L366](JsonSerializerInternalWriter.GetReference). From there, I went to [https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Serialization/DefaultReferenceResolver.cs#L69](DefaultReferenceResolver.GetReference). I suspect that _JsonSerializer_ is not thread-safe because of this resolver. Although _JsonSerializerInternalWriter_ is instantiated for each serialization/deserialization process (therefore the reference mappings are also "local"), the resolver is provided by _JsonSerializer_ via method [https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/JsonSerializer.cs#L1150](JsonSerializer.GetReferenceResolver), which means that the resolver is reused between different threads (which could be parallel). Since the **__referenceCount_** field is reused, any access to it should be locked, as two parallel threads might start incrementing and using the same value for reference. Can you please confirm my suspicions and maybe fix this in next release? Thank you!

### Steps to reproduce

Highly unlikely to reproduce, as the referenceCount incrementation must occur at almost the same time on two parallel threads.

TylerBrinkley:
Fixes #1391
Fixes #1452 
Fixes #870


JamesNK:
The reason I left JsonSerializer with a property that returns IReferenceResolver is the property can be used inside a JsonConverter's WriteJson/ReadJson (each has a JsonSerializer parameter). A func that returns a new instance each time will break people who want to use the resolver inside a converter.

TylerBrinkley:
Hmm, could `ReferenceResolver` be added to `JsonReader` and `JsonWriter` instead?

JamesNK:
No. References are specific to serialization.

TylerBrinkley:
I feel you could use the same argument for `DateFormatHandling` but it's still exposed on `JsonWriter`.

It seems to me that `JsonReader`s and `JsonWriter`s are not used concurrently and thus would work well to expose the currently used `IReferenceResolver`. As such I've updated the pull request to support that.

TylerBrinkley:
Hmm, maybe we could use a `JsonSerializerProxy` for each serialization call to pass to `JsonConverter`s whose `ReferenceResolver` property is specific to that serialization call?

TylerBrinkley:
I just implemented what I proposed above. It looks like a `JsonSerializerProxy` was already being passed to `JsonConverter`s so no changes needed there.

TylerBrinkley:
I just realized you baked your `DefaultReferenceResolver`'s state into `JsonSerializerInternalBase`'s `DefaultReferenceMappings` property. The only state that is not baked in is the reference count hence the creation of #1391. As opposed to baking the reference count into `JsonSerializerInternalBase` it seems like the right thing to do in this situation would be to move that state into the resolver itself and use the changes I have made. This way custom `IReferenceResolver`s will work just as well as the `DefaultReferenceResolver`.

TylerBrinkley:
Resolved merge conflicts

TylerBrinkley:
@JamesNK Please review when you have the time. Thanks.

Terebi42:
I am experiencing this bug in production, please accept this PR

dtila:
This PR does not fixes the issue, so don't merge.
There is still an issue with the threading and the bidirectional mapping. The first equality comparer is the default, and the second is a reference only. For class files, it will causes an unbalanced dictionary for classes implementing IEqutable.

TylerBrinkley:
Closing due to age.

PR #1410, opened 2017-08-28
Lazy async statemachine in WriteToAsync

benaadams:
Its a bit horrible... but...

### DIFF
--- a/Src/Newtonsoft.Json/Linq/JObject.Async.cs
+++ b/Src/Newtonsoft.Json/Linq/JObject.Async.cs
@@ -45,7 +45,7 @@ namespace Newtonsoft.Json.Linq
         public override Task WriteToAsync(JsonWriter writer, CancellationToken cancellationToken, params JsonConverter[] converters)
         {
             var t = writer.WriteStartObjectAsync(cancellationToken);
-            if (t.Status != TaskStatus.RanToCompletion)
+            if (!t.IsCompletedSucessfully())
             {
                 return AwaitStartObject(t, writer, cancellationToken, converters);
             }
@@ -53,7 +53,7 @@ namespace Newtonsoft.Json.Linq
             for (int i = 0; i < _properties.Count; i++)
             {
                 t = _properties[i].WriteToAsync(writer, cancellationToken, converters);
-                if (t.Status != TaskStatus.RanToCompletion)
+                if (!t.IsCompletedSucessfully())
                 {
                     return AwaitProperties(t, i + 1, writer, cancellationToken, converters);
                 }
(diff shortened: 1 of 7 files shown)


JamesNK:
That is horrible. Does it help performance much?

Src/Newtonsoft.Json/Utilities/AsyncUtils.cs
```diff
@@ -92,6 +92,16 @@ public static Task<int> ReadAsync(this TextReader reader, char[] buffer, int ind
             Debug.Assert(reader != null);
             return cancellationToken.IsCancellationRequested ? FromCanceled<int>(cancellationToken) : reader.ReadAsync(buffer, index, count);
         }
+
+        public static bool IsCompletedSucessfully(this Task task)
+        {
+            // IsCompletedSucessfully is the faster method, but only currently exposed on .NET Core 2.0
+#if NETCOREAPP2_0
```
TylerBrinkley: I find it unlikely @JamesNK will want to target .NET Core given the availability of .NET Standard. Is `Task.Status` all that slow?

benaadams: Is why I didn't add a target :)

[`Task.Status`](https://github.com/dotnet/coreclr/blob/3758e847988cf8d0e950205657dd0bbdd6818fe6/src/mscorlib/src/System/Threading/Tasks/Task.cs#L1338-L1360) is a pretty ugly constructed property that doesn't inline

benaadams: It was an internal property previously that was exposed post .NET Standard 2.0 https://github.com/dotnet/corefx/issues/16745

TylerBrinkley: So the `NETCOREAPP2_0` directive is in case he wants to add it later or for users that source import Json.NET?

Yeah, `Task.Status` does look pretty ugly. I guess an alternative would be `t.IsCompleted && !t.IsFaulted && !t.IsCanceled` which might inline.

benaadams: Yeah, and moving the test to one place (so for example if changed to the 3 tests, instead of `.Status == TaskStatus.RanToCompletion` its only one place that changes - though the 3 tests is pretty horrible too :)

JamesNK: I don't plan to add .NET Core 2 target but centralizing testing task completion like this is a good change.

Src/Newtonsoft.Json/Linq/JObject.Async.cs
```diff
@@ -41,16 +42,54 @@ public partial class JObject
...
+                t = _properties[i].WriteToAsync(writer, cancellationToken, converters);
+                if (!t.IsCompletedSucessfully())
+                {
+                    return AwaitProperties(t, i + 1, writer, cancellationToken, converters);
+                }
+            }
+
+            t = writer.WriteEndObjectAsync(cancellationToken);
+
+            return (t.IsCompletedSucessfully()) ? AsyncUtils.CompletedTask : AwaitEndObject(t);
```
JamesNK: Is this needed? Couldn't you just return the Task from `writer.WriteEndObjectAsync(cancellationToken);`?

JamesNK:
Cleaned sine stuff up https://github.com/JamesNK/Newtonsoft.Json/commit/fe2ec84c18b74e242765193e1c4a9f1a6fc6bd31

Src/Newtonsoft.Json/Linq/JObject.Async.cs
benaadams: Yeah... is what you get for over thinking :)

benaadams: Or underthinking... :/

PR #1413, opened 2017-08-29
[Wip] Slimmer JPropertyKeyedCollection

benaadams:


JamesNK:
A simpler solution, and one I have been thinking about for a while is not tracking keys in a dictionary until someone actually uses them.

JamesNK:
Hmmm

Writing JSON doesn't care about the key - it is looping over collections and writing the content, but I guess parsing will always care because of duplicate keys, and having to check for that by the key.

Src/Newtonsoft.Json/Linq/JPropertyKeyedCollection.cs
```diff
@@ -280,5 +247,637 @@ public bool Compare(JPropertyKeyedCollection other)
...
+                    return this;
+                }
+
+                protected override bool TryGetNext(ref int index, out KeyValuePair<string, JToken> value)
+                {
+                    value = default(KeyValuePair<string, JToken>);
+                    return false;
+                }
+
+                public override ICollection<string> Keys => new string[0];
```
TylerBrinkley: Perhaps use `Array.Empty<T>()` where available?

Src/Newtonsoft.Json/Linq/JPropertyKeyedCollection.cs
```diff
@@ -280,5 +247,637 @@ public bool Compare(JPropertyKeyedCollection other)
...
+
+                public override ICollection<JToken> Values => new JToken[0];
```
TylerBrinkley: Perhaps use `Array.Empty<T>()` where available?

JamesNK:
I'm going to close this. If you make further progress on the issue I raised above you are welcome to submit a new PR.

PR #1416, opened 2017-09-02
Avoid packing test projects

AArnott:
This makes it so when I use `msbuild /t:pack` at the command line, it only packs the product assembly and not the test projects.


JamesNK:
Makes sense. Done.

PR #1417, opened 2017-09-03
Add AppVeyor.yml file

AArnott:
Adds the build definition for building and testing to source code so forks can readily be built by appveyor.


JamesNK:
Hmm, if I did add this the build should probably use `build\runbuild.cmd`. That is the "offical" way to build, test and pack Json.NET

AArnott:
OK, I've switched it up to use runbuild.cmd. I agree consolidating on one official way to build is a good thing.

Two things I'd like to suggest you consider though:

1. Your script seems pretty verbose. The log is huge when I build with that. I wonder if you'd consider offering switches to your script to reduce the logging verbosity. 
1. I wonder what the use is for all the additional copying, etc that your script does beyond what MSBuild does. I personally prefer for a repo's 'official' build steps to be `msbuild /t:restore && msbuild /t:build,pack`, and for these steps to be fast enough that it can be the same whether done in VS or from the command line. For a library such as yours I believe this is within reach, but I don't know what all the copying your script does is for, so perhaps I'm wrong.

JamesNK:
1. I don't care but if someone wants to add options then I don't mind
2. Source files are copied when the script is run so I can keep working while a build is running. Also when I first wrote the build script the preferred acquisition method was downloading a zip file, so the build script zips up dlls and then unzips to a test directory to simulate testing the real-world files. That extra step should be removed.

AArnott:
If I added a switch to skip the copying then, do you have a presence regarding which mode is default?

JamesNK:
The switch to suppress taking a copy of the source code prior to build should default to current behavior.

I'm happy to completely removing the `Deploy` step and this copy - https://github.com/JamesNK/Newtonsoft.Json/blob/master/Build/build.ps1#L244-L250
There is no reason not to run the NUnit tests in the build test directory.

JamesNK:
Remove deploy and some copies - https://github.com/JamesNK/Newtonsoft.Json/commit/a05864d40d93d24a76cf313f5d0617483130001f

AArnott:
Cool. Is this good to complete then? Anything else?

AArnott:
Actually it seems that the build script doesn't produce a newtonsoft.json.nupkg file. Instead, the artifacts captured are a couple of dependencies. We should stop collecting these artifacts, and (preferably) collect a newtonsoft.json.nupkg artifact. What are your thoughts, @JamesNK ?

JamesNK:
A NuGet package is produced if buildNuGet is true.

One thing I'd still like to do is have an argument for controlling the msbuild verbosity. A lot of the remaining log is from msbuild.

AArnott:
OK, I've added a switch for msbuild verbosity (and it defaults to minimal).
The logging is still very long because of the `dotnet test` runner which evidently doesn't redirect STDOUT which I guess the other runners you use do. 

I see you've already set buildNuGet=true by default.

I'll push in a few minutes.

JamesNK:
Yeah I couldn't find a way to redirect `dotnet test` output.

AArnott:
>  I couldn't find a way to redirect Console.WriteLine to a file with dotnet test

Does NUnit offer an API to print messages to that gets captured in its test results? xunit does, so I would hope NUnit does also.

JamesNK:
https://github.com/nunit/docs/wiki/Console-Command-Line

`--out`

AArnott:
> I changed the build script a week ago to send print messages to a file with --out arg instead of the log file.

Why then do we still see it at the console as well?

I think this PR is ready.

JamesNK:
I'll get in merged tonight or this weekend.

JamesNK:
Looks good to me. Between your changes and mine the build output has gone from 11791 lines to 2623.

Merged

JamesNK:
https://github.com/JamesNK/Newtonsoft.Json/commit/5cdad2e2752a370234453d384aac3ffde2bc8eff

Build is now 923 lines. Good enough!

PR #1418, opened 2017-09-03
Use Nerdbank.GitVersioning

AArnott:
This makes every commit produce a unique package version, while preserving control of all 3 integers of the version.

Before this change, the source code built to a package version that was 10.0.1, even though 10.0.3 is already released.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ _ReSharper.*
 *.nuget.targets
 *.orig
 .DS_Store
+*.log
--- a/Src/Directory.Build.props
+++ b/Src/Directory.Build.props
@@ -1,0 +1,5 @@
+<Project>
+  <ItemGroup>
+    <PackageReference Include="Nerdbank.GitVersioning" Version="2.0.37-beta" PrivateAssets="all" />
+  </ItemGroup>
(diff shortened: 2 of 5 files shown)


AArnott:
This isn't counting version height properly. I'll look into it.

AArnott:
It appears to be due to your appveyor build script not maintaining consistent casing. In git, your folder is `Src` but if your build script uses `src` anywhere, the git version height won't be calculated correctly.

JamesNK:
Versions are updated by runbuild.cmd. I don't really want to introduce a second method.

AArnott:
I will take a look at that script. Suppose we remove the original method you use, so that the method you use matches what many other projects use rather than a custom built solution?

AArnott:
...and it means the version is correct even when built from VS or with a standard msbuild.exe invocation.

304NotModified:
It seems there are no plans to merge this one (with good reasons)? Then I would recommend to close this PR.

JamesNK:
Thanks for this but I'm happy with the current versioning scheme.

304NotModified:
Good choice IMO.

PR #1420, opened 2017-09-03
Fix typo in build.ps1 and use nuget.config

AArnott:


Build/build.ps1
```diff
@@ -193,12 +193,13 @@ function EnsureNuGetExists()
...
 }
 
-function EnsureNuGetPacakge($packageName, $packagePath, $packageVersion)
+function EnsureNuGetPackage($packageName, $packagePath, $packageVersion)
 {
   if (!(Test-Path $packagePath))
   {
     Write-Host "Couldn't find $packagePath. Downloading with NuGet"
-    exec { & $nugetPath install $packageName -OutputDirectory $buildDir\Temp -Version $packageVersion | Out-Default } "Error restoring $packagePath"
+    $configFilePath = Resolve-Path "$PSScriptRoot\..\Src\nuget.config"
```
JamesNK: What about just `"$sourceDir\nuget.config"`

JamesNK:
Was this being missing causing any issues?

AArnott:
> Was this being missing causing any issues?

Yes. On my machine I have nuget package sources added machine-wide that require authentication. So your package install that runs from your runbuild script failed because no auth token was available.

Build/build.ps1
AArnott: Ah, good idea. Done.

JamesNK:
Okie dokie

PR #1421, opened 2017-09-03
Read only source tree

AArnott:
This keeps the source tree clean.

### DIFF
--- a/Build/build.ps1
+++ b/Build/build.ps1
@@ -101,7 +101,7 @@ task Package -depends Build {
   {
     $finalDir = $build.Framework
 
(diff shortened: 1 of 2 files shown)


JamesNK:
Are build files in the source tree causing any issues? Everything is correctly ignored, any there is always going to be apps like VS or merge tools writing files there.

AArnott:
>Are build files in the source tree causing any issues?

No, not really. It's more a convention or philosophical argument rather than anything else. I like to see source under 'src' and binaries under 'bin' rather than binaries appearing under src. Also, it can be more convenient to finding a particular binary when it doesn't require searching all the source directories as well.

But if you don't buy into that, feel free to close the PR. It's not a big enough deal that I would debate it.

PR #1450, opened 2017-09-29
Reuse variable

zivillian:
While researching the DateTimeConverter details, I stumbled across this redundant call to `ReflectionUtils.IsNullableType`


JamesNK:
Good spotting. Merged.

Issue #1245, opened 2017-03-15
JsonSerializerSettings defaults documentation


jminer:
Reading the documentation, I don't see what the `JsonSerializerSettings` defaults are. I started at the `JsonConvert` class, and it doesn't mention what settings it uses if you don't specify any. Going to the `JsonSerializerSettings` class, it doesn't seem to say what the defaults are either. Without trial and error, one can't know what settings need to be changed without knowing what the defaults are. For example, `CheckAdditionalContent`, `DateFormatHandling`, and `MissingMemberHandling` are some settings that don't say what their default is. It would be helpful having this information in the documentation.

JamesNK:
They're here - http://www.newtonsoft.com/json/help/html/SerializationSettings.htm

Feel free to create a PR adding them to JsonSerializerSettings properties.

JamesNK:
Done - https://github.com/JamesNK/Newtonsoft.Json/commit/be032b562e05192c6ca7feb634a3fee11810ef3a

Thanks to @sebasrp

PR #1454, opened 2017-10-02
add JsonSerializerSettings defaults documentation. See issue #1245

sebasrp:
add JsonSerializerSettings defaults documentation. See issue #1245


JamesNK:
Thanks for the PR. I've updated it slightly to refer to the default enum value rather than a description to follow the .NET Framework docs standard.

I also placed default docs on JsonSerializer.

PR #1459, opened 2017-10-06
JsonTextReader.cs and work with big files.

Romasz:
We are doubling the _chars.Length when calculation newArrayLength, thus it's worth to check if we are not out of int.MaxValue.
This fix allows to work on bigger files.


JamesNK:
That is a good idea. Can you include a unit test that this change fixes?

Romasz:
@JamesNK Ok, I've tried to put someting. Just take a close look at this as it's the very first unit test I've made (yeah, I know shame on me ;) ).

JamesNK:
Thanks. I refactored the test so it didn't require a great amount of memory.

PR #1480, opened 2017-11-01
Added support for case insensitive enums

### LINKED ISSUE #1426: Error deserializing case insencitive enum whith EnumMember attribute
Hi all, I need to deserialize the enum using the EnumMember attribute and I need to ignore the case, but as a result I get an error, see the code.
`
class Program
    {
        [JsonConverter(typeof(StringEnumConverter))]
        enum LengthOfResidence
        {
            [EnumMember(Value ="other")]
            one,
            [EnumMember(Value = "two")]
            two
        }

        class test
        {
            public LengthOfResidence residence { get; set; }
        }

        static void Main(string[] args)
        {
            var json = "{residence:\"Other\"}";
            var result = JsonConvert.DeserializeObject<test>(json);

            Console.WriteLine(result.residence);
            Console.ReadKey();
        }
    }
`

sahb1239:
See: https://github.com/JamesNK/Newtonsoft.Json/commit/b2afa9d674f4dc0d5137ce6f175e06e565fc66f5#diff-5fa0fe7683d99557f112f18c7556f1aa

Fixes #1426

### DIFF
--- a/Src/Newtonsoft.Json/Utilities/EnumBidirectionalDictionary.cs
+++ b/Src/Newtonsoft.Json/Utilities/EnumBidirectionalDictionary.cs
@@ -1,0 +1,102 @@
+﻿using System;
+
+namespace Newtonsoft.Json.Utilities
+{
+    internal class EnumBidirectionalDictionary
+    {
+        private bool _hasCaseSensitiveMembers = false;
+        private readonly BidirectionalDictionary<string, string> _caseInsensitiveDictionary = 
+            new BidirectionalDictionary<string, string>(StringComparer.OrdinalIgnoreCase, StringComparer.OrdinalIgnoreCase);
+
+        private BidirectionalDictionary<string, string> _caseSensitiveDictionary;
+
+        private void InitilizeCaseSensitiveDictionary()
+        {
+            _caseSensitiveDictionary = new BidirectionalDictionary<string, string>(StringComparer.Ordinal, StringComparer.Ordinal);
+        }
+
+        public void Set(string first, string second)
+        {
+            // Add to caseSensitive if second is found, since second is used by TryResolvedEnumName in EnumUtils
+            if (_caseInsensitiveDictionary.TryGetBySecond(second, out string _))
+            {
+                if (!_hasCaseSensitiveMembers)
+                {
+                    _hasCaseSensitiveMembers = true;
+
+                    // Initilize case sensitive dictionary
+                    InitilizeCaseSensitiveDictionary();
+                }
+
+                // Add new element
+                _caseSensitiveDictionary.Set(first, second);
+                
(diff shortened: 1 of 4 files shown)


sahb1239:
Will be reopened when test case is fixed

Src/Newtonsoft.Json/Utilities/EnumUtils.cs
```diff
@@ -247,6 +247,14 @@ public static object ParseEnumName(string enumText, bool isNullable, bool disall
                 }
             }
 
+            // Try parse case sensitive
+            try
+            {
+                return Enum.Parse(t, finalEnumText, false);
+            }
+            catch(ArgumentException) { }
```
JamesNK: This will completely kill performance.

sahb1239: I have now changed this.
The method that i have chosen is using more memory and a extra initilization time instead of each time a enum is deserilized. See EnumBidirectionalDictionary

sahb1239:
Please see build https://ci.appveyor.com/project/JamesNewtonKing/newtonsoft-json/build/1.0.1364
Exception: The active test run was aborted. Reason: Could not convert string to DateTimeOffset: 2017. Path 'NewTestResults[0].StartTime'.

I'm not sure why the exception has occured

Src/Newtonsoft.Json/Utilities/EnumBidirectionalDictionary.cs
```diff
@@ -0,0 +1,102 @@
...
+                    _hasCaseSensitiveMembers = true;
+
+                    // Initilize case sensitive dictionary
+                    InitilizeCaseSensitiveDictionary();
+                }
+
+                // Add new element
+                _caseSensitiveDictionary.Set(first, second);
+                
+                // Not needed to add the current element from caseInsentive to caseSensitive since we lookup in _caseInsensitive after _caseSensitive
```
sahb1239: Side effect of not moving it is that the first enum will match all other cases.
For example enum with the following members: Abc, abc.
Abc will then match Abc
abc will then match abc
AbC will then match the first

How should I handle this? Should the user be given a exception or should it be first match like now?

Jericho:
@JamesNK @sahb1239 I have a suggestion: instead of trying to figure out if the dictionary *should* be case insensitive or not, why not simply allow developers to declare whether they want it case sensitive or not?

What I mean is that developers would be able to decorate their enum with the following attribute:
```
[JsonConverter(typeof(StringEnumConverter), caseSensitive: false)]
```

I'll submit a PR shortly so you can see how I propose implementing my idea.

sahb1239:
Another possibility could be to implement it in the EnumMemberAttribute to support enums with both case sensitive and case insensitive members which could give even more flexibility.

However I think I like the per enum configuration more which @Jericho is proposing.

sahb1239:
#1519 supports configuration of the behavior, and therefore this PR is closed.

Issue #1429, opened 2017-09-08
Add ItemNullValueHandling to JsonContainerAttribute


TylerBrinkley:
Most of the time I use `NullValueHandling.Ignore` globally as I don't have a need for `null` to be represented in the json. However, now working on an api client library which needs `null`s I find it inconvenient to have to specify `NullValueHandling.Ignore` on each property of whole classes that can't have serialized `null` properties. I would appreciate it if an `ItemNullValueHandling` property was added to `JsonContainerAttribute` that controls the `NullValueHandling` for each property. I would suppose it should work for collections as well.

### Current

```csharp
class ClassWithNoSerializedNulls
{
    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
    public string A { get; set; }
    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
    public ConsoleColor? B { get; set; }
    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
    public DateTime? C { get; set; }
    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
    public Version D { get; set; }
    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
    public Uri E { get; set; }
}
```

### Would Like

```csharp
[JsonObject(ItemNullValueHandling = NullValueHandling.Ignore)]
class ClassWithNoSerializedNulls
{
    public string A { get; set; }
    public ConsoleColor? B { get; set; }
    public DateTime? C { get; set; }
    public Version D { get; set; }
    public Uri E { get; set; }
}
```

JamesNK:
That would be a worthwhile feature. Should also add ItemNullValueHandling to JsonPropertyAttribute at the same time.

mdg1019:
I'm working on a large project and would save me a lot of pain.

paulirwin:
+1 for this. Wouldn't naming it `NullValueHandling` rather than `ItemNullValueHandling` help it match the same property on `JsonSerializerSettings` and `JsonPropertyAttribute`? I don't think the `Item` prefix is necessary as since the attribute is at the class level that would imply that the setting applies to its members. (In other words, if the object itself were null, the attribute wouldn't apply anyways.)

And @JamesNK - `JsonPropertyAttribute` already has `NullValueHandling`. What would the difference of `ItemNullValueHandling` be?

TylerBrinkley:
[...]

PR #1482, opened 2017-11-02
Add ItemNullValueHandling to JsonObjectAttribute

paulirwin:
Resolves the main issue in #1429 - I would recommend spinning off separate issues for the other things discussed in the thread. This PR adds an ItemNullValueHandling property to JsonObjectAttribute that controls null value handling for all properties of that object. This reduces the need for multiple redundant JsonPropertyAttributes. So the heirarchy is JsonPropertyAttribute overrides JsonObjectAttribute overrides JsonSerializerSettings.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/NullValueHandlingTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/NullValueHandlingTests.cs
@@ -40,6 +40,20 @@ namespace Newtonsoft.Json.Tests.Serialization
     [TestFixture]
     public class NullValueHandlingTests : TestFixtureBase
     {
+        private const string MovieNullValueHandlingIncludeExpectedResult = @"{
+  ""Name"": ""Bad Boys III"",
+  ""Description"": ""It's no Bad Boys"",
+  ""Classification"": null,
+  ""Studio"": null,
+  ""ReleaseDate"": null,
+  ""ReleaseCountries"": null
+}";
+
+        private const string MovieNullValueHandlingIgnoreExpectedResult = @"{
+  ""Name"": ""Bad Boys III"",
+  ""Description"": ""It's no Bad Boys""
+}";
+
 #if !NET20
         [Test]
         public void DeserializeNullIntoDateTime()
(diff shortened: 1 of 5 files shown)


TylerBrinkley:
The `ItemNullValueHandling` property should be added to the `JsonContainerAttribute` not the `JsonObjectAttribute`. Also I think the PR should include the addition of the `ItemNullValueHandling` property to the `JsonPropertyAttribute` instead of separating it.

paulirwin:
@TylerBrinkley `JsonContainerAttribute` has three derived classes: `JsonObjectAttribute`, `JsonArrayAttribute`, and `JsonDictionaryAttribute`. For `JsonObjectAttribute`, per your example on #1429, the meaning of `ItemNullValueHandling` is clear. It is less clear for the latter two. Would `ItemNullValueHandling` on `JsonArrayAttribute` mean that null items are left out of the array? That would lead to surprising and unexpected behavior, not to mention likely bugs. For `JsonDictionaryAttribute` I assume that would mean that key-value-pairs with values that are null are left out of serialization, which makes more sense than applying it to `JsonArrayAttribute`.

As for the `JsonPropertyAttribute` change, I'll see what I can do but that is a much more involved change and touches a completely different area of the code.

TylerBrinkley:
@paulirwin I'm thinking you're right in putting it on `JsonObjectAttribute` as currently a `NullValueHandling.Ignore` on `JsonSerializerSettings` does not apply to collection items. See below.

```c#
JsonConvert.SerializeObject(new { array = new[] { "dog", null, "cat" }, date = (DateTime?)null, dictionary = new Dictionary<string, Uri> { { "first", new Uri("https://abc.com") }, { "second", null } } }, new JsonSerializerSettings { NullValueHandling = NullValueHandling.Ignore })
```
results in

```js
{"array":["dog",null,"cat"],"dictionary":{"first":"https://abc.com","second":null}}
```

And in that case I'm not sure we would need to add it to `JsonPropertyAttribute`.

@JamesNK thoughts?

paulirwin:
@TylerBrinkley Good point. In other words, I view `ItemNullValueHandling` on `JsonObjectAttribute` more in line with `MemberSerialization` than `ItemTypeNameHandling`.

JamesNK:
NullValueHandling only applies to properties, not array or dictionary items so `JsonObjectAttribute` is better than the base class.

JamesNK:
I modified the logic slightly but otherwise it looks good. Merged. Thanks!

Issue #1479, opened 2017-11-01
Unexpected behaviour when using custom JsonConverters


joelwkall:
When having multiple JsonConverters defined in the Converters property of JsonSerializer, the converter chosen for serialization is the first JsonConverter where CanConvert returns true for the type, regardless of whether CanWrite is true. If CanWrite is false, the default serialization will be used, even if there are JsonConverters further down the list that could handle serialization of the type.

This is probably working as defined, however a more sensible behavior (in my opinion) would be to ignore that JsonConverter and fall through and check if the next one can handle serialization for the type. Otherwise, it is impossible to have one JsonConverter that handles serialization and another that handles deserialization for the same type.

### Source/destination types

```csharp
Any type that has at least 2 custom JsonConverters defined for it, where the first one has CanWrite=false and the second (or later) has CanWrite=true.
```

### Source/destination JSON

```javascript
JSON generated by custom JsonConverter
```

### Expected behavior

Using the first JsonConverter where CanConvert returns true for the type, and where CanWrite returns true. The Converters list should be iterated until a JsonConverter is found that returns true for both.

### Actual behavior

The first JsonConverter where CanConvert returns true for the type is selected, regardless of what CanWrite returns. If CanWrite is false, the default serialization is invoked.

### Steps to reproduce

```csharp
public class CustomConverterA : JsonConverter
{
    public override bool CanWrite
    {
        get
        {
            Console.WriteLine("Inside CustomConverterA.CanWrite (returns false)");
            return false;
        }
    }
    public override bool CanRead => true;

    public override bool CanConvert(Type objectType)
    {
        Console.WriteLine("Inside CustomConverterA.CanConvert");

        if (objectType == typeof(string))
        {
            Console.WriteLine("return true");
            return true;
        }

        Console.WriteLine("return false");
        return false;
    }

    public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
    {
        Console.WriteLine("Inside CustomConverterA.WriteJson");
    }

    public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
    {
        Console.WriteLine("Inside CustomConverterA.ReadJson");

        return new object();
    }
}

public class CustomConverterB : JsonConverter
{
    public override bool CanWrite
    {
        get
        {
            Console.WriteLine("Inside CustomConverterB.CanWrite (returns true)");
            return true;
        }
    }
    public override bool CanRead => true;

[...]

PR #1497, opened 2017-11-13
Fix for issue #1479 - support for distinct read & write converters

wizofaus:
Allows having distinct converters for writing & reading - when finding converters for writing, attempts to find one with CanWrite = true, when finding for reading, looks for one with CanRead = true, otherwise falls back to converter that can handle type.

### DIFF
--- a/Src/Newtonsoft.Json/JsonSerializer.cs
+++ b/Src/Newtonsoft.Json/JsonSerializer.cs
@@ -1187,12 +1187,12 @@ namespace Newtonsoft.Json
             return _referenceResolver;
         }
 
-        internal JsonConverter GetMatchingConverter(Type type)
+        internal JsonConverter GetMatchingConverter(Type type, bool forWriting)
         {
-            return GetMatchingConverter(_converters, type);
+            return GetMatchingConverter(_converters, type, forWriting);
         }
 
-        internal static JsonConverter GetMatchingConverter(IList<JsonConverter> converters, Type objectType)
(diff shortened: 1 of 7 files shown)


joelwkall:
Can confirm that this change would solve my issue. Although it might break any users that depend on the current behavior. Maybe a configuration flag that you can enable if you want this new behavior?

Src/Newtonsoft.Json/JsonSerializer.cs
```diff
@@ -1187,28 +1187,32 @@ internal IReferenceResolver GetReferenceResolver()
...
                 for (int i = 0; i < converters.Count; i++)
                 {
                     JsonConverter converter = converters[i];
 
                     if (converter.CanConvert(objectType))
                     {
-                        return converter;
+                        if (forWriting ? converter.CanWrite : converter.CanRead)
+                            return converter;
+                        possibleConverter = converter;
```
joelwkall: If there is no suitable converter for writing or for reading, shouldnt we pick the first converter that can convert the type? This code looks like it picks the last one.

wizofaus: There's a case for either, this was just less code :-)

joelwkall: Allright, but the old behavior is to choose the first one, I think. So it might be a good idea to keep that, to make the change footprint as small as possible and break fewer implementations.

PR #1506, opened 2017-11-18
Added exception handling logic, in order for the GetTypeFromTypeNameK…

InKolev:
…ey() method to continue execution even if the assembly is not found in the root directory of the executable file (in order to check later in the loaded assemblies collection if the required Assembly can be found).

### DIFF
--- a/Src/Newtonsoft.Json/Serialization/DefaultSerializationBinder.cs
+++ b/Src/Newtonsoft.Json/Serialization/DefaultSerializationBinder.cs
@@ -69,9 +69,27 @@ namespace Newtonsoft.Json.Serialization
                 assembly = Assembly.LoadWithPartialName(assemblyName);
 #pragma warning restore 618,612
 #elif DOTNET || PORTABLE
-                assembly = Assembly.Load(new AssemblyName(assemblyName));
+                try 
+	            {	        
+                    assembly = Assembly.Load(new AssemblyName(assemblyName));
+	            }
+	            catch (System.IO.FileNotFoundException)
+	            {
+                    // we catch this exact exception without rethrowing it, 
+                    // because we want to check if the required Assembly is already loaded
+                    // in the current AppDomain, in order to retrieve it from there
+	            }
 #else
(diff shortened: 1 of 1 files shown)


InKolev:
We accidentally found this bug while running one of our microservices.
We have an application, which loads a given microservice (located in a folder different  than the one of the current executable).
When trying to deserialize a message with a "$type" handling property which points to an assembly from a folder, different than the root one, the **DefaultSerializationBinder** failed with **FileNotFoundException** before it tried to check if the **Assembly is already loaded** in the **app domain**.

JamesNK:
Hi

I'm not a fan of those approach because I don't like having try/catch is normal program flow. Throwing exceptions destroys performance. Is there a better way to do this?

Also you haven't provided an example of what it fixes in a unit test.

I think this sort of edge case might be something that is best left to people creating their own `SerializationBinder`.

InKolev:
Hi, 

I understand your concerns, they are absolutely valid.
I just want to mention that I am not throwing any exceptions here, It's the Assembly.Load() method that does this. I just catch it in order to continue with the expected flow.

There is another way to solve this problem.
Using a **custom assembly resolver**, we can attach an **event handler** to **AppDomain.AssemblyResolve** event, and extract the remaining logic there. I think this is the appropriate way to approach this issue.

For example:  
```cs
// Attach the event handler
AppDomain.AssemblyResolve += OnAssemblyResolve;

// The handler itself
private Assembly OnAssemblyResolve(object sender, ResolveEventArgs args)
{
            var requestedAssemblyName = args.Name;
            
            // Try to find the requested assembly inside the list of assemblies 
            // that are already loaded into the execution context of the provided AppDomain.
            Assembly[] loadedAssemblies = AppDomain.GetAssemblies();
                    foreach (Assembly a in loadedAssemblies)
                    {
                        // check for both full name or partial name match
                        if (assembly.FullName == requestedAssemblyName || assembly.GetName().Name == requestedAssemblyName )
                        {
                            return assembly;
                        }
                    }

            return null;
}

```

PR #1519, opened 2017-11-27
Solution for issue 1516 - Configurable case sensitivity StringEnumConverter

### LINKED ISSUE #1516: StringEnumConverter case sensitivity
Up until Json.net v9, the StringEnumConverter class was case insensitive and it was changed in v10. Here's the [discussion](https://github.com/JamesNK/Newtonsoft.Json/issues/1067) explaining why it was changed and here's the [commit](https://github.com/JamesNK/Newtonsoft.Json/commit/b2afa9d674f4dc0d5137ce6f175e06e565fc66f5) released in v10 that changed the behavior to case-sensitive.

I can't argue with the reasoning behind this change but I find myself in a situation where case-insensitivity is desired and it seems like I can't change the comparison behavior. I'm stuck, as far as I can tell.

I was wondering if we could make case-sensitivity configurable? We could ensure the default behavior is to be case-sensitive to match the current behavior and also allow developers to indicate they want a different behavior. 

Does that sound reasonable? I'm willing to submit PR with this improvement.

Jericho:
Allow developers to declare if they want "string to enum" conversion to be case sensitive or not

Resolves #1516

### DIFF
--- a/Src/Newtonsoft.Json/Utilities/EnumUtils.cs
+++ b/Src/Newtonsoft.Json/Utilities/EnumUtils.cs
@@ -191,25 +191,35 @@ namespace Newtonsoft.Json.Utilities
             }
 
             string finalEnumText;
+            bool ignoreCase = true;
 
             EnumBidirectionalDictionary map = EnumMemberNamesPerType.Get(t);
             if (TryResolvedEnumName(map, enumText, out string resolvedEnumName, caseSensitive))
             {
                 finalEnumText = resolvedEnumName;
+
+                // We found an exact match, therefore Enum.Parse must be case sensitive
+                ignoreCase = false;
             }
             else if (enumText.IndexOf(',') != -1)
             {
+                bool allEnumNamesResolved = true;
                 string[] names = enumText.Split(',');
                 for (int i = 0; i < names.Length; i++)
                 {
                     string name = names[i].Trim();
-
-                    names[i] = TryResolvedEnumName(map, name, out resolvedEnumName, caseSensitive)
-                        ? resolvedEnumName
-                        : name;
+                    bool enumNameResolved = TryResolvedEnumName(map, name, out resolvedEnumName, caseSensitive);
+                    names[i] = enumNameResolved ? resolvedEnumName : name;
+                    allEnumNamesResolved &= enumNameResolved;
                 }
 
                 finalEnumText = string.Join(", ", names);
+
+                if (allEnumNamesResolved)
+                {
+                    // We found an exact match for all items, therefore Enum.Parse must be case sensitive
+                    ignoreCase = false;
+                }
             }
             else
             {
@@ -226,7 +236,7 @@ namespace Newtonsoft.Json.Utilities
                 }
             }
 
-            return Enum.Parse(t, finalEnumText, true);
+            return Enum.Parse(t, finalEnumText, ignoreCase);
         }
 
         public static string ToEnumName(Type enumType, string enumText, bool camelCaseText, bool caseSensitive)
--- a/Src/Newtonsoft.Json/Utilities/EnumUtils.cs
+++ b/Src/Newtonsoft.Json/Utilities/EnumUtils.cs
@@ -38,13 +38,11 @@ namespace Newtonsoft.Json.Utilities
 {
     internal static class EnumUtils
     {
-        private static readonly ThreadSafeStore<Type, BidirectionalDictionary<string, string>> EnumMemberNamesPerType = new ThreadSafeStore<Type, BidirectionalDictionary<string, string>>(InitializeEnumType);
+        private static readonly ThreadSafeStore<Type, EnumBidirectionalDictionary> EnumMemberNamesPerType = new ThreadSafeStore<Type, EnumBidirectionalDictionary>(InitializeEnumType);
 
-        private static BidirectionalDictionary<string, string> InitializeEnumType(Type type)
+        private static EnumBidirectionalDictionary InitializeEnumType(Type type)
         {
-            BidirectionalDictionary<string, string> map = new BidirectionalDictionary<string, string>(
-                StringComparer.Ordinal,
-                StringComparer.Ordinal);
+            EnumBidirectionalDictionary map = new EnumBidirectionalDictionary();
 
             foreach (FieldInfo f in type.GetFields(BindingFlags.Public | BindingFlags.Static))
             {
@@ -60,7 +58,7 @@ namespace Newtonsoft.Json.Utilities
                 n2 = f.Name;
 #endif
 
-                if (map.TryGetBySecond(n2, out _))
+                if (map.TryGetBySecond(n2, out _, true))
                 {
                     throw new InvalidOperationException("Enum name '{0}' already exists on enum '{1}'.".FormatWith(CultureInfo.InvariantCulture, n2, type.Name));
                 }
@@ -185,7 +183,7 @@ namespace Newtonsoft.Json.Utilities
             return values;
         }
 
-        public static object ParseEnumName(string enumText, bool isNullable, bool disallowValue, Type t)
+        public static object ParseEnumName(string enumText, bool isNullable, bool disallowValue, Type t, bool caseSensitive)
         {
             if (enumText == string.Empty && isNullable)
             {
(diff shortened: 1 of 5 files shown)


Src/Newtonsoft.Json/Utilities/EnumUtils.cs
```diff
@@ -231,16 +229,16 @@ public static object ParseEnumName(string enumText, bool isNullable, bool disall
             return Enum.Parse(t, finalEnumText, true);
```
sahb1239: If you change the third parameter and change it to false for Enum.Parse you will also support enum members which only differs in case.

See testcase DeserializeEnumWithDifferentCases

Since the EnumBidirectionalDictionary now support case insentitivity it will automatically match the required case since it loads the enumMember from the EnumBidirectionalDictionary.

2 other test cases can be seen here.
`
        [JsonConverter(typeof(StringEnumCaseInsensitiveConverter))]
        public enum EnumMemberWithDiffrentCases
        {
            [EnumMember(Value = "first_value")]
            First,
            [EnumMember(Value = "First_value")]
            first
        }

        [Test]
        public void DeserializeEnumMemberWithDifferentCasing_ByEnumMemberValue_First()
        {
            var e = JsonConvert.DeserializeObject<EnumMemberWithDiffrentCases>(@"""first_value""", new StringEnumCaseInsensitiveConverter());
            Assert.AreEqual(EnumMemberWithDiffrentCases.First, e);
        }

        [Test]
        public void DeserializeEnumMemberWithDifferentCasing_ByEnumMemberValue_Second()
        {
            var e = JsonConvert.DeserializeObject<EnumMemberWithDiffrentCases>(@"""First_value""", new StringEnumCaseInsensitiveConverter());
            Assert.AreEqual(EnumMemberWithDiffrentCases.first, e);
        }
`

sahb1239:
I would prefer that the default was case insensitive and when required case sensitivity can be turned on. This will be more simple and work for the most developers, 

This will however will break backwards compatability, however not in the testcase DeserializeEnumWithDifferentCases which I have explained in another comment.

sahb1239:
This case here does not work. This is due to the fact that you are not adding the member when no other match was found in both first and second. This gives the issue that if first or second is case insensitive equal it will not add the case to the dictionary. (This is required otherwise it will throw)

However to fix this please see: https://github.com/sahb1239/Newtonsoft.Json/blob/31b740ad8b43d8af867e533c087dff24e906c301/Src/Newtonsoft.Json/Utilities/EnumBidirectionalDictionary.cs and update the TryGetFirst and TryGetSecond methods.

The testcases to illustrate the problem (only the second fails):
```
        [JsonConverter(typeof(StringEnumCaseInsensitiveConverter))]
        public enum EnumMemberWithDiffrentCases
        {
            [EnumMember(Value = "first_value")]
            First,
            [EnumMember(Value = "second_value")]
            first
        }

        [Test]
        public void DeserializeEnumMemberWithDifferentCasing_ByEnumMemberValue_First()
        {
            var e = JsonConvert.DeserializeObject<EnumMemberWithDiffrentCases>(@"""first_value""", new StringEnumCaseInsensitiveConverter());
            Assert.AreEqual(EnumMemberWithDiffrentCases.First, e);
        }

        [Test]
        public void DeserializeEnumMemberWithDifferentCasing_ByEnumMemberValue_Second()
        {
            var e = JsonConvert.DeserializeObject<EnumMemberWithDiffrentCases>(@"""second_value""", new StringEnumCaseInsensitiveConverter());
            Assert.AreEqual(EnumMemberWithDiffrentCases.first, e);
        }
```

Jericho:
> I would prefer that the default was case insensitive

I would prefer that too because in my situation case-insensitivity is preferable, but that would be a break in behavior. I made the decision to respect current behavior.

> This case here does not work. 

The enum in this example you bring up has two items that differ only due to their casing. Therefore in a situation like this you want the case sensitive converter. Case insensitive converter will not work with such an enum.

sahb1239:
DeserializeEnumMemberWithDifferentCasing_ByEnumMemberValue_Second throws a JsonSerializationException with the message Error converting value "second_value" to type.

I think it would be better to throw a exception when initilizing the EnumBidirectionalDictionary than just to ignore a enum member.

In the unit test it has two distinct enum member values (first_value and second_value) which I would like to support deserilizing case insensitive, so I could write First_value and Second_value.

Another possibility would be to initilize _caseInsensitive in EnumBidirectionalDictionary to _caseInsensitive = new BidirectionalDictionary<string, string>(StringComparer.Ordinal, StringComparer.OrdinalIgnoreCase);. This would preserve the existing behavior of returning wrong Enum membersm however now also when EnumMembers is specified (See DeserializeEnumWithDifferentCases testcase).

This should work since the method which changes behavior is TryGetByFirst and this method is only used in ToEnumName which is used in ToEnumName, however this method recieves the enumText from a ToString("G") on the Enum value and therefore does not need to be case insensitive.

Src/Newtonsoft.Json/Utilities/EnumUtils.cs
Jericho: The third parameter should match the case-sensitivity requested by the developer (instead of being hard-coded to a given Boolean value). This would empower developers to chose which behavior they want and would also ensure the case sensitivity when converting by EnumMember value and when converting by enum item name are consistent.

However, there are 3 existing unit tests that assume `Enum.Parse` is always case insensitive despite the fact they use the case sensitive converter (which is contradictory, IMHO).

The first unit test:

```csharp
[Test]
public void DeserializeFlagEnum()
{
    string json = @"{
        ""StoreColor"": ""Red, White"",
        ""NullableStoreColor1"": 0,
        ""NullableStoreColor2"": ""black, Red, White""
    }";

    EnumClass enumClass = JsonConvert.DeserializeObject<EnumClass>(json, new StringEnumConverter());

    Assert.AreEqual(StoreColor.Red | StoreColor.White, enumClass.StoreColor);
    Assert.AreEqual((StoreColor)0, enumClass.NullableStoreColor1);
    Assert.AreEqual(StoreColor.Red | StoreColor.White | StoreColor.Black, enumClass.NullableStoreColor2);
}
```

In this example, notice that "black" is converted to `StoreColor.Black` despite the fact that a case sensitive converter is used. This is a trivial situation that can be resolved by either capitalizing the "b" or by changing to a case insensitive converter.

The other two cases are a little more complex: they involve camel casing.

```csharp
[Test]
public void CamelCaseTextFlagEnumDeserialization()
{
    string json = @"{
        ""Enum"": ""first, second""
    }";

    EnumContainer<FlagsTestEnum> c = JsonConvert.DeserializeObject<EnumContainer<FlagsTestEnum>>(json, new StringEnumConverter { CamelCaseText = true });
    Assert.AreEqual(FlagsTestEnum.First | FlagsTestEnum.Second, c.Enum);
}

[Test]
public void Deserialize_CamelCaseFromAttribute()
{
    CamelCaseEnum e = JsonConvert.DeserializeObject<CamelCaseEnum>(@"""camelCase""");
    Assert.AreEqual(CamelCaseEnum.CamelCase, e);
}
```

Notice that "first" is converted to `FlagsTestEnum.First`, "second" is converted to `FlagsTestEnum.Second` and "camelCase" is converted to `CamelCaseEnum.CamelCase`.

I'm not sure how to resolve this situation. It doesn't make sense to desire a converter that is both case sensitive **AND** allows camel cased values at the same time.

Jericho:
Let me think about your suggestion:

```
_caseInsensitive = new BidirectionalDictionary<string, string>(StringComparer.Ordinal, StringComparer.OrdinalIgnoreCase);
```

At first glance seems reasonable but I need to explore it further. I'll get back to you.

Jericho:
Took me a while but I finally have a solution that satisfies your `DeserializeEnumMemberWithDifferentCasing_ByEnumMemberValue_Second` unit test. I changed the internal case sensitive dictionary as you suggested by that was not enough. I also had to add logic to decide if Enum.Parse should be case sensitive or not. I also added the two unit tests you suggested to demonstrate that my changes address your scenario.

sahb1239:
Looks good to me.

However since this breaks some existing behavior (because of the case sensitive match now with Enum.Parse I would stille prefer that case insensitivity was the default which was the case before version 10.

I however think that is @JamesNK which should decide this.

Jericho:
@JamesNK can we get your input on this PR?

JamesNK:
Hi

Bad news is I'm going to close this PR. Good news is I will fix this issue and a case insensitive fallback will be automatic.

I'm re-implementing Enum.Parse which will avoid the current hacky solution with updating a string before passing it to Enum.Parse. I will make sure to copy over your unit tests to ensure everything your changes made will continue to work.

PR #1539, opened 2017-12-15
Use cancellationToken in two places it was missed.

JonHanna:
I'm not sure if this was a simple slip up, or if I was thinking of skipping the check for speed reasons when I did this, but as these two calls can both often happen at as the first actual write to stream, which can have the longest delay waiting for the underlying layer, they should both have the opportunity to cancel if requested.


JamesNK:
Ok. If you think of a reason why they shouldn't have the token, remove it and add a comment.

Issue #1460, opened 2017-10-10
Throw System.NullReferenceException when I serializing a DataSet


qtzheng:
### MySQL Table Script
```sql
CREATE TABLE `user` (
`sys_id`  varchar(20) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL ,
`user_id`  varchar(16) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL ,
`sap_id`  char(8) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL ,
`user_name`  varchar(100) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL ,
`password`  char(32) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL ,
`salt`  char(20) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL ,
`status`  tinyint(2) NOT NULL ,
`org_id`  int(6) NOT NULL ,
`std_id`  varchar(20) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL ,
`last_login`  datetime NOT NULL DEFAULT '0000-00-00 00:00:00' ,
`login_count`  int(11) NOT NULL ,
`create_time`  timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ,
`create_user_id`  varchar(16) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL ,
`create_user_name`  varchar(100) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL ,
`update_time`  timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ,
`update_user_id`  varchar(16) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL ,
`update_user_name`  varchar(100) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL ,
PRIMARY KEY (`sys_id`, `user_id`)
)
ENGINE=InnoDB
DEFAULT CHARACTER SET=utf8 COLLATE=utf8_bin
ROW_FORMAT=DYNAMIC
;
```
### Source/destination types

```csharp
            DataSet ds = new DataSet();
            using (DbConnection conn = this.EfContext.Database.Connection)
            {
                conn.Open();
                DbCommand cmd = conn.CreateCommand();
                cmd.CommandText = "select * from user";
                cmd.CommandType = CommandType.Text;
                cmd.Parameters.AddRange(parameters);
[...]

PR #1540, opened 2017-12-16
 Apply fix to #1460 to async writing. 

JonHanna:
Also the recursion there can be easily made iterative.

### DIFF
--- a/Src/Newtonsoft.Json/JsonWriter.Async.cs
+++ b/Src/Newtonsoft.Json/JsonWriter.Async.cs
@@ -1787,6 +1787,12 @@ namespace Newtonsoft.Json
                         return WriteValueAsync(writer, resolvedTypeCode, convertedValue, cancellationToken);
                     }
 #endif
+                    // write an unknown null value, fix https://github.com/JamesNK/Newtonsoft.Json/issues/1460
+                    if (value == null)
+                    {
+                        return writer.WriteNullAsync(cancellationToken);
+                    }
+
                     throw CreateUnsupportedTypeException(writer, value);
             }
         }
--- a/Src/Newtonsoft.Json/JsonWriter.Async.cs
+++ b/Src/Newtonsoft.Json/JsonWriter.Async.cs
@@ -1680,7 +1680,7 @@ namespace Newtonsoft.Json
 
         internal static Task WriteValueAsync(JsonWriter writer, PrimitiveTypeCode typeCode, object value, CancellationToken cancellationToken)
         {
-            for (;;)
+            while (true)
             {
                 switch (typeCode)
                 {
--- a/Src/Newtonsoft.Json/JsonWriter.Async.cs
+++ b/Src/Newtonsoft.Json/JsonWriter.Async.cs
@@ -1680,120 +1680,122 @@ namespace Newtonsoft.Json
 
         internal static Task WriteValueAsync(JsonWriter writer, PrimitiveTypeCode typeCode, object value, CancellationToken cancellationToken)
         {
-            switch (typeCode)
+            for (;;)
             {
-                case PrimitiveTypeCode.Char:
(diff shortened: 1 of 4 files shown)


JamesNK:
Opps

Can you add a unit test of what it is fixing?

Src/Newtonsoft.Json/JsonWriter.Async.cs
```diff
@@ -1680,114 +1680,122 @@ protected Task SetWriteStateAsync(JsonToken token, object value, CancellationTok
...
-                    return writer.WriteValueAsync(value == null ? (decimal?)null : (decimal)value, cancellationToken);
-                case PrimitiveTypeCode.Guid:
-                    return writer.WriteValueAsync((Guid)value, cancellationToken);
-                case PrimitiveTypeCode.GuidNullable:
-                    return writer.WriteValueAsync(value == null ? (Guid?)null : (Guid)value, cancellationToken);
-                case PrimitiveTypeCode.TimeSpan:
-                    return writer.WriteValueAsync((TimeSpan)value, cancellationToken);
-                case PrimitiveTypeCode.TimeSpanNullable:
-                    return writer.WriteValueAsync(value == null ? (TimeSpan?)null : (TimeSpan)value, cancellationToken);
+            for (;;)
```
JamesNK: I prefer `while (true)`

JonHanna: Ah yes, I recall that you do. (It'll probably never not seem strange to me)

Src/Newtonsoft.Json/JsonWriter.Async.cs
```diff
@@ -1680,114 +1680,122 @@ protected Task SetWriteStateAsync(JsonToken token, object value, CancellationTok
...
+                        if (value == null)
+                        {
+                            return writer.WriteNullAsync(cancellationToken);
+                        }
 
-                        return WriteValueAsync(writer, resolvedTypeCode, convertedValue, cancellationToken);
-                    }
-#endif
-                    throw CreateUnsupportedTypeException(writer, value);
+                        throw CreateUnsupportedTypeException(writer, value);
```
JamesNK: Should this be throwing in Task returning, non-async method? Should the ex be returned via `Task.FromException`?

JonHanna: To my mind if one can tell straight away that it's impossible to create a task that will be successful it's better to throw immediately outside of the task mechanism (comparable to how `default(IEnumerable<int>).Select(i => i)` throws immediately while `actualSequence.Select(funcThatThrows)` doesn't throw until enumeration), but I don't go so far as to make sure it happens (doing validation in a non `async` method and then calling into an `async`). I probably should though.

It would be good to be consistent, one way or another.

JamesNK: Ok, that makes sense. I might go through and audit the thrown exceptions in Task methods that aren't async.

Src/Newtonsoft.Json/JsonWriter.cs
```diff
@@ -1482,163 +1482,200 @@ protected virtual void Dispose(bool disposing)
...
-                    {
-                        // the value is a non-standard IConvertible
-                        // convert to the underlying value and retry
+                    case PrimitiveTypeCode.Decimal:
+                        writer.WriteValue((decimal)value);
+                        return;
+
+                    case PrimitiveTypeCode.DecimalNullable:
+                        writer.WriteValue((value == null) ? (decimal?)null : (decimal)value);
+                        break;
```
JamesNK: Opps https://github.com/JamesNK/Newtonsoft.Json/commit/e57cb6acea721b46c2dede63702d22f2617c51ee 😄

JonHanna: 🤦

PR #1544, opened 2017-12-18
Performance optimization for integer value serialization.

MarkPflug:
Replaces a modulus operation with a multiply and subtract, which is measurably faster. Added a benchmark to target the change.

Running the benchmark in this configuration:
.NET Framework 4.7 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.2558.0

Before:
 SerializeIntegers | 895.3 us | 1.851 us | 1.546 us |

After:
 SerializeIntegers | 726.8 us | 3.092 us | 2.892 us |

### DIFF
--- a/Src/Newtonsoft.Json/JsonTextWriter.cs
+++ b/Src/Newtonsoft.Json/JsonTextWriter.cs
@@ -854,5 +854,56 @@ namespace Newtonsoft.Json
 
             return totalLength;
         }
+
+        private void WriteIntegerValue(int value)
+        {
+            if (value >= 0 && value <= 9)
+            {
+                _writer.Write((char)('0' + value));
+            }
+            else
+            {
+                bool negative = value < 0;
+                WriteIntegerValue(negative ? (uint)-value : (uint)value, negative);
+            }
+        }
+
+        private void WriteIntegerValue(uint uvalue, bool negative)
+        {
+            if (!negative & uvalue <= 9)
+            {
+                _writer.Write((char)('0' + uvalue));
+            }
+            else
+            {
+                int length = WriteNumberToBuffer(uvalue, negative);
+                _writer.Write(_writeBuffer, 0, length);
+            }
+        }
+
+        private int WriteNumberToBuffer(uint value, bool negative)
+        {
+            EnsureWriteBuffer();
+
+            int totalLength = MathUtils.IntLength(value);
+
+            if (negative)
+            {
+                totalLength++;
+                _writeBuffer[0] = '-';
+            }
+
+            int index = totalLength;
+
+            do
+            {
+                uint quotient = value / 10;
+                uint digit = value - (quotient * 10);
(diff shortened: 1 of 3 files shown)


MarkPflug:
I found more performance, by adding method overloads for int32 values. Turns out 64 bit integer division is *especially* expensive.

On a different machine now, so benchmark baseline numbers are different:

Before:
 SerializeIntegers | 637.2 us | 0.6748 us | 0.5635 us |

After:
SerializeIntegers | 318.8 us | 1.158 us | 1.026 us |

That is bizarrely close to 50%.

Src/Newtonsoft.Json/JsonTextWriter.cs
```diff
@@ -846,8 +846,61 @@ private int WriteNumberToBuffer(ulong value, bool negative)
...
+                _writer.Write((char)('0' + value));
+            }
+            else
+            {
+                bool negative = value < 0;
+                WriteIntegerValue(negative ? (uint)-value : (uint)value, negative);
+            }
+        }
+
+        private void WriteIntegerValue(uint uvalue, bool negative)
```
JamesNK: Rename arg to value

JamesNK:
Looks good, just rename the parameter. Not sure why I had `uvalue` in the original overload.

MarkPflug:
Another minor tweak occurred to me after I logged off last night. The 64 bit path will now call over to the faster 32 bit overload when possible.

Src/Newtonsoft.Json/JsonTextWriter.cs
```diff
@@ -817,20 +817,77 @@ private void WriteIntegerValue(long value)
...
             {
-                int length = WriteNumberToBuffer(uvalue, negative);
+                int length = WriteNumberToBuffer(value, negative);
                 _writer.Write(_writeBuffer, 0, length);
             }
         }
 
         private int WriteNumberToBuffer(ulong value, bool negative)
+        {
+            if(value < uint.MaxValue)
```
TylerBrinkley: You might want to add a space between `if` and the opening parenthesis.

Src/Newtonsoft.Json/JsonTextWriter.cs
```diff
@@ -817,20 +817,77 @@ private void WriteIntegerValue(long value)
...
             {
-                int length = WriteNumberToBuffer(uvalue, negative);
+                int length = WriteNumberToBuffer(value, negative);
                 _writer.Write(_writeBuffer, 0, length);
             }
         }
 
         private int WriteNumberToBuffer(ulong value, bool negative)
+        {
+            if (value < uint.MaxValue)
```
JamesNK: Should this be `<` or `<=`?

It would be a good idea to have unit tests that test around this boundary (just smaller than MaxValue, equals MaxValue, just greater than MaxValue).

JamesNK:
Looks good to me. Thanks!

Issue #1545, opened 2017-12-19
Error parsing JSON with comments at array end (if using certain custom converter)


luanshixia:
This is a tricky error. Suppose we have the following C# definition

### Destination Types

```csharp
public class Simple
{
    [JsonProperty(Required = Required.Always)]
    public SimpleObject[] Array { get; set; }
}

[JsonConverter(typeof(LineInfoConverter))]
public class SimpleObject : JsonLineInfo
{
    public string Value { get; set; }
}

public class JsonLineInfo
{
    [JsonIgnore]
    public int? LineNumber { get; set; }

    [JsonIgnore]
    public int? LinePosition { get; set; }
}

public class LineInfoConverter : JsonConverter
{
    public override bool CanWrite
    {
        get { return false; }
    }

    public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
    {
        throw new NotImplementedException("Converter is not writable. Method should not be invoked");
    }

    public override bool CanConvert(Type objectType)
    {
        return typeof(JsonLineInfo).IsAssignableFrom(objectType);
    }

    public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
    {
        if (reader.TokenType == JsonToken.Null)
        {
            return null;
        }

        var lineInfoObject = Activator.CreateInstance(objectType) as JsonLineInfo;
        serializer.Populate(reader, lineInfoObject);

        var jsonLineInfo = reader as IJsonLineInfo;
        if (jsonLineInfo != null && jsonLineInfo.HasLineInfo())
        {
            lineInfoObject.LineNumber = jsonLineInfo.LineNumber;
            lineInfoObject.LinePosition = jsonLineInfo.LinePosition;
        }

        return lineInfoObject;
    }
}
```

and the following JSON file `simple.json`

### Source JSON

```json
{
  "array": [
    /* comment0 */
    {
      "value": "item1"
    },
    /* comment1 */
    {
      "value": "item2"
    }
    /* comment2 */
  ]
}
```

then run the following C# code

### Usage/Repro Steps

```csharp
var json = System.IO.File.ReadAllText("simple.json");
var simple = JsonConvert.DeserializeObject<Simple>(json);
```

[...]

PR #1548, opened 2017-12-21
Fixed issue #1545

luanshixia:
https://github.com/JamesNK/Newtonsoft.Json/issues/1545


JamesNK:
Thanks for including a test. Looks good!

luanshixia:
@JamesNK Thanks! May I ask when can we expect the next release (10.0.4?) which includes this fix?

Issue #1499, opened 2017-11-15
Error handler is not skipping past the error properly when handled on a higher level object


gilad61:
### Source/destination types

```csharp
public enum Color { Red, Green, Blue }

public class MyClass
{
    public IList<Color> Values { get; set; }
}
```

### Source/destination JSON

```javascript
{"Values":["Black", "Green"]}
```

### Expected behavior

In the above example we have a class with a list of Color enum objects, and a Json with an invalid color ('Black').
Normally, when using an error handler and handling the error immediately on the original object level, the deserialization will succeed and the item will have the default value (0 or 'Red' in this case).
```csharp
new JsonSerializerSettings() { Error = (sender, e) => { e.ErrorContext.Handled = true; }}
```
However, if we try to handle the error on a higher level object (e.g. on the MyClass object) like this:
```csharp
new JsonSerializerSettings()
{
    Error = (sender, e) => 
    {
        if (e.CurrentObject is MyClass)
        {
            e.ErrorContext.Handled = true;
        }
    }
}
```
then the expected behavior is that the 'Values' list will be null, or at least that deserialization succeeds with some result.

### Actual behavior

The actual behavior is that deserialization fails with exception, because the reader is not skipped properly after handling the error (in JsonSerializerInternalReader.HandleError). 
The exception is:
Unexpected token when deserializing object: String. Path 'Values[1]'

### Steps to reproduce

```csharp
var myClass = JsonConvert.DeserializeObject<MyClass>("{\"Values\":[\"Black\", \"Red\"]}", 
new JsonSerializerSettings()
{
    Error = (sender, e) => 
    {
        if (e.CurrentObject is MyClass)
        {
            e.ErrorContext.Handled = true;
        }
    }
});
```

gilad61:
This issue has a proposed fix with pull request #1550 
The example above was actually fixed with commit f88ae288bc00836760ee8dfab526e4b944a191bb, however that commit only handled the Object case, while the PopulateDictionary still suffers from the same issue.
The added unit test in pull request #1550 demonstrates the failed case.

PR #1550, opened 2017-12-22
Update initialDepth for error handling

gilad61:
When reading past the error in HandleError() method, the reader should keep reading until the depth is equal to the initial depth, rather then initialDepth + 1.
Only in case of a list or array (or multidimensional array), the initial depth should be advanced to +1, and that is because we don't want to skip list items in the same depth.

This change is also a fix to issue #1499 
A fix to the same issue was added earlier with commit f88ae288bc00836760ee8dfab526e4b944a191bb, however, that was a partial fix, as it only fixes the Object case but not the Dictionary case.
A unit test is added here to demonstrate the Dictionary error handling case, which was not covered by the previous fix.


JamesNK:
Looks good to me! Thanks for the fix + a unit test

Issue #1284, opened 2017-04-15
Deserialized cyclic references are null


Alois-xx:
I was trying to test the reference handling of JSON.NET and either it is a feature or I have not yet found the right settings for it. 
I would expect that when I set 
       PreserveReferencesHandling = PreserveReferencesHandling.Objects,
       ReferenceLoopHandling = ReferenceLoopHandling.Serialize,

that cyclic references are serialized and deserialized. Serialization works without failure but the the deserialized cyclic field is nulled out. Am I missing something? I am using Nuget package 10.0.2. 

`        [DataContract]
        public class DataForDataContract
        {
            [DataMember]
            public int Id { get; set; }

            [DataMember]
            public DataForDataContract Parent;

            [DataMember]
            private string Message;

            public DataForDataContract(int id, string message)
            {
                Id = id;
                Message = message;
            }
        }

        static void JsonNET()
        {
            var ser = JsonSerializer.Create(new JsonSerializerSettings()
            {
                PreserveReferencesHandling = PreserveReferencesHandling.Objects,
                ReferenceLoopHandling = ReferenceLoopHandling.Serialize,
                
            });
            var memory = Memory;
            var parent = new DataForDataContract(1, "Hello");
            parent.Parent = parent;
            var textWriter = new StreamWriter(memory);
            ser.Serialize(textWriter, parent);
            textWriter.Flush();
            memory.Position = 0;
            using (var textReader = new StreamReader(memory))
            {
                var copy = (DataForDataContract)ser.Deserialize(textReader, typeof(DataForDataContract));
                if( copy.Parent == null )
                {
                    throw new NullReferenceException("copy.Parent is null");
                }
            }
        }`

mohannad-abwah:
I've come across the same problem. Is this a known issue?

It seems to only happen at the top level, since this [dotnetfiddle](https://dotnetfiddle.net/j1Qhu6) seems to have the cyclical references working fine after the first level. By that I mean the root object's parent field is null, but the childrens' parents fields have proper references.

P.S. I've simplified the demo code a bit.
```
void Main()
{
	var ser = new JsonSerializerSettings()
	{
		PreserveReferencesHandling = PreserveReferencesHandling.Objects
	};
	var parent = new DataForDataContract(1, "Hello");
	parent.Parent = parent;
	string json = JsonConvert.SerializeObject(parent, ser); // works great!
	var deserializedObject = JsonConvert.DeserializeObject<DataForDataContract>(json);
[...]

PR #1567, opened 2018-01-05
Fixe for issue #1284

AndreasTruetschel:
# Issue
The problem when deserializing a cyclic object graph and not using default constructors is, that the object is put in the reference resolver after the child-objects are deserialized. They try to load the parent reference from the reference resolver that is not inserted yet.
# Proposed Solution
The proposed and implemented solution is to deffer the deserialization of child objects till the parent object is constructed and put to the reference resolver. For this to work, a new JsonReader type is implemented that queues the tokens of the child objects. When the parent object is registered in the reference resolver, the desesialization of the queued tokens is started.
# Drawbacks
When deserializing large json files, many tokens have to be stored, that is both, time and memory intensive to copy.
If the constructor of the parent type has a property for the referenced child type, the reference cannot be resolved, because the constructor is invoked before the child object is available.

```
public class Parent
{
    public int Id { get; set; }
    private string Message;
    private readonly Child _child;

    public Parent(Child child, int id, string message)
    {
        Id = id;
        Message = message;

        // child will be null
    }
}

public class Child
{
    public double D { get; set; }
    public Parent Parent { get; }

    public Child(Parent parent, double d)
    {
        Parent = parent; // This contains the parent just fine
        D = d;
    }
}
```

In the example above, the child parameter of the parents constructor will be null, while the child contructor receives the correct parent object.

If the child object is not injected into the parents constructor but a field or property, everything works as expected.

# Other possible solutions
Another possible solution is [instantiating the object without calling a constructor](https://msdn.microsoft.com/de-de/library/system.runtime.serialization.formatterservices.getuninitializedobject(v=vs.110).aspx), registering it in the reference-resolver and then [call the constructor](https://stackoverflow.com/questions/27380307/calling-constructor-in-c-sharp-for-a-second-time#answer-27380691).

This is the best option in my opinion but is impractical to implement here. It is a braking change. The ObjectConstructor<T> type had to be modified.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Issues/Issue1284.cs
+++ b/Src/Newtonsoft.Json.Tests/Issues/Issue1284.cs
@@ -1,0 +1,137 @@
+﻿#region License
+// Copyright (c) 2007 James Newton-King
+//
+// Permission is hereby granted, free of charge, to any person
+// obtaining a copy of this software and associated documentation
+// files (the "Software"), to deal in the Software without
+// restriction, including without limitation the rights to use,
+// copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following
+// conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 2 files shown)


waldnercharles:
Until this is merged, is there a workaround?

PR #1577, opened 2018-01-13
 Add JsonAttributeForXMLAttribute

qq405165798:
JsonAttributeForXMLAttribute is added   to flag the property is attribute in xml


JamesNK:
I don't understand what this is achieving. If it is just metadata then I believe you could implement this via JsonProperty.AttributeProvider

PR #1581, opened 2018-01-15
Fix support list value converters when input is not an array or an object, fixes #1576

### LINKED ISSUE #1576: JsonConverter is not called when target type is list and token is string
**OS**: Windows 7 64bit
**Target framework**: .NET 4.5.2
**Version**: 10.0.3 and reproduced in 9.0.1 too

I have the following test type.

```
class Simple
{
    public int A { get; set; } = 3;
    public bool B { get; set; } = true;
    public string C { get; set; } = "test string";
}

class Nested
{
    public List<int> ChildList { get; set; } = new List<int>();
    public List<Simple> ChildObjectList { get; set; } = new List<Simple>();
}
```

I've set up a `JsonConverter` for handling inputs where a single element is sent without an `array` wrapper.

```
    public class CustomContractResolver : DefaultContractResolver
    {
        protected override JsonProperty CreateProperty(MemberInfo member, MemberSerialization memberSerialization)
        {
            var property = base.CreateProperty(member, memberSerialization);

            if (member.Name == "ChildObjectList" || member.Name == "ChildList")
            {
                property.Converter = new OneItemListJsonConverter();
                property.MemberConverter = new OneItemListJsonConverter();
            }

            return property;
        }
    }
    
    public class OneItemListJsonConverter : JsonConverter
    {
        public override bool CanWrite => false;

        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
        {
            throw new NotSupportedException();
        }

        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
        {
            var token = JToken.Load(reader);
            if (token.Type == JTokenType.Array)
            {
                return token.ToObject(objectType, serializer);
            }

            var array = new JArray();
            array.Add(token);

            return array.ToObject(objectType, serializer);
        }

        public override bool CanConvert(Type objectType)
        {
            return false;
        }
    }
```

When the list has primitive types the converter won't ever get called.

```
var workingInput = "{ 'ChildObjectList': { 'A': '1', 'B': 'true', 'C': 'test string' }}";
var failingInput = "{ 'ChildList': '2' }";

var settings = new JsonSerializerSettings { ContractResolver = new CustomContractResolver() };

var good = JsonConvert.DeserializeObject<Nested>(workingInput);
good.ChildObjectList.Count.Should().Be(1); // passes

var bad = JsonConvert.DeserializeObject<Nested>(failingInput);
bad.ChildList.Count.Should().Be(1); // FAILES without exception, list is empty
```

Thank you in advance for your help.

cervengoc:
When deserializing into a list with a custom value converter the converter was not called if

- The input was not an 'object' or an 'array' token and
- The target property was not settable

This PR fixes this support by allowing using the existing value logic when there is a custom converter for the given property so the value converter will get called and the list will get deserialized properly.

Fixes #1576

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Issues/Issue1576.cs
+++ b/Src/Newtonsoft.Json.Tests/Issues/Issue1576.cs
@@ -58,21 +58,6 @@ namespace Newtonsoft.Json.Tests.Issues
             Assert.AreEqual(result.Items[0], 11);
         }
 
-        [Test]
-        public void Test2()
-        {
-            var settings = new JsonSerializerSettings()
-            {
-                Converters = new[] { new OneItemListJsonConverter() }
-            };
-
-            var result = JsonConvert.DeserializeObject<List<int>>("'11'", settings);
-
-            Assert.IsNotNull(result);
-            Assert.AreEqual(result.Count, 1);
-            Assert.AreEqual(result[0], 11);
-        }
-
         public class TestClass
         {
             public List<int> Items { get; } = new List<int>();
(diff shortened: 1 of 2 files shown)


cervengoc:
Hi @JamesNK,

Here at Dealogic we would really need this fix for a project. Could you please give us some notes and estimates about a possible merge? Thank you in advance.

JamesNK:
Merged. I don't know when I'll next do a beta or full release. The last one was only a few weeks ago.

cervengoc:
Thank you, of course that's understandable. We'll try to find some workaround for the time being.

Issue #1582, opened 2018-01-15
Deserialize to ConcurrentDictionary with null values throws System.ArgumentException


fabicz:
Using 10.0.3 and also using 11.0.1-beta3, when parsing dictionary to ConcurrentDictionary I get 
```csharp
   at System.Collections.Concurrent.ConcurrentDictionary`2.System.Collections.IDictionary.set_Item(Object key, Object value)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateDictionary(IDictionary dictionary, JsonReader reader, JsonDictionaryContract contract, JsonProperty containerProperty, String id)
```

Standard Dictionary happily deserializes. 

```csharp
JsonConvert.DeserializeObject<Dictionary<string, string>>("{\"foo\":null}");
Count = 1
    [0]: {[foo, ]}
```

### Source/destination types

```csharp
ConcurrentDictionary<string, string>
```

### Source/destination JSON

```javascript
"{\"foo\":null}"
```

### Expected behavior

When deserializing I get ConcurrentDictionary with foo => null

### Actual behavior

I get `System.ArgumentException`

### Steps to reproduce

```csharp
JsonConvert.DeserializeObject<System.Collections.Concurrent.ConcurrentDictionary<string, string>>("{\"foo\":null}");
```

testfirstcoder:
@fabicz have created a pull request [1584](https://github.com/JamesNK/Newtonsoft.Json/pull/1584)

JamesNK:
I used an alternate approach. Fixed with https://github.com/JamesNK/Newtonsoft.Json/commit/8ece2b31b90c479572c70713c8e72f3ed48684d5

timdoke:
@JamesNK this doesn't work, by the way, if you have a case insensitive concurrent dictionary (as of version 12.0.3)..

PR #1585, opened 2018-01-18
Deserialization of a concurrent dictionary with null value (fix for issue #1582)

testfirstcoder:


testfirstcoder:
Issue [1582](https://github.com/JamesNK/Newtonsoft.Json/issues/1582)

JamesNK:
Hmmm, the problem I have with this is now every dictionary will call contains before setting the value which hurts performance. Also some user implementations of dictionary might not have implemented Contains or Add, and would start throwing errors.

Why does `ConcurrentDictionary` error when using the index setter with null? IMO it looks like a bug with `ConcurrentDictionary`

testfirstcoder:
@JamesNK Added an issue [26447](https://github.com/dotnet/corefx/issues/26447) to [dotnet/corefx](https://github.com/dotnet/corefx).

JamesNK:
Worked around with https://github.com/JamesNK/Newtonsoft.Json/commit/8ece2b31b90c479572c70713c8e72f3ed48684d5

Issue #1575, opened 2018-01-12
Properties which devided into two interfaces could not be serialized and deserialied


TruePluto:
I have a interface with a properties Uid which setter and getter derived from two interface individually. a instance with the interface could not be serialized (deserialized of course ) 
A JsonSerializationException was getted and Message=A member with the name 'Uid' already exists on 'ConsoleApp3.IGuid'. Use the JsonPropertyAttribute to specify another name.
  
the following is the code:

`
using System;
using System.Collections.Generic;
using System.IO;

using Newtonsoft.Json;

namespace ConsoleApp3
{
    class Program
    {
        static void Main(string[] args)
        {
            Model model = new Model();

            DataRef refDate = new DataRef();
            IGuid iRef = refDate;

            model.GuidRefList.Add(iRef);

            using (FileStream s = File.Open("D:\\Test.Json", FileMode.Create))
            {
                s.Serialize(model);
            }
        }
    }

    public class Model
    {
        public List<IGuid> GuidRefList { get; } =
            new List<IGuid>();
    }

    public interface IGuidSet
    {
        Guid Uid { set; }
    }
    public interface IGuidGet
    {
        Guid Uid { get; }
    }

    public interface IGuid :
        IGuidGet,
        IGuidSet
    { }

    public class DataRef :
        IGuid
    {
        public Guid Uid { get; set; }
    }

    public static class JsonHelpers
    {
        public static void Serialize(this Stream stream, object value)
        {
            using (StreamWriter writer = new StreamWriter(stream))
            using (JsonTextWriter jsonWriter = new JsonTextWriter(writer))
            {
                var serializer = JsonSerializer.Create(new JsonSerializerSettings
                {
                    NullValueHandling = NullValueHandling.Ignore,
                    TypeNameHandling = TypeNameHandling.Auto,
                    TypeNameAssemblyFormatHandling = TypeNameAssemblyFormatHandling.Simple
                });
                serializer.Formatting = Formatting.Indented;
                serializer.Serialize(jsonWriter, value);
                jsonWriter.Flush();
            }
        }

        public static T Deserialize<T>(this Stream stream)
        {
            using (StreamReader reader = new StreamReader(stream))
            using (JsonTextReader jsonReader = new JsonTextReader(reader))
            {
                var serializer = JsonSerializer.Create(new JsonSerializerSettings
                {
                    NullValueHandling = NullValueHandling.Ignore,
[...]

PR #1586, opened 2018-01-18
serialization of a property which was defined in several interfaces

testfirstcoder:
Issue [1575](https://github.com/JamesNK/Newtonsoft.Json/issues/1575)


JamesNK:
Thanks for the fix with tests! Merged.

PR #1612, opened 2018-02-10
Support for regex match operator for jsonpath queries

andre-n:
Added implementation for feature request #858. It is based on previous request #1382.
Implementation is using .NET regex syntax, and it is compatible with Jayway implementation.


JamesNK:
Thanks. I made some fixes to handle escaped chars in regexes but overall looks good!

https://github.com/JamesNK/Newtonsoft.Json/commit/6d3b5077e2379065bcd7de246139961e502fa92c

Issue #1613, opened 2018-02-15
ReadStringIntoBuffer optimization


IlyaGrebnov:
While profiling my application I found that Newtonsoft.ReadStringIntoBuffer main loop is costly and potentially could be optimized. Currently main loop looks like this

```
while (true)
{
    switch (_chars[charPos++])
    {
	case '\0':
	case '\\':
	case '\r':
	case '\n':
        case '"':
        case '\'':
    }
}
```

which is sub-optimal for mainline alphanumeric characters. I was thinking this could be optimized like

```
while (true)
{
    char currentChar = _chars[charPos++];
    if (currentChar > quote && currentChar != '\\')
    {
        continue;
    }

    switch (currentChar)
    {
	case '\0':
	case '\\':
	case '\r':
	case '\n':
        case '"':
        case '\'':
    }
}
```

Note, I have not tried this due to luck of time and VS2017 (read-only structs give me errors).

IlyaGrebnov:
I was able to test idea.

- DeserializeLargeJsonText improved from 108.124 ms to 107.841 ms
- DeserializeLargeBase64JsonText with large base64 encoded string proprieties improved from 8.559 to 7.923 ms.

So it doesn't looks like it is worth it today, but it is possible to optimize it in future with conditional moves. Once https://github.com/dotnet/coreclr/issues/7447 is fixed.

JamesNK:
Such a small difference is within the margin of error. I'd prefer to leave it the same until changing creates a more significant difference.

PR #1615, opened 2018-02-16
Proposed change for #1613

IlyaGrebnov:
As I mention improvement is not big, so feel free to reject this PR.


JamesNK:
Hi

It is a small improvement but I'd prefer to leave it for now. Next time I focus on performance then I'll investigate what you've done.

PR #1634, opened 2018-03-05
Ensure that RepositoryUrl follows documentation

LordMike:
In the documentation for the [MSBuild Pack Target](https://github.com/emgarten/docs.microsoft.com-nuget/blob/patch-1/docs/reference/msbuild-targets.md#pack-target) it should be `Repository URL used to clone or retrieve source code` and that the example is `https://github.com/NuGet/NuGet.Client.git`.

This change makes the Nuget packaging follow that example.


JamesNK:
Looks good! Thanks.

PR #1668, opened 2018-04-03
Fixes deserializing decimal.MaxValue + 1 not overflowing

### LINKED ISSUE #1667: De-serialization of Decimal not always overflowing when it should
### Source/destination types

```javascript
79228162514264337593543950336
```
### Source/destination type

```csharp
79228162514264337593543950330
```

### Expected behavior

I expected a overflow exception to be thrown since it is over the limit of the Decimal type

### Actual behavior

When the first 28 digits are "7922816251426433759354395033" it always loses the last digit and gets rounded down without throwing an exception.

### Steps to reproduce

Trying to parse to Decimal any number between 79228162514264337593543950336 and 79228162514264337593543950339
Same problem happens with negative numbers between -79228162514264337593543950336
and -79228162514264337593543950339

```csharp
Newtonsoft.Json.JsonConvert.DeserializeObject<Decimal>("79228162514264337593543950336")
```

### Cause

I did some digging and it looks related to [this pull](https://github.com/JamesNK/Newtonsoft.Json/pull/1204)

In ConversionUtils.DecimalTryParse when exponent =1 and mantissaDigits == 29 it always multiplies the value *10 losing the last digit.

TylerBrinkley:
Fixes #1667

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Utilities/ConvertUtilsTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Utilities/ConvertUtilsTests.cs
@@ -233,6 +233,8 @@ namespace Newtonsoft.Json.Tests.Utilities
 
             AssertDecimalTryParse("1E+29", ParseResult.Overflow, null);
             AssertDecimalTryParse("-1E+29", ParseResult.Overflow, null);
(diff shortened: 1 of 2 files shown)


JamesNK:
Looks good. Thanks for the fix.

PR #1690, opened 2018-05-01
specify that it supports net standard 2.0

CoolGoose:


JamesNK:
Look s good to me. Thanks.

CoolGoose:
@JamesNK any way I can help a bit more with the homepage / docs ? A preferred flow or any kind of to do list ?

Issue #1682, opened 2018-04-18
[.NET Core 2.0] Cannot serialize class with property which has abstract base class and it's marked as Serializable


Enumix:
**Package version on which I've tested that it throws: Newtonsoft.Json 10.0.1, 10.0.2, 10.0.3, 11.0.1, 11.0.2** 
**It works on: 9.0.1**
**Runtime: .NET Core 2.0**

Serialization of class with serializable property throws an exception. That serializable property needs to inherit from base abstract class that is marked as [Serializable]. I put extra comments in code. I confirmed that it works on .NET Framework 4.6.1 with Newtonsoft.Json with same version.

### Source/destination types

```csharp
    public class ClassWithSerializableProperty
    {
        public BaseSerializable Serializable { get; }
    }

    [Serializable]
    public class ConcreteSerializable : BaseSerializable
    {
        public ConcreteSerializable()
        {
        }

        protected ConcreteSerializable(SerializationInfo info, StreamingContext context)
            : base(info, context)
        {
        }
    }

    [Serializable] // it won't blow up after removing that attribute
    public abstract class BaseSerializable : ISerializable //or that interface, or when we will remove "abstract" keyword
    {
        public BaseSerializable()
        {
        }

        //it won't fail when that constructor is missing
        protected BaseSerializable(SerializationInfo info, StreamingContext context)
        {
        }

        public void GetObjectData(SerializationInfo info, StreamingContext context)
        {
        }
    }
```

### Source/destination JSON

It blows up, so I cannot check if I will be able to deserialize it later.

### Expected behavior

It should serialize object correctly (without blowing up).

### Actual behavior
```csharp
Test method TestCoreProject.TrickyBugs.FailOnCore threw exception: 
System.InvalidOperationException: Can't compile a NewExpression with a constructor declared on an abstract class
    at System.Linq.Expressions.Compiler.LambdaCompiler.EmitNewExpression(Expression expr)
   at System.Linq.Expressions.Compiler.LambdaCompiler.EmitExpression(Expression node, CompilationFlags flags)
   at System.Linq.Expressions.Compiler.LambdaCompiler.EmitLambdaBody(CompilerScope parent, Boolean inlined, CompilationFlags flags)
   at System.Linq.Expressions.Compiler.LambdaCompiler.EmitLambdaBody()
   at System.Linq.Expressions.Compiler.LambdaCompiler.Compile(LambdaExpression lambda)
[...]

PR #1712, opened 2018-05-20
Fixed serializing abstract base class ISerializable on .NET Core

JamesNK:
Fixes https://github.com/JamesNK/Newtonsoft.Json/issues/1682


sorgpro:
Dont fixed. Create empty test .net core 2.1 project.
Insert into created UnitTest1.cs:
    [TestFixture]
    public class Issue1682 : TestFixtureBase...
Add File TestFixtureBase to project.
Add nuget Newtonsoft.Json (11.0.2)

Test_Serialize result:
Message: System.InvalidOperationException : Can't compile a NewExpression with a constructor declared on an abstract class.

PR #1718, opened 2018-05-21
‘Annoucing’ → ‘Announcing’

jnm2:
Just spotted this in https://www.newtonsoft.com/json/help/html/QueryingLINQtoJSON.htm and replaced all occurrences.


JamesNK:
Good catch. Thanks

PR #1746, opened 2018-06-22
[WIP] Enable source link

JamesNK:


JamesNK:
> C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\Roslyn\Microsoft.Managed.Core.targets(90,5): error : SourceRoot items must include at least one top-level (not nested) item when DeterministicSourcePaths is true [C:\projects\newtonsoft-json\Src\Newtonsoft.Json\Newtonsoft.Json.csproj]

@ctaggart What is the cause of this error and how do I fix it?

ctaggart:
I don't know. What is your environment? I know it requires .NET SDK 2.1 is required. Can you give me a `msbuild /version` and a `dotnet --info` for that environment?

ctaggart:
Any just to make sure your git repo is normal, can you run `git rev-parse HEAD` & `git config --get remote.origin.url`?

JamesNK:
MSBuild is 15.7.179.6572

https://ci.appveyor.com/project/JamesNewtonKing/newtonsoft-json/build/1.0.1535#L231

I'm not actually building with dotnet cli.

JamesNK:
```
C:\Development\Source\Newtonsoft.Json [jamesnk/source-link ≡]> git rev-parse HEAD
33d60e09415a79ba9d397c436938f466858deccc
C:\Development\Source\Newtonsoft.Json [jamesnk/source-link ≡]> git config --get remote.origin.url
https://github.com/JamesNK/Newtonsoft.Json.git
C:\Development\Source\Newtonsoft.Json [jamesnk/source-link ≡]>
```

Normal looking?

JamesNK:
I updated dotnet CLI to latest just in case and still have the same error.

```
C:\Development\Source\Newtonsoft.Json [jamesnk/source-link ≡]> dotnet --version
2.1.301
```

ctaggart:
I think you are hitting this bug: https://github.com/dotnet/sourcelink/issues/62

May be try setting `DeterministicSourcePaths` to `false`.

ctaggart:
I opened up #1747 for that change to this branch. My build is still going, but I'm going to sleep:
https://ci.appveyor.com/project/ctaggart/newtonsoft-json/build/28

TylerBrinkley:
When adding SourceLink support to my library I just used the generated NuGet Package in a local folder NuGet Package Source and tested it with a new console app. I also temporarily renamed my repo folder so it wouldn't find my local files.

ctaggart:
You can use `dotnet sourcelink test`:
https://github.com/ctaggart/SourceLink#test

Unfortunately, it is not working yet. Both of these fail:
```
dotnet sourcelink test "C:\Users\taggac\Downloads\Newtonsoft.Json.11.0.3-beta1.nupkg"
dotnet sourcelink print-json "C:\Users\taggac\Downloads\Newtonsoft.Json.11.0.3-beta1\lib\netstandard2.0\Newtonsoft.
Json.pdb"
```

Src/Newtonsoft.Json/Newtonsoft.Json.csproj
```diff
@@ -21,17 +21,21 @@
...
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <MinClientVersion>2.12</MinClientVersion>
+    <!-- including PDB files in NuGet for source link because symbolsource.org does not support portable PDBs -->
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder> 
   </PropertyGroup>
   <ItemGroup>
     <None Remove="**\*.orig" />
     <None Include="..\..\LICENSE.md" Pack="true" PackagePath="LICENSE.md" />
   </ItemGroup>
+  <ItemGroup Condition="'$(ContinuousIntegrationBuild)'=='true'">
```
ctaggart: `Microsoft.SourceLink.GitHub` is not being installed. I removing this condition in my pull request and will try it out.

ctaggart:
Rather than adding `/p:DeterministicSourcePaths=false` to everything like #1747, It is probably better to follow https://github.com/dotnet/sourcelink/issues/91 by adding this to the WPF projects:
``` xml
<DeterministicSourcePaths Condition="'$(EnableSourceLink)' == ''">false</DeterministicSourcePaths>
```

ctaggart:
The `dotnet sourlink test`s pass with #1747 changes.

ctaggart:
You can now install `sourcelink` as a global .NET tool
https://www.nuget.org/packages/SourceLink/3.0.0-build.732
https://natemcmaster.com/blog/2018/05/12/dotnet-global-tools/

JamesNK:
> It is probably better to follow dotnet/sourcelink#91 by adding this to the WPF projects:

There are no WPF projects.

What exactly is `ContinuousIntegrationBuild` and `DeterministicSourcePaths`? Why doesn't `DeterministicSourcePaths` work and what does setting it to false mean?

I don't like setting values without knowing what the cause and effect is.

JamesNK:
Merged. Thanks for your help @ctaggart

MagicAndre1981:
do you have a preview/nightly NuGet Package to test this? On myget I also only see the [11.0.2](https://www.myget.org/feed/nugety-core/package/nuget/Newtonsoft.Json/11.0.2).

JamesNK:
https://www.myget.org/feed/json-net/package/nuget/Newtonsoft.Json

MagicAndre1981:
thanks. Works, I see the SourceLink entry in the PDB (with dotPeek from Jetbrains) and stepping through the code works fine in VS2017 15.7.4

tmat:
> What exactly is ContinuousIntegrationBuild and DeterministicSourcePaths? Why doesn't DeterministicSourcePaths work and what does setting it to false mean?

`ContinuousIntegrationBuild` is an indicator that the build is a CI build as opposed to a dev build. Targets may decide to behave differently when running on CI vs. local dev build. For example, when both `Deterministic` is `true` and `ContinuousIntegrationBuild` is `true` we enable `DeterministicSourcePaths`. 

`DeterministicSourcePaths` causes the build to normalize all paths generated to the DLL and PDB to be deterministic - that is do not depend on the directory the source code was built from and the OS it was built on. Without this feature on the PDBs would include full paths like `C:\MyRepos\Newtonsoft.Json\Src\Newtonsoft.Json\ConstructorHandling.cs`. With `DeterministicSourcePaths` on the paths will be normalized to `/_/Src/Newtonsoft.Json/ConstructorHandling.cs`. You can see this happening in the SourceLink entry - since SourceLink needs to map the directory of the repository to the raw github URL it maps `/_/*`.

The reason why `DeterministicSourcePaths` is only switched on for CI builds is local debugging. If it was enabled when you e.g. F5 in Visual Studio the debugger wouldn't be able to find the sources as they do not exist at `/_/*` location on your dev machine and they were not pushed to the github repo yet, assuming you don't push every time you debug.

ctaggart:
Looking forward to 11.0.3 being published :-)

ctaggart:
I blogged about it:
http://blog.ctaggart.com/2018/06/newtonsoftjson-enabling-source-link.html

dezsiszabi:
This PR had a possibly unintentional side effect:

https://stackoverflow.com/a/53983050/4620101

Maybe it's worth mentioning it in documentation or somewhere (migration doc?).

https://docs.microsoft.com/en-us/visualstudio/test/troubleshooting-code-coverage?view=vs-2017#pdb-symbol-files-are-unavailable

tmat:
Since nuget.org server now supports uploading symbol packages, I'd recommend to remove the PDB from the package (remove `AllowedOutputExtensionsInPackageBuildOutputFolder` property update) and instead produce .snupkg by setting:

```xml
<IncludeSymbols>true</IncludeSymbols>
<SymbolPackageFormat>snupkg</SymbolPackageFormat>
```

.snupkg file can then be published to nuget.org along with the main package.

Issue #1753, opened 2018-06-29
IndexOutOfRangeException in PropertyNameTable.Get()


dco123:
I'm seeing an intermittent issue where an IndexOutOfRangeException is thrown from PropertyNameTable.Get(), using 11.0.2. I have many processes loading and deserializing the same set of files, and within a process the same JsonSerializer is deserializing these files in parallel. Most of them are successful but occasionally one or two of them fail with this error.

I suspect it might be a race condition in PropertyNameTable but not sure about that. I reproduced the issue by repeatedly running an app that deserialises files in parallel using the same serializer (though it can take thousands of iterations before I see it), but I can't provide the json / code as it's sensitive. Hopefully the stack below might give some clues.

```
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Newtonsoft.Json.Utilities.PropertyNameTable.Get(Char[] key, Int32 start, Int32 length)
   at Newtonsoft.Json.JsonTextReader.ParseProperty()
   at Newtonsoft.Json.JsonTextReader.ParseObject()
   at Newtonsoft.Json.JsonTextReader.Read()
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.ResolvePropertyAndCreatorValues(JsonObjectContract contract, JsonProperty containerProperty, JsonReader reader, Type objectType)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObjectUsingCreatorWithParameters(JsonReader reader, JsonObjectContract contract, JsonProperty containerProperty, ObjectConstructor`1 creator, String id)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateNewObject(JsonReader reader, JsonObjectContract objectContract, JsonProperty containerMember, JsonProperty containerProperty, String id, Boolean& createdFromNonDefaultCreator)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
[...]

PR #1756, opened 2018-07-02
Avoid race condition in DefaultJsonNameTable.Get by ensuring array index is evaluated before accessing the array

dco123:
From https://github.com/JamesNK/Newtonsoft.Json/issues/1753

Couldn't see a way to unit test this so put a comment in to make sure it doesn't get reverted


IlyaGrebnov:
What if we mark _entries with volatile keyword?

IlyaGrebnov:
I think even better approach is remove _mask and just compute it:

var entries= this._entries;
var mask = entries.length - 1;

use entries and mask.

JamesNK:
I added your change with https://github.com/JamesNK/Newtonsoft.Json/pull/1781

I made a small modification to also copy the entries to a local variable.

PR #1770, opened 2018-07-18
Update XmlNodeConverter.cs to enable reading Type attribute from XML …

oigbokwe73:
…nodes

This modification provides the class a way to read an optional attribute from XML nodes called "Type" that holds the desired serialization of a node value. By default, the converter serializes all values as string in the result JSON string, but now you can add an attribute that specifies the desired DataType output. The allowed types are Integer, Float, Boolean and Date.

Example:

    <Object>
      <ID json:Type='Integer'>12</ID>
      <Title>mytitle</Title>
      <Visible json:Type='Boolean'>false</Visible>
      <Price json:Type='Float'>1.55</Price>
      <ExpirationDate json:Type='Date'>2013-12-31</ExpirationDate>
    </Object>

More info:

http://stackoverflow.com/questions/18611139/force-jsonconvert-serializexmlnode-to-serialize-node-value-as-an-integer-or-a-bo

### DIFF
--- a/Src/Newtonsoft.Json/Converters/XmlNodeConverter.cs
+++ b/Src/Newtonsoft.Json/Converters/XmlNodeConverter.cs
@@ -887,6 +887,13 @@ namespace Newtonsoft.Json.Converters
       
       return (jsonArrayAttribute != null && XmlConvert.ToBoolean(jsonArrayAttribute.Value));
     }
+    
+    private string GetDataType(IXmlNode node) {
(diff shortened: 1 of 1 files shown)


Yuttana12wr:
#c

snicol21:
@JamesNK Would love to get this functionality!! How close is this to being merged in?

Falco20019:
@JamesNK Would also need this currently.

PR #1784, opened 2018-08-01
Make json path filter and expressions public

chris579:
**When merged this pull request will**
- Make all Linq Json Path Filter and Expressions public

Internally we are using a self written json path validation logic for custom expressions and filtering. To rely on tested parts we are using the filters of Newtonsoft.Json however they are internally by default. Our current workaround is to make a custom build where all those classes are made public. However this is difficult in workflow and often conflicts with frameworks referencing the official Newtonsoft.Json NuGet package.

We think this filters and expressions could also be helpful to others so we would like to make them public.
Plus, this opens the possibility to write custom filters and expressions as the base class is made public too.

### DIFF
--- a/Src/Newtonsoft.Json/Linq/JsonPath/ArrayIndexFilter.cs
+++ b/Src/Newtonsoft.Json/Linq/JsonPath/ArrayIndexFilter.cs
@@ -4,10 +4,22 @@ using Newtonsoft.Json.Utilities;
 
 namespace Newtonsoft.Json.Linq.JsonPath
 {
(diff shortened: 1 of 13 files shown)


JamesNK:
I would prefer to leave them internal. If they're made public then I can't make large changes to this area to improve performance without breaking anyone who uses the types.

I suggest you copy and paste just the JSON Path filters into a new library and make them public.

chris579:
Thanks for the reply, in this case we will proceed to extract them into an own library.

PR #1789, opened 2018-08-08
Doesn't write type details when typeName is empty

### LINKED ISSUE #302: Allow certain types to have no type information produced
When `SerializationBinder` produces `null` for typename, even when type handling is set to Auto/Object/Array, no `$type` information should be produced. Right now an empty string gets output instead.

TylerBrinkley:
Closes #302.


TylerBrinkley:
Closing due to age.

PR #1813, opened 2018-08-31
Fix summary of JObject.Property method

stoyandimov:
Change "Gets a JProperty the specified name."
to "Gets a JProperty with the specified name."


JamesNK:
Looks good to me. Thanks for the PR!

PR #1820, opened 2018-09-11
Add new Date constructor format

wybb:


JamesNK:
Done. I rewrote it to support multiple arguments and be supported from JavaScriptDateConverter

https://github.com/JamesNK/Newtonsoft.Json/pull/1853

Issue #1835, opened 2018-09-17
Identity operator not supported


WillSullivan:
While, as best as I can tell (and that's only by testing it in the online evaluator at jsonpath.com), it should be possible to use the identity operator (===) in a JPath query. This doesn't appear to be supported.

Using their example, 

```javascript
{
  "firstName": "John",
  "lastName" : "doe",
  "age"      : 26,
  "address"  : {
    "streetAddress": "naist street",
    "city"         : "Nara",
    "postalCode"   : "630-0192"
  },
  "phoneNumbers": [
    {
      "type"  : "iPhone",
      "number": "0123-4567-8888"
    },
    {
      "type"  : "home",
      "number": "0123-4567-8910"
    }
  ]
}
```

The JPath

```javascript
$.phoneNumbers[?(@.type==='home')].number
```
returns the expected result

```javascript
[
  "0123-4567-8910"
]
```
However, when executing the following

```csharp
    var sampled = JObject.Parse(sample);
    var results2 = sampled.SelectTokens("$.phoneNumbers[?(@.type==='home')].number");```
```
throws a JsonException

```
Newtonsoft.Json.JsonException: Unexpected character while parsing path query: =
   at Newtonsoft.Json.Linq.JsonPath.JPath.ParseSide()
   at Newtonsoft.Json.Linq.JsonPath.JPath.ParseExpression()
   at Newtonsoft.Json.Linq.JsonPath.JPath.ParseQuery(Char indexerCloseChar, Boolean scan)
   at Newtonsoft.Json.Linq.JsonPath.JPath.ParseIndexer(Char indexerOpenChar, Boolean scan)
   at Newtonsoft.Json.Linq.JsonPath.JPath.ParsePath(List`1 filters, Int32 currentPartStartIndex, Boolean query)
   at Newtonsoft.Json.Linq.JsonPath.JPath.ParseMain()
   at Newtonsoft.Json.Linq.JsonPath.JPath..ctor(String expression)
   at Newtonsoft.Json.Linq.JToken.SelectTokens(String path)
   at GetDemCredsYo.Program.Main(String[] args)
```
Using `==` works as expected, but it can result in unforced errors when those familiar with javascript and the JPath syntax stumble into this.

WillSullivan:
There appears to be other issues with equality.  Testing on http://jsonpath.com/, given the following json


```javascript
{
	"Values": [{
		"Coercible": 1,
		"Name": "Number"
	}, {
[...]

PR #1837, opened 2018-09-18
Identity operator support added

WillSullivan:
Added support for the identity and negated identity operators ===, !==. Behavior of the JPath parser matches the behavior exhibited by these ops on jsonpath.com.  Added tests for the new behavior and includes a couple sanity tests that show current behavior doesn't match expected JSONPath results.

Related to issue #1835

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Documentation/LinqToJsonTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Documentation/LinqToJsonTests.cs
@@ -518,6 +518,40 @@ namespace Newtonsoft.Json.Tests.Documentation
             Assert.AreEqual(50m, productPrice);
             Assert.AreEqual("Elbow Grease", productName);
         }
+        [Test]
+        public void SelectTokenSupportsIdentityOperator()
+        {
+            #region SelectTokenSupportsIdentityOperator
+            JObject o = JObject.Parse(@"{
+	            'Values': [{
+
+                    'Coercible': 1,
+                    'Name': 'Number'
+
+                }, {
+		            'Coercible': '1',
+		            'Name': 'String'
+	            }]
+            }");
+
+            // just to verify expected behavior hasn't changed
+            IEnumerable<string> sanity1 = o.SelectTokens("Values[?(@.Coercible == '1')].Name").Select(x => (string)x);
+            IEnumerable<string> sanity2 = o.SelectTokens("Values[?(@.Coercible != '1')].Name").Select(x => (string)x);
+            // new behavior
+            IEnumerable<string> mustBeNumber1 = o.SelectTokens("Values[?(@.Coercible === 1)].Name").Select(x => (string)x);
+            IEnumerable<string> mustBeString1 = o.SelectTokens("Values[?(@.Coercible !== 1)].Name").Select(x => (string)x);
+            IEnumerable<string> mustBeString2 = o.SelectTokens("Values[?(@.Coercible === '1')].Name").Select(x => (string)x);
+            IEnumerable<string> mustBeNumber2 = o.SelectTokens("Values[?(@.Coercible !== '1')].Name").Select(x => (string)x);
+            #endregion
+            // FAILS-- JPath returns { "String" }
+            //CollectionAssert.AreEquivalent(new[] { "Number", "String" }, sanity1);
+            // FAILS-- JPath returns { "Number" }
+            //Assert.IsTrue(!sanity2.Any());
+            Assert.AreEqual("Number", mustBeNumber1.Single());
+            Assert.AreEqual("String", mustBeString1.Single());
+            Assert.AreEqual("Number", mustBeNumber2.Single());
+            Assert.AreEqual("String", mustBeString2.Single());
+        }
 
         [Test]
         public void SelectTokenLinq()
--- a/Src/Newtonsoft.Json/Linq/JsonPath/QueryExpression.cs
+++ b/Src/Newtonsoft.Json/Linq/JsonPath/QueryExpression.cs
@@ -24,7 +24,9 @@ namespace Newtonsoft.Json.Linq.JsonPath
         GreaterThanOrEquals = 7,
         And = 8,
         Or = 9,
-        RegexEquals = 10
+        RegexEquals = 10,
+        Identity = 11,
+        NotIdentity = 12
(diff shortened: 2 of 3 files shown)


Src/Newtonsoft.Json/Linq/JsonPath/QueryExpression.cs
```diff
@@ -24,7 +24,9 @@ internal enum QueryOperator
         GreaterThanOrEquals = 7,
         And = 8,
         Or = 9,
-        RegexEquals = 10
+        RegexEquals = 10,
+        Identity = 11,
```
JamesNK: StrictEquals
StrictNotEquals

Src/Newtonsoft.Json/Linq/JsonPath/QueryExpression.cs
```diff
@@ -257,5 +271,10 @@ private bool EqualsWithStringCoercion(JValue value, JValue queryValue)
 
             return string.Equals(currentValueString, queryValueString, StringComparison.Ordinal);
         }
+        private bool EqualsWithoutStringCoercion(JValue value, JValue queryValue)
```
JamesNK: nit: line above

Src/Newtonsoft.Json/Linq/JsonPath/QueryExpression.cs
```diff
@@ -257,5 +271,10 @@ private bool EqualsWithStringCoercion(JValue value, JValue queryValue)
...
+        {
+            // Looking at the impl I believe it considers type and value and may be euqivalent to javascript's identity op. I think. :/
+            return value.Equals(queryValue);
```
JamesNK: This will need to handle:
1 === 1.0: True

WillSullivan: The test suite needs work, re my previous comment. Asking around for docs on coercion rules, will update.

Src/Newtonsoft.Json.Tests/Documentation/LinqToJsonTests.cs
```diff
@@ -518,6 +518,40 @@ public void SelectTokenComplex()
...
+            }");
+
+            // just to verify expected behavior hasn't changed
+            IEnumerable<string> sanity1 = o.SelectTokens("Values[?(@.Coercible == '1')].Name").Select(x => (string)x);
+            IEnumerable<string> sanity2 = o.SelectTokens("Values[?(@.Coercible != '1')].Name").Select(x => (string)x);
+            // new behavior
+            IEnumerable<string> mustBeNumber1 = o.SelectTokens("Values[?(@.Coercible === 1)].Name").Select(x => (string)x);
+            IEnumerable<string> mustBeString1 = o.SelectTokens("Values[?(@.Coercible !== 1)].Name").Select(x => (string)x);
+            IEnumerable<string> mustBeString2 = o.SelectTokens("Values[?(@.Coercible === '1')].Name").Select(x => (string)x);
+            IEnumerable<string> mustBeNumber2 = o.SelectTokens("Values[?(@.Coercible !== '1')].Name").Select(x => (string)x);
```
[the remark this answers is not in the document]
WillSullivan: Going to create tests based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#Loose_equality_using

JamesNK:
Would you like me to handle the `1 === 1.0` case? I don't mind taking this PR to the finish.

WillSullivan:
I'm going to work on a test suite which should cover every case in the algorithm. There's more issues than 1 === 1.0. Should be reviewable by lunch tomorrow, if nothing more important comes up. I'm really hoping to have this querying feature working for my project as it's part of my oh-sh*t backup plan (we're working with a lot of unknowns on the business side rn).

WillSullivan:
(if nothing else I'll have the bones of it in and available to tinker with and adapt)

WillSullivan:
So, I've worked up a unit test based on the Strict algorithm, and have noticed some issues (or possibly issues in my understanding) that prevents the algorithm from being implemented as it is in ECMAScript.

First off, the algorithm has special handling for different types. But since JSON is itself typeless (errbody strings), that puts up a major roadblock to implementation. In my tests, ```JToken.Parse``` only assigns the valid Type to the token for ISO dates. For every other javascript type involved in the test (numbers, other valid date formats, null, undefined, Booleans), they're all strings. 

That makes sense, as we don't know what types they were originally. We could *assume* that when the value is "null" that it was originally null, but that's still an assumption. 

So it occurs to me that maybe, within the context of a JSONPath expression, maybe the strict equivalence operator shouldn't have the same meaning as it does in javascript. Maybe it *should* mean strict equivalence of text without any coercion. And maybe the abstract equality operator should mean just that--try to coerce text into its original data type before performing comparisons. Almost like flipping the two operators meaning in this context.

If this is the case, that brings up more issues. First, since the current abstract equality algorithm ```QueryExpressions.EqualsWithStringCoercion``` uses JTokenType to determine what coercion to perform, and since only ISO dates are, when parsing, assigned a type different than String, this won't work either. We will still not be able to correctly match 1 to 1.0 using the == operator. 

This brings up a further question... Is this, in fact, a different bug? Should the JSON parser be trying to identify the type of the string value? The way the ```EqualsWithStringCoercion``` method works it seems to imply this. Or should this method instead attempt to identify the type of the value by TryParsing everything (considering the note of concern for performance impacts in the code that might be ill advised).

So there's a bunch of questions right there. Unfortunately, I think because (afaict) JSONPath isn't an ISO standard with a well defined syntax and descriptions of operator algorithms, we're left with this unanswerable ambiguity. I don't know the best way forward from here.

I'll finish up the tests and sync them with the PR this morning. Feel free to tinker around with them. You might reach a conclusion I've missed.

WillSullivan:
Okay, I guess I got a bit distracted about all the implications and I completely whiffed what actually gets serialized by stringify. Specifically, numbers don't get serialized as strings, duh. So I can completely support that. I'm going to verify what actually comes out the other end when stringify gets done, update the test (and get rid of a lot of fluff) and add in support for the special cases that can be handled correctly. I've already verified we can support 1 === 1.0. BRB.

WillSullivan:
Great! Happy to help.

PR #1887, opened 2018-10-19
Duplicate properties handling

### LINKED ISSUE #931: Handling duplicate elements in JSON payload
What is the best way with Json.NET to restrict occurrence of duplicate elements in JSON payload?
For example:
 {
    "role" : "default",
    "username" : "mallory",
    "role": "admin",
    "password" : "Evil123!"
}

Rebel028:
Fixes #931
Added property for JsonLoadSettings for handling elements with the same name. 
By default duplicates are ignored and only last value used. 
To throw an exception set `DuplicatePropertyNamesHandling = DuplicatePropertyNamesHandling.Throw`.


JarLob:
@JamesNK are you against merging it?

Rebel028:
Thank you!

Issue #1798, opened 2018-08-14
SerializeObject hangs until OutOfMemoryException when serializing AutoMapperMappingException object


wuqi618:
### Source type

```csharp
AutoMapper.AutoMapperMappingException
```

### Expected behavior

No error

### Actual behavior

JsonConvert.SerializeObject method hangs until it throws OutOfMemoryException.

stack trace from a memory dump shows it was stuck at InternalAssemblyBuilder.Equals
```
00000067efb5dc88 000007fa0c6b8cbd System.Reflection.Emit.InternalAssemblyBuilder.Equals(System.Object)
00000067efb5dc90 000007fa08ea1042 System.Collections.Generic.ObjectEqualityComparer`1[[System.__Canon, mscorlib]].Equals(System.__Canon, System.__Canon)
00000067efb5dcc0 000007fa08eb46d1 System.Collections.Generic.List`1[[System.__Canon, mscorlib]].Contains(System.__Canon)
00000067efb5dd20 000007fa0b581fe6 Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.CheckForCircularReference(Newtonsoft.Json.JsonWriter, System.Object, Newtonsoft.Json.Serialization.JsonProperty, Newtonsoft.Json.Serialization.JsonContract, Newtonsoft.Json.Serialization.JsonContainerContract, Newtonsoft.Json.Serialization.JsonProperty)
00000067efb5dd80 000007fa0b581ab7 Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.CalculatePropertyValues(Newtonsoft.Json.JsonWriter, System.Object, Newtonsoft.Json.Serialization.JsonContainerContract, Newtonsoft.Json.Serialization.JsonProperty, Newtonsoft.Json.Serialization.JsonProperty, Newtonsoft.Json.Serialization.JsonContract ByRef, System.Object ByRef)
00000067efb5de10 000007fa0b5808a7 Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeObject(Newtonsoft.Json.JsonWriter, System.Object, Newtonsoft.Json.Serialization.JsonObjectContract, Newtonsoft.Json.Serialization.JsonProperty, Newtonsoft.Json.Serialization.JsonContainerContract, Newtonsoft.Json.Serialization.JsonProperty)
[...]

Issue #1896, opened 2018-10-27
Cannot serialize custom exception types under .NET Core


superstator:
When serializing custom exception types, there seems to be a regression between .NET Framework and .NET Core. Under the full framework, I can serialize both built-in and custom exception types with no unexpected behavior. Under .NET Core I see crashes and/or malformed JSON, but only when serializing custom exception types. Built-in types like `System.Exception` or `System.ArgumentException` still work fine.

### Source/destination types

A simple exception type like:
```csharp
public class TestException : Exception
{
    public TestException(string message) : base(message) { }
}
```

### Expected behavior

A simple JSON object, like:

```json
{
  "Message": "Oops",
  "Data": {},
  "InnerException": null,
  "TargetSite": {
    "Name": "CanSerializeException",
    "AssemblyName": "UnitTestProject1, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null",
    "ClassName": "UnitTestProject1.UnitTest1",
    "Signature": "Void CanSerializeException()",
    "Signature2": "System.Void CanSerializeException()",
    "MemberType": 8,
    "GenericArguments": null
  },
  "StackTrace": "   at UnitTestProject1.UnitTest1.CanSerializeException() in C:\\...\\UnitTest1.cs:line 20",
  "HelpLink": null,
  "Source": "UnitTestProject1",
  "HResult": -2146233088
}
```

Note this JSON was generated by running my unaltered test sample below under .NET Framework 4.6.1.

### Actual behavior

Under .NET Core, If I serialize the caught exception naively, I get this error:

`Self referencing loop detected for property 'Module' with type 'System.Reflection.RuntimeModule'. Path 'TargetSite.Module.Assembly.EntryPoint'.`

If I include the `ReferenceLoopHandling = ReferenceLoopHandling.Ignore` setting for serialization, serialization succeeds, but performance is very poor and the resulting JSON is absurdly long (tens of millions of characters) and full of repeated data.

Under .NET Framework, serialization behaves as expected.

### Steps to reproduce

A minimal MSTest example:

```csharp
[TestClass]
public class ExceptionTests
{
    public class TestException : Exception
    {
        public TestException(string message) : base(message) { }
    }

    [TestMethod]
    public void CanSerializeException()
    {
        try
        {
            throw new TestException("Oops");
        }
        catch (Exception ex)
        {
[...]

Issue #1854, opened 2018-09-26
Jsonconvert.seralizeobject include targetsite so exception is huge in size.


GomesNayagam:
### Source/destination types

```csharp
HttpRequestException
```

### Source/destination JSON

```javascript
{
	"ClassName": "System.Net.Http.HttpRequestException",
	"Message": "A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond",
	"Data": {},
	"InnerException": {
		"ClassName": "System.Net.Sockets.SocketException",
		"Message": "A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond",
		"Data": {},
		"InnerException": null,
		"HelpURL": null,
		"StackTraceString": "   at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)",
		"RemoteStackTraceString": null,
		"RemoteStackIndex": 0,
		"ExceptionMethod": null,
		"HResult": -2147467259,
		"Source": "System.Private.CoreLib",
		"WatsonBuckets": null,
		"NativeErrorCode": 10060
	},
	"HelpURL": null,
	"StackTraceString": "   at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)\r\n   at System.Threading.Tasks.ValueTask`1.get_Result()\r\n   at System.Net.Http.HttpConnectionPool.CreateConnectionAsync(HttpRequestMessage request, CancellationToken cancellationToken)\r\n   at System.Threading.Tasks.ValueTask`1.get_Result()\r\n   at System.Net.Http.HttpConnectionPool.WaitForCreatedConnectionAsync(ValueTask`1 creationTask)\r\n   at System.Threading.Tasks.ValueTask`1.get_Result()\r\n   at System.Net.Http.HttpConnectionPool.SendWithRetryAsync(HttpRequestMessage request, Boolean doRequestAuth, CancellationToken cancellationToken)\r\n   at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)\r\n   at System.Net.Http.HttpClient.FinishSendAsyncBuffered(Task`1 sendTask, HttpRequestMessage request, CancellationTokenSource cts, Boolean disposeCts)",
	"RemoteStackTraceString": null,
	"RemoteStackIndex": 0,
[...]

PR #1897, opened 2018-10-27
Exclude TargetSite from non-serializable Exception serialization

JamesNK:
Fixes https://github.com/JamesNK/Newtonsoft.Json/issues/1798, https://github.com/JamesNK/Newtonsoft.Json/issues/1896, https://github.com/JamesNK/Newtonsoft.Json/issues/1854


0xced:
For reference: this change shipped as part of Newtonsoft.Json version 12.0.1.

Issue #1355, opened 2017-06-23
Issue with JsonTextReader as of version 10.0.1 (unchanged code was working fine on v.9.0.1)


opterios:
Hi,

I have a very large file (≈ 20GB) that contains an array of JSON objects, like this:

`{ "MyJsonObjects" : [{"Id":"1"},{"Id":"2"},{"Id":"3"}]}`

The array contains a few million records of "MyJsonObject" and I need to parse the file very fast. In order to achieve that, I have a parsing application running on several different machines and each instance is starting to read the file from a specific byte (different for every instance), so each instance is parsing a specific chunk of the file. So, if we say we had 10 million records to begin with, the 1000 instances running on 1000 CPUs will only have to parse 10,000 records each, reducing the parsing time to a few seconds, instead of several minutes.

I'm using the JsonTextReader and the JsonSerializer to parse the file. I place the stream to a specific byte that I feed each instance with, and I then deserialize 10,000 objects.

The "catchy" part of all this is that since I'm not parsing the whole json text, but only some records of the root array, the JsonTextReader is raising an exception each time I do a Read() between the records, because it reads the comma that separates the records.

I guess this is something expected, since I'm "hiding" from the JsonTextReader the complete file, so it thinks the stream contains invalid JSON text. This is why I just do a try-catch there to suppress this error (at the end of the code I'm posting below).

This has been working fine for quite some time now (several months), but as of version 10.0.1, the exception is causing the JsonTextReader to halt. As soon as this exception is raised, the JsonTextReader.Read() is returning null and the stream position no longer progresses.

The exact same code works fine as soon as I revert back to version 9.0.1.

Is this something intended? Maybe the stream would continue working even after the exception, as it was up to version 9.0.1?

In case you don't see this being fixed in the future, any advice would be more than welcome.

```
int position = 0;
using (StreamReader sr = new StreamReader(input))
using (JsonTextReader jr = new JsonTextReader(sr) { SupportMultipleContent = true })
{
    // Set the stream to a specific byte
    sr.BaseStream.Seek(nextByteToRead, SeekOrigin.Begin);

    JsonSerializer js = new JsonSerializer();
    while (position < 10000 && jr.Read())
    {
        if (jr.TokenType == JsonToken.EndArray) break;
        MyJsonClass jsonObject = js.Deserialize<MyJsonClass>(jr);
          // At this point, we have the jsonObject, so we handle it.
          position++;

          // After having finished, we want to go to the next record.
[...]

PR #1905, opened 2018-11-01
Added Record Separator (ASCII RS/0x1E) as a valid top level record separator in multiple content mode

yurikus:
Added Record Separator (ASCII RS/0x1E) as a valid top level record separator in multiple content mode. We use ASCII RS as record separator in our json record stream.

See:
https://github.com/JamesNK/Newtonsoft.Json/issues/1355
https://github.com/JamesNK/Newtonsoft.Json/commit/6aed848eca0bb7ee8612495c203dc44a71e0f12a

### DIFF
--- a/Src/Newtonsoft.Json/JsonTextReader.cs
+++ b/Src/Newtonsoft.Json/JsonTextReader.cs
@@ -1455,6 +1455,12 @@ namespace Newtonsoft.Json
                         // finished parsing
                         SetStateBasedOnCurrent();
                         return false;
+                    case '\x1E':
+                        _charPos++;
+
+                        // finished parsing
+                        SetStateBasedOnCurrent();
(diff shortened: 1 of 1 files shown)


yurikus:
@JamesNK Could you merge this please? A fairly simple change, which would helps us out a lot! Thanks!

yurikus:
Hi, thank you for quick response. ASCII defines 0x1e as Record Separator character, which was chosen to delimit individual records in the stream of multiple tens of megabytes of data. In not at the liberty to change the character. My benchmarking showed 40% perf gain when using JsonTextReader which is patched via reflection to skip the character vs reading the stream and then examining data. In my opinion, using this character is justified.

yurikus:
@JamesNK Hi James, what's the verdict?

PR #1909, opened 2018-11-03
Fix docs for generic custom json converter

tzachshabtay:
The previous text was for the non generic custom json converter sample (probably copy/paste error).


JamesNK:
Good catch. Thanks!

PR #1950, opened 2018-12-28
 Update Newtonsoft.Json project to use nullable reference types

JamesNK:
PCL builds disabled because of VS2019 issue https://developercommunity.visualstudio.com/content/problem/402902/class-library-cannot-be-compiled-gives-sdk-root-fo.html

Some warnings disabled with pragmas to workaround analysis errors

Fixed NRE with JToken.BeforeSelf()
Fixed NRE when using some converters to write a null value


TylerBrinkley:
Just curious, how painful was this process? I did it for one of my libraries awhile back in a separate branch and it was actually quite fun squashing the warnings and annotating my nullability intentions. Sounds like you fixed 2 bugs due to the feature!

Is this ready for `master` as VS2019 isn't released yet? Do you have a timeline for that? I'm assuming you'd like to get a release out before RTM?

Src/Newtonsoft.Json/JsonConvert.cs
```diff
@@ -526,7 +526,7 @@ public static string ToString(object value)
         [DebuggerStepThrough]
         public static string SerializeObject(object value)
```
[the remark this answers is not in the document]
JamesNK: Thanks. I haven't done the tests so there are probably lots of places that can take nullable inputs that haven't been properly marked as nullable.

JamesNK:
> Just curious, how painful was this process?

It sucked.

Src/Newtonsoft.Json/Linq/JPropertyKeyedCollection.cs
```diff
@@ -189,7 +191,7 @@ protected override void SetItem(int index, JToken item)
             }
         }
 
-        public bool TryGetValue(string key, out JToken value)
+        public bool TryGetValue(string key, [NotNullWhen(true)]out JToken? value)
```
TylerBrinkley: I've added the `using System.Diagnostics.CodeAnalysis` to my own code but it's not finding this `NotNullWhenAttribute` in VS 2019 16.3 Preview 1. Do you know what could be the issue?

JamesNK: I don't know. I use internal copies in this solution.

ahsonkhan: These were added to corefx a couple of months ago as part of preview 6: 
https://github.com/dotnet/corefx/pull/38020
https://github.com/dotnet/coreclr/pull/24679

https://github.com/dotnet/core/blob/master/release-notes/3.0/preview/api-diff/preview6/3.0-preview6_System.Diagnostics.CodeAnalysis.md

What version of the dotnet sdk do you have (if you run `donet --version` on the command line, you will see it)? I believe VS 16.3 preview 2+ has the newer SDK that probably has the change you are looking for. Alternatively, you can install the latest preview sdk independently of VS (at preview 7 now), or the nightly:
https://dotnet.microsoft.com/download/dotnet-core/3.0
https://github.com/dotnet/core-sdk/blob/master/README.md#installers-and-binaries

PR #1963, opened 2019-01-24
Make JsonSerializationException.Create public

Pzixel:
I'm writing my own `JsonConverter` which works perfectly, except for his erros being a bit cryptic.
There is an API that allows converters to provide more information, specifially about the place where error occured.
Unfortunately, there are internal for some reason.

This PR makes this methods public to solve this issue.

### DIFF
--- a/Src/Newtonsoft.Json/JsonSerializationException.cs
+++ b/Src/Newtonsoft.Json/JsonSerializationException.cs
@@ -115,14 +115,25 @@ namespace Newtonsoft.Json
             LinePosition = linePosition;
         }
 
-        internal static JsonSerializationException Create(JsonReader reader, string message)
+        /// <summary>
+        /// Initializes a new instance of the <see cref="JsonSerializationException"/> class
+        /// with a specified error message
+        /// </summary>
+        /// <param name="message">The error message that explains the reason for the exception.</param>
+        public static JsonSerializationException Create(JsonReader reader, string message)
         {
             return Create(reader, message, null);
         }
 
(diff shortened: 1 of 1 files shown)


JamesNK:
I don't like making internal methods public that aren't designed to be used publicly. Is there any reason you can't copy of the code from JsonSerializationException.Create and JsonPosition.FormatMessage, and throw the error yourself?

Pzixel:
Because copying method implementation is a bad practice marks it as "needed publicly"?

Is there any reasons why exception constructor is public but a thin wrapper on top of it is internal, and you suggest to copypaste its implementation?

bergmeister:
Using reflection you can still get you access to internal or private methods or constructors. Yes, reflection is not very performant but exceptions are not cheap in the first place, therefore using reflection should not have a big impact in your case and is probably the best compromise here

Pzixel:
So? I really hope this gets merged. What's the danger of making those methods public? I don't see any reason why they couldn't be public. Users cannot break it in any way and may find them useful: they acually will when defining their owns JsonConverters, because currently there is no way to express it via API and clearly Reflection backdoor cannot be considered as a reasonable alternative.

lostincomputer:
I also think that making `JsonSerializationException.Create` public is better that copy-pasting code. It also ensures that custom `JsonConverter`s populate `JsonSerializationException`s the same way as built-in ones.

forteddyt:
+1 on making `JsonSerializationException.Create` public. For the same reason @lostincomputer mentioned in wanting to create a consistent experience with our team's custom JsonConverters

Issue #1241, opened 2017-03-09
JObject.Parse vs DateTime


ujaroe:
Given the following code snippet
```
String js = "{ \"DateTime\": \"2017-03-07T17:26:40.352Z\" }";
var jo = JObject.Parse(js);
String ds1 = (String)jo["DateTime"];
```
results "ds1 = "03/07/2017 17:26:40"" which is not quite correct, because it misses fractions of second and timezone information. DateTime.Parse will not ne correct then.

Solybum:
It's behaviour actually seems to be correct.
The only thing different is the Kind, Local vs Utc, but the Hours are parsed correctly (I live in UTC -6)

```
DateTime dt = DateTime.Parse("2017-03-07T17:26:40.352Z");
Console.WriteLine("DateTime: {0}, Milliseconds: {1}, Kind: {2}", dt, dt.Millisecond, dt.Kind);
```
Output: DateTime: 07/03/2017 11:26:40, Milliseconds: 352, Kind: Local


```
JObject dtjson = JObject.Parse("{\"DateTime\": \"2017-03-07T17:26:40.352Z\"}");
Console.WriteLine("DateTime: {0}, Milliseconds: {1}, Kind: {2}", 
    dtjson["DateTime"], 
    ((DateTime)dtjson["DateTime"]).Millisecond, 
    ((DateTime)dtjson["DateTime"]).Kind);
```

Output: DateTime: 07/03/2017 17:26:40, Milliseconds: 352, Kind: Utc

gitfriend1:
Parfait

Aarondorn2:
I ran into this issue as well.  Specifically, we had configured JsonSerializerSettings for date handling and these settings were not being used by JObject.Parse().

A solution we chose was to use JsonConvert to parse into a JObject:
`JsonConvert.DeserializeObject<JObject>(js)`

ModestasK:
This is actually a serious issue for the systems that don't have that format understanding. For example RspberryPi (with Raspbian, which is ubuntu stretch).
If string comes like this:
`{ "datetime": "2019-01-27T09:26:57+02:00" }`
The string produced by `JObject.Parse(content)`will look like this:
`1/27/19 9:26:57 AM`
And that will fail, when trying to parse to DateTime object like that:
`DateTime.Parse(JObject.Parse(content))`
Error: 
`String '1/27/19 9:26:57' was not recognized as a valid DateTime.`

[...]

PR #1969, opened 2019-01-30
Workaround for issue #1241

k-yoshimoto:
Added DateParseHandling option to JsonLoadSettings as a workaround for #1241.

The problem is that JObject.Parse method creates a JsonTextReader with default settings to parse a JSON string.

When the reader finds a string like `2000-12-12T12:12:12+12:00`, [it parses the value as DateTime](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/JsonTextReader.cs#L218) since the default value of DateParseHandling is DateParseHandling.DateTime.

JToken.ToObject(JsonSerializer) method couldn't be a workaround because the token already converted the value to DateTime when it Parse-d and it has no information about the raw value.

I'm not sure if adding DateParseHandling option to JsonLoadSettings is a great idea but it should be an acceptable solution to work this around maintaining backward compatibilities.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Issues/Issue1241.cs
+++ b/Src/Newtonsoft.Json.Tests/Issues/Issue1241.cs
@@ -1,0 +1,58 @@
+﻿#region License
+// Copyright (c) 2007 James Newton-King
+//
+// Permission is hereby granted, free of charge, to any person
+// obtaining a copy of this software and associated documentation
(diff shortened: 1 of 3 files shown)


k-yoshimoto:
Closing this because JObject.Load method could be better workaround. https://github.com/JamesNK/Newtonsoft.Json/issues/1241#issuecomment-461283844

PR #1983, opened 2019-02-16
Use TypeConverter.ConvertTo to give additional type information

abbradar:
This gives more type information to the converter, which is useful for certain generic converters. My example converter is one that wraps values in F# newtypes:

```f#
 type ToNewtypeConverter () =
    inherit TypeConverter ()

    override this.ConvertTo (context : ITypeDescriptorContext, culture : CultureInfo, value : obj, destinationType : Type) : obj =
        let objectType = value.GetType()
        let (case, argType) = unionCases destinationType |> isNewtype |> Option.get
        FSharpValue.MakeUnion(case, [|value|])
```


abbradar:
Tests are failing and I need to think more about this, closing for now.

PR #1985, opened 2019-02-17
Fix #1984

### LINKED ISSUE #1984: Deserializing a null JSON property to a nullable ImmutableArray throws an exception
To generalize, the exception is thrown when there is a nullable value type that implements IEnumerable.
I believe it related to #1512 (and it is perhaps a regression from its fix).

The issue occurs in the version *12.0.1* while it was not present in *10.0.3*

### Destination types

```csharp
public class A
{
      public ImmutableArray<string>? Values { get; set; }
}
```

### Source JSON
```javascript
{ Values: null}
```

### Expected behavior

Test pass (`actual.Values` is null).

### Actual behavior

An exception with the following message is thrown;
```
Newtonsoft.Json.JsonSerializationException : Error converting value {null} to type 'System.Nullable`1[System.Collections.Immutable.ImmutableArray`1[System.String]]'. Path 'Values', line 1, position 14.
  ----> System.ArgumentException : Could not cast or convert from {null} to System.Collections.Immutable.ImmutableArray`1[System.String].
```

### Steps to reproduce
Paste the following test in the Newtonsoft.Json solution:

```csharp

#if !(NET20 || NET35 || NET40)
using System.Collections;
using System.Collections.Generic;
using System.Collections.Immutable;
#if DNXCORE50
using Xunit;
using Test = Xunit.FactAttribute;
using Assert = Newtonsoft.Json.Tests.XUnitAssert;
#else
using NUnit.Framework;
#endif

namespace Newtonsoft.Json.Tests.Issues
{
    public class IssueNullableStructImplementingIEnumerable
    {
        [Test]
        public void Test()
        {
            var actual = JsonConvert.DeserializeObject<A>("{ Values: null}");
            Assert.IsNotNull(actual);
            Assert.IsNull(actual.Values);
        }
        
        public class A
        {
            public ImmutableArray<string>? Values { get; set; }
        }
    }
}
#endif
```

Galad:
This is an attempt to fix #1984.
The issue is that the type passed to the contract does not allow to create a contract that indicates that the type is nulable and is an array at the same type.
The change is to pass the full type to the contract constructor?

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Issues/Issue1984.cs
+++ b/Src/Newtonsoft.Json.Tests/Issues/Issue1984.cs
@@ -1,0 +1,64 @@
+﻿#region License
+// Copyright (c) 2007 James Newton-King
+//
+// Permission is hereby granted, free of charge, to any person
+// obtaining a copy of this software and associated documentation
+// files (the "Software"), to deal in the Software without
+// restriction, including without limitation the rights to use,
+// copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following
+// conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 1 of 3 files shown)


JamesNK:
I made some additional changes to clean up. Will merge as part of PR https://github.com/JamesNK/Newtonsoft.Json/pull/2025

PR #1988, opened 2019-02-20
MissingMemberHandling.Error should not exception when ExtensionData is present

AartBluestoke:
This is a PR to make it not an error to deserialize a json element into ExtensionData when MissingMemberHandling.Error is set on the serializer.

Without this, catching the extra members in a parent with an expandable child (Scenario C) is not currently possible because otherwise valid objects deserialized into the expandable child exception (Scenario B throws MissingMemberHandling.Error Exception)

Three scenarios have been identified and units added:
a) (TestMissingMemberHandlingForDirectObjects) 
    it is not an error to have extra fields on an entity with ExtensionData present.
b) (TestMissingMemberHandlingForChildObjects) 
   it is not an error to have extra fields on children when the children have ExtensionData present
c) (TestMissingMemberHandlingForChildObjects) 
   it is an error to have extra fields on the parent, even when the children have ExtensionData present.

This required changing 2 checks in the JsonSerializerInternalReader to only throw the MissingMemberHandling exception when the relevant ExtensionData was null.

As Written this is a mildly breaking change, as objects which previously failed to deserialize into their ExtensionData, will now begin to. I'm happy to alter this PR to have MissingMemberHandling have a new third state which allows ExtensionData, while throwing on missing fields.

I implemented this as a breaking change, as when i explicitly say "put all other fields into this extension data dictionary" i do not expect an exception "nowhere to put this data".

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/ExtensionDataTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/ExtensionDataTests.cs
@@ -169,6 +169,54 @@ namespace Newtonsoft.Json.Tests.Serialization
             Assert.AreEqual(3, (int)o1[2]);
         }
 
+
+        public class SimpleExtendableObject
+        {
+            [JsonExtensionData]
+            public IDictionary<string, object> Data { get; } = new Dictionary<string, object>();
+        }
+
+        public class ObjectWithExtendableChild
+        {
+            public SimpleExtendableObject Data;
+        }
+        [Test]
+        public void TestMissingMemberHandlingForDirectObjects()
+        {
+            string json = @"{""extensionData1"": [1,2,3]}";
+            SimpleExtendableObject e2 = JsonConvert.DeserializeObject<SimpleExtendableObject>(json, new JsonSerializerSettings { MissingMemberHandling = MissingMemberHandling.Error });
+            JArray o1 = (JArray)e2.Data["extensionData1"];
+            Assert.AreEqual(JTokenType.Array, o1.Type);
+        }
+
+        [Test]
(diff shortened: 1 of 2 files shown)


AartBluestoke:
This PR came about because I have strict deserialization on the objects i consume (MissingMemberHandling.Error set), however one of the objects i want to consume has a unknown and variable schema, as it has a generic key-value store exposed as an unpredictable set of json properties, which i wanted to store into the ExtensionData dictionary. 
Without this PR, i have to turn of the MissingMemberHandler, add a ExtensionData  to the parent, and then have an onDeserialized check asserting that the dictionary is empty.

yatkimkaj:
Please speak English.  Exactly what are you doing, and what have you done


Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

elgonzo:
@yatkimkaj, the report is in English. No morning coffee yet?

yatkimkaj:
Yea something like that.  Why don’t you dumb it down for me please.  What is it u r trying to do?


Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

elgonzo:
Nobody is going to dumb down something for you here. This is an issue report about Newtonsoft.Json, not some place for tutoring someone who likes/needs stuff "dumbed down"

JamesNK:
Hi

I've added MissingMemberHandling to JsonObjectAttribute. I prefer this because it is not a breaking change and it keeps each feature simpler. Also now people can choose to opt-in to error on some objects and no others.

https://github.com/JamesNK/Newtonsoft.Json/pull/2050

Issue #1991, opened 2019-02-22
Memory leak using enumeration in Newtonsoft.Json v12


mantasaudickas:
Hello,

we have identified a memory leak when using Newtonsoft.Json v12.0.1.
There was a code written like this and executed multiple times:
```
var serialized = JsonConvert.SerializeObject(message, Formatting.None, new StringEnumConverter(true));
```

The code above was working without issues with v11, but consumes a lot of memory in v12.

Test code:
```csharp
[Fact]
public void SerializeManyTimes()
{
    var message = new Message();
    while (true)
    {
        JsonConvert.SerializeObject(message, Formatting.None, new StringEnumConverter(true));
    }
}

private class Message
{
    public string Content { get; set; } = "Sample";
    public MessageType Type { get; set; }
}

private enum MessageType
{
    Test
}
```

Same effect for both: serialize and deserialize methods.

Guessing that issue most likely is inside EnumUtils, which is adding items using key:
```
new StructMultiKey<Type, NamingStrategy>(enumType, namingStrategy) 
```

StructMultiKey has overridden Equals method:
```
return (Equals(Value1, other.Value1) && Equals(Value2, other.Value2));
```

But all `NamingStartegy` implementations does not have it..
That means test case:
```
Assert.True(Equals(new CamelCaseNamingStrategy(), new CamelCaseNamingStrategy()));
```
is failing and ConcurrentDictionary which is used in ThreadSafeStore is always creating (and keeping) new instances.

Workaround for us was to reuse StringEnumConverter as this is right thing to do anyway, but also this should be fixed inside serializer implementation.

JamesNK:
Fixed with the merged PR.

PR #1992, opened 2019-02-22
Fix memory leak when (de)serializing objects with enums (#1991)

mantasaudickas:
Implements Equals and GetHashCode methods in Naming Strategy.

### DIFF
--- a/Src/Newtonsoft.Json/Serialization/NamingStrategy.cs
+++ b/Src/Newtonsoft.Json/Serialization/NamingStrategy.cs
@@ -101,5 +101,44 @@ namespace Newtonsoft.Json.Serialization
         /// <param name="name">The property name to resolve.</param>
         /// <returns>The resolved property name.</returns>
         protected abstract string ResolvePropertyName(string name);
+
+        /// <summary>
+        /// Hash code calculation
+        /// </summary>
+        /// <returns></returns>
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                var hashCode = GetType().GetHashCode();     // make sure different types do not result in equal values
+                hashCode = (hashCode * 397) ^ ProcessDictionaryKeys.GetHashCode();
+                hashCode = (hashCode * 397) ^ ProcessExtensionDataNames.GetHashCode();
+                hashCode = (hashCode * 397) ^ OverrideSpecifiedNames.GetHashCode();
+                return hashCode;
+            }
+        }
+
+        /// <summary>
+        /// Object equality implementation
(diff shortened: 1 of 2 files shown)


Src/Newtonsoft.Json/Serialization/NamingStrategy.cs
```diff
@@ -101,5 +101,49 @@ public virtual string GetDictionaryKey(string key)
...
+        }
+
+        /// <summary>
+        /// Compare to another NamingStrategy
+        /// </summary>
+        /// <param name="other"></param>
+        /// <returns></returns>
+        protected bool Equals(NamingStrategy other)
+        {
+            return GetType() == other.GetType() &&
```
TylerBrinkley: `other` might be `null`. In `Equals(object)` I would just try to cast `obj as NamingStrategy` instead of using the `is` operator.

mantasaudickas: The only call to this method is from public override bool Equals(object obj) which already does the null check.

TylerBrinkley: It's protected so it may be called from a derived class.

mantasaudickas: Added null check.

TylerBrinkley: Since you do the null check here you can use the `as` operator instead in the other Equals method and make it a one-liner, perhaps as an expression-bodied member.

`public override bool Equals(object obj) => Equals(obj as NamingStrategy);`

mantasaudickas: Thank you, made another fix :)

JamesNK:
Thanks for the fix. I think this is ok, but I'll test it out more before 12.0.2.

For 13.0 I think I'll seal CamelCaseNamingStrategy and SnakeCaseNamingStrategy.

Narolas:
I think i've just come across this memory leak issue with entity framework + newtonsoft, using Enums in models. The memory object count rises slowly and the culprit seems to be Newtonsoft.Json.Serialization.CamelCaseNamingStrategy, Newtonsoft.Json.Utilities.EnumInfo and Newtonsoft.Json.Utilities.StructMultiKey.

I'm glad to see this has been found already ! When could we expect the 12.0.2 release ?

JamesNK:
12.0.2 will come soon. This is a subtle yet high-impact bug.

PR #1995, opened 2019-02-24
Remove duplication of memberName calculation

DamirAinullin:
Remove duplication of memberName calculation in according to DRY principle.


JamesNK:
Good spotting. Thanks

PR #1996, opened 2019-02-24
Fix unit test in according to the name

DamirAinullin:
There are three unit tests: DeserializeStringRegex_NoStartSlash_Error, DeserializeStringRegex_NoEndSlash_Error and DeserializeStringRegex_NoStartAndEndSlashes_Error.
Unit tests DeserializeStringRegex_NoEndSlash_Error and DeserializeStringRegex_NoStartAndEndSlashes_Error have the same body. It is not correct. In according to the test name regex in the unit test DeserializeStringRegex_NoStartAndEndSlashes_Error should not have start slash.


yatkimkaj:
Please stop with any devices that are in my name u can play with rays stuff the modem and the rest are all mine including my phone thx

yatkimkaj:
No stop

DamirAinullin:
@yatkimkaj What?

PR #2006, opened 2019-03-10
Update ReflectionUtils.cs#RemoveAssemblyDetails

xp44mm:
switch statement have two same cases. use fallthrough merged the two cases.


JamesNK:
Thanks for the PR. Merged!

PR #2007, opened 2019-03-11
Update ReflectionObject.cs#Create

xp44mm:
redundant but harmless: setter and getter is assigned same value two times.


xp44mm:
setter and getter will be assigned two times.

JamesNK:
Good spotting. Thanks

PR #2008, opened 2019-03-11
Update LateBoundReflectionDelegateFactory.cs#Instance

xp44mm:
use auto initial prop. return base type, and public in internal is internal.


JamesNK:
That internal property is actually used by an external library. I don't want to touch it because I could break that library. Thanks for looking into this.

PR #2013, opened 2019-03-15
Replaced ContainsKey and getting by value with TryGetValue method

DamirAinullin:
in order to improve perfomance


Src/Newtonsoft.Json/Linq/JPropertyKeyedCollection.cs
```diff
@@ -129,7 +129,7 @@ public bool Remove(string key)
 
             if (_dictionary != null)
             {
-                return _dictionary.ContainsKey(key) && Remove(_dictionary[key]);
+                return _dictionary.TryGetValue(key, out var value) && Remove(value);
```
TylerBrinkley: Jame's going to ask you to not use `var`.

JamesNK:
A nice small improvement. Thanks for the PR

PR #2016, opened 2019-03-19
Change constraint on DuplicatePropertyNameHandling

marta-krzyk-dev:
Assigning DuplicatePropertyNameHandling to Replace resulted in ArgumentOutOfRangeException, eg.
var jsonLoadSettings = new JsonLoadSettings { DuplicatePropertyNameHandling = DuplicatePropertyNameHandling.Replace };
The setter checks if value is lesser than 1 (Ignore) or greater than 2 (Error).
It should check if the value is lesser than 0 (Replace) or greater than 2 (Error).


JamesNK:
Opps. Thanks for the PR. Merged!

marta-krzyk-dev:
Glad I could help :)

Issue #2043, opened 2019-04-12
JArrray's Merge() implementation clears out array when merged with itself


pinath88:
### Steps to reproduce

var value = new JArray {"1", "2"};

value .Merge(value , new JsonMergeSettings { MergeArrayHandling = MergeArrayHandling.Replace });

### Expected behavior

[
  "1",
  "2"
]

### Actual behavior

[]

PR #2060, opened 2019-04-28
Fixed merging with self bug.

iRumba:
Issue https://github.com/JamesNK/Newtonsoft.Json/issues/2043

JArray cleared when merged with itself with setting `MergeArrayHandling = MergeArrayHandling.Replace`.

### DIFF
--- a/Src/Newtonsoft.Json/Linq/JContainer.cs
+++ b/Src/Newtonsoft.Json/Linq/JContainer.cs
@@ -1157,6 +1157,8 @@ namespace Newtonsoft.Json.Linq
 #endif
                     break;
                 case MergeArrayHandling.Replace:
+                    if (target == content)
+                        break;
                     target.ClearItems();
                     foreach (JToken item in content)
                     {
--- a/Src/Newtonsoft.Json.Tests/Linq/MergeTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Linq/MergeTests.cs
@@ -47,6 +47,14 @@ namespace Newtonsoft.Json.Tests.Linq
     public class MergeTests : TestFixtureBase
     {
         [Test]
+        public void MergeSelf()
+        {
+            var a = new JArray { "1", "2" };
+            a.Merge(a, new JsonMergeSettings { MergeArrayHandling = MergeArrayHandling.Replace });
+            Assert.AreEqual(a, new JArray { "1", "2" });
+        }
+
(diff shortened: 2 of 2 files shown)


Src/Newtonsoft.Json/Linq/JContainer.cs
```diff
@@ -1157,6 +1157,8 @@ internal static void MergeEnumerableContent(JContainer target, IEnumerable conte
 #endif
                     break;
                 case MergeArrayHandling.Replace:
+                    if (target == content)
```
304NotModified: Maybe this should be `Object.ReferenceEquals` instead of `==`, as this isn't a problem when the values are clones.

iRumba: This can be. But what about reason? I had execute benchmarks and comparable via operator works with same speed. If there are other reasons besides execution time, explain please.

``` ini

BenchmarkDotNet=v0.11.5, OS=Windows 10.0.18362
Intel Core i7-8700 CPU 3.20GHz (Coffee Lake), 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=2.2.300
  [Host]     : .NET Core 2.2.5 (CoreCLR 4.6.27617.05, CoreFX 4.6.27618.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.2.5 (CoreCLR 4.6.27617.05, CoreFX 4.6.27618.01), 64bit RyuJIT


```
|          Method |      Mean |     Error |    StdDev | Rank |
|---------------- |----------:|----------:|----------:|-----:|
|     OperatorDif | 1.0026 ns | 0.0146 ns | 0.0122 ns |    1 |
|    OperatorSame | 0.9954 ns | 0.0039 ns | 0.0035 ns |    1 |
|  ObjectReferDif | 0.9965 ns | 0.0100 ns | 0.0083 ns |    1 |
| ObjectReferSame | 0.9994 ns | 0.0031 ns | 0.0027 ns |    1 |

304NotModified: The main reason IMO is that it's only needed if the references are the same. But equals is also fine to me. 
 

Maybe it's slower with a large (nested) jarray, but that wasn't my main reason

JamesNK:
Thanks! Merged here - https://github.com/JamesNK/Newtonsoft.Json/pull/2134

Issue #2089, opened 2019-06-15
memory leak with dynamic classes?


srinivasupadhya:
### Source/destination types

```csharp
using Newtonsoft.Json;
using System;
using System.Collections.Generic;
using System.Reflection;
using System.Reflection.Emit;

namespace newtonsoft
{
    public class Program
    {
        public static void Main(string[] args)
        {
            Console.WriteLine("collect start snapshot");
            Console.ReadLine();

            for (int i=0; i < 10; i++)
            {
                string json = "{}";
                Type dynamicClass1Type = MyTypeBuilder.CompileResultType();

                Deserialize(json, dynamicClass1Type);

                GC.Collect();

                Console.WriteLine("collect snapshot");
                Console.ReadLine();
            }
        }

        private static void Deserialize(string json, Type type)
        {
            JsonConvert.DeserializeObject(json, type);
        }
    }

    public static class MyTypeBuilder
    {
        public static void CreateNewObject()
        {
            var myType = CompileResultType();
            var myObject = Activator.CreateInstance(myType);
        }
        public static Type CompileResultType()
        {
            TypeBuilder tb = GetTypeBuilder();
            ConstructorBuilder constructor = tb.DefineDefaultConstructor(MethodAttributes.Public | MethodAttributes.SpecialName | MethodAttributes.RTSpecialName);

            // NOTE: assuming your list contains Field objects with fields FieldName(string) and FieldType(Type)
            Dictionary<string, Type> fields = new Dictionary<string, Type>() { { "number1", typeof(Int64) }, { "text1", typeof(string) } };
            foreach (KeyValuePair<string, Type> field in fields)
                CreateProperty(tb, field.Key, field.Value);

            Type objectType = tb.CreateType();
            return objectType;
        }

        private static TypeBuilder GetTypeBuilder()
        {
            var typeSignature = "DynamicClass1";
            var an = new AssemblyName(typeSignature);
            AssemblyBuilder assemblyBuilder = AppDomain.CurrentDomain.DefineDynamicAssembly(an, AssemblyBuilderAccess.Run);
            ModuleBuilder moduleBuilder = assemblyBuilder.DefineDynamicModule("MainModule");
            TypeBuilder tb = moduleBuilder.DefineType(typeSignature,
                    TypeAttributes.Public |
                    TypeAttributes.Class |
                    TypeAttributes.AutoClass |
                    TypeAttributes.AnsiClass |
[...]

PR #2105, opened 2019-07-19
add test to reproduce the problem

srinivasupadhya:


srinivasupadhya:
added test to reproduce #2089

PR #2087, opened 2019-06-13
Update crefs

mderriey:
Right now I see the following in VS built-in decompiler:
```
Date formatted strings, e.g. "\/Date(1198908717056)\/" and "2012-03-21T05:40Z", are parsed to Newtonsoft.Json.DateParseHandling.DateTime.
```

PR #2111, opened 2019-07-31
Fix date parse handling summary

Prologh:
There is a small typo in `DateTime` and `DateTimeOffset` references in `DateParseHandling` summaries causing the references to be translated to `DateTime` and `DateTimeOffset` enum members instead of structs from `System` namespace, which I belive is the expected result. I added full namespace path `System`, so that should do the trick.


304NotModified:
PS the fix is also here (and still open)
https://github.com/JamesNK/Newtonsoft.Json/pull/2087/files

JamesNK:
Thanks. I merged https://github.com/JamesNK/Newtonsoft.Json/pull/2087

PR #2113, opened 2019-08-02
Improve performance of calling ICustomTypeDescriptor.GetProperties

304NotModified:
Hi,

I noticed this 

> A frequent performance issue that comes up is JSON.NET's JObject, which for reasons we haven't fully researched, has very slow reflection characteristics when used as a model in Handlebars.Net. 

from https://github.com/rexm/Handlebars.Net/blob/master/README.md

also see https://github.com/rexm/Handlebars.Net/issues/56#issuecomment-114599359

I checked the code of `ICustomTypeDescriptor.GetProperties` and I see that `PropertyDescriptorCollection` starts with a capacity of 0... (see https://referencesource.microsoft.com/#System/compmod/system/componentmodel/PropertyDescriptorCollection.cs,97) - not sure if that's the real issue, but it isn't a good init capacity if you known the capacity already ;)

So I improved that by creating an own array and don't use the PropertyDescriptorCollection.Add

### DIFF
--- a/Src/Newtonsoft.Json/Linq/JObject.cs
+++ b/Src/Newtonsoft.Json/Linq/JObject.cs
@@ -744,13 +744,15 @@ namespace Newtonsoft.Json.Linq
 
         PropertyDescriptorCollection ICustomTypeDescriptor.GetProperties(Attribute[] attributes)
         {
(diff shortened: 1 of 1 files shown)


JamesNK:
Thanks! Merged here https://github.com/JamesNK/Newtonsoft.Json/pull/2136

PR #2149, opened 2019-09-07
Added a new way to parse floats as a string

mdiller:
I noticed if you read a file with a high precision double in it like `0.094369298521547657`, if you read a file into a JToken and then dump it back to a file, the value will be different. I'm working on building a library that reads and writes json files and it would be very nice if I could count on these values not changing. Because `double` and `decimal` don't have high enough precision, I propose that we add a new `FloatParseHandling` value of `String` so that we can support reading and writing these values without modifying them.

Is this change ok?
If so should I add some testing for this or make the modifications elsewhere?


mdiller:
@JamesNK Does this suggested change look ok and how do I get it approved? Thanks.

mdiller:
I found a better way to do this by reading everything in as a `decimal` and being careful to never try to convert it to a `double`, without needing changes to the library. Closing this PR.

PR #2161, opened 2019-09-13
Added KebabCaseNamingStrategy

### LINKED ISSUE #2160: Add a KebabCaseNamingStrategy
No idea how I'm the first person to request this...but can we please get a `KebabCaseNamingStrategy`? Requests sent to Apple Push Notification Service [require kebab-case](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/generating_a_remote_notification#2943360) JSON fields...I'm sure there are other examples out there.

### Source types

```cs
[JsonObject(NamingStrategyType = typeof(KebabCaseNamingStrategy))]
public class MyType
{
    public string SomeString { get; set; }
    public bool SomeBool { get; set; }
    public int SomeInt { get; set; }
}
```

### Destination JSON

```javascript
{
  "some-string":"some string",
  "some-bool": true,
  "some-int": 5,
}
```

### Expected behavior

This would be possible

### Actual behavior

It's not, unless I derive from `NamingStrategy` myself, like one of [these](https://gist.github.com/amilos/9ec30c60bc50b40bf0cfabb8fdec4966) [examples](https://gist.github.com/duruld/a67a80f05cb610fd39f6c2e6ad03844d) on the internets.

TylerBrinkley:
Closes #2160

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Documentation/Samples/Serializer/NamingStrategyKebabCase.cs
+++ b/Src/Newtonsoft.Json.Tests/Documentation/Samples/Serializer/NamingStrategyKebabCase.cs
@@ -1,0 +1,86 @@
+﻿#region License
+// Copyright (c) 2007 James Newton-King
+//
+// Permission is hereby granted, free of charge, to any person
+// obtaining a copy of this software and associated documentation
+// files (the "Software"), to deal in the Software without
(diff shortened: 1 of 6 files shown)


JamesNK:
Thanks. Looks good to me! Merged.

PR #2164, opened 2019-09-16
Enhance the ability to find and compare missing members and required properties

bigjonroberts:
Our team needs a way to isolate exceptions for missing members.
We want to be able to check for specific properties on specific object types so that we can avoid raising exceptions on certain json schema deprecated members, bug just log warnings instead.

We could have attempted to catch and parse exception messages with the existing exception type, but that seemed risky that the names could change later.

Design decisions that need review:

- Right now, this is one exception type to cover all member read exceptions. It might be better to split out MissingMember and Required exceptions into their own types?
- Write exceptions were not converted.
- The property `ObjectTypeName` was created instead of making a property of type `Type`.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
@@ -370,13 +370,31 @@ namespace Newtonsoft.Json.Tests.Serialization
         [Test]
         public void CoercedEmptyStringWithRequired()
         {
-            ExceptionAssert.Throws<JsonSerializationException>(() => { JsonConvert.DeserializeObject<Binding>("{requiredProperty:''}"); }, "Required property 'RequiredProperty' expects a value but got null. Path '', line 1, position 21.");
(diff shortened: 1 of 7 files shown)


bigjonroberts:
Regarding the warning messages and failing build.
I'll merge master, update null handling, and try building with sdk version `3.0.100-preview9-014004`.
I couldn't find a dotnet core sdk download for `3.0.100-preview9-013927`, but I found the official preview 9 release, 3.0.100-preview9-014004, over [here](https://github.com/dotnet/core/blob/master/release-notes/3.0/preview/3.0.0-preview9-download.md).

bigjonroberts:
Looks like we'll also need to enhance the types/flow for Error handler for our use-case.  I'm going to add that, too.

bigjonroberts:
@JamesNK I've added the remaining tests and samples.
This is fully compatible with the old error handling except the messages have been enhanced to add the Object type name in some cases. Do you think that constitutes a breaking change?
My other main design question is around the use of the enum to denote the three error scenarios instead of two new exception classes with a bool flag or three new exception classes.
Unless someone was checking for an exact error message, this shouldn't break any existing code, as the exception class derives from the same exception class that was created before.

Issue #2162, opened 2019-09-15
Trailing zeros are removed after parse & format


caioproiete:
I'm using [**JSON .NET v12.0.2**](https://www.nuget.org/packages/Newtonsoft.Json/12.0.2) to "_beautify_" minified JSON tokens that are being written to a store by multiple applications. Sometimes these tokens have numbers with one or more trailing zeros, for example: `{"x":100.50,"y":200.400}`. The trailing zeros are removed in the indented JSON token.

I'm assuming this is caused by numbers being parsed to `decimal` and the information is being lost.

Is there any workaround?

### Expected behavior

```json
Minified:
{"x":100.50,"y":200.400}

Indented:
{
  "x": 100.50,
  "y": 200.400
}
```

### Actual behavior

```json
Minified:
{"x":100.50,"y":200.400}

Indented:
{
  "x": 100.5,
  "y": 200.4
}
```

### Steps to reproduce

```csharp
var jsonStringMinified = @"{""x"":100.50,""y"":200.400}";

var parsedJson = JToken.Parse(jsonStringMinified);
var jsonStringIndented = parsedJson.ToString(Newtonsoft.Json.Formatting.Indented);

Console.WriteLine("Minified:");
Console.WriteLine(jsonStringMinified);
Console.WriteLine();
Console.WriteLine("Indented:");
Console.WriteLine(jsonStringIndented);
```

Seems to be related to #1884 and #1367.

TylerBrinkley:
You need to specify the `FloatParseHandling` as `Decimal` when parsing to a `JToken` as by default your floating point numbers are being parsed as `double`s.

caioproiete:
Thanks @TylerBrinkley. Your reply pointed me in the right direction.

Unfortunately it is not possible to specify `FloatParseHandling`  when using `JToken.Parse`, as [it does not expose a way for configuring settings on the `JsonTextReader` that it creates internally](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Linq/JToken.cs#L2197).

[![image](https://user-images.githubusercontent.com/177608/64990690-aa9b8380-d8a6-11e9-8e0d-23b37356cf96.png)](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Linq/JToken.cs#L2197)

---

[...]

PR #2167, opened 2019-09-17
Allow configuring JsonTextReader settings via JsonToken.Parse and JArray.Parse

augustoproiete:
By default, floating point numbers are parsed as `double` because the default value for `FloatParseHandling` is `Double`. This causes trailing zeros to be removed when using `JsonToken.Parse`, as described in #2162. And the same happens with `JArray.Parse`.

Because of that, parsing the JSON below with `JsonToken.Parse`
```json
{"x":100.50,"y":200.400}
```

Results in:
```json
{"x":100.5,"y":200.4}
```

And the trailing zeros are lost...

---

This PR adds method overloads to `JsonToken.Parse` and `JArray.Parse` to allow the developer to specify `xxxHandling` settings to configure the `JsonTextReader`.

Example of usage:

```csharp
JObject o = (JObject)JToken.Parse(@"{""x"":100.50,""y"":200.400}",
    new JsonTextReaderSettings
    {
        FloatParseHandling = FloatParseHandling.Decimal,
    });

string json = o.ToString(Formatting.None);
```

The code above results in the JSON below, preserving the trailing zeros as expected:

```json
{"x":100.50,"y":200.400}
```

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Linq/JArrayTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Linq/JArrayTests.cs
@@ -647,5 +647,28 @@ Parameter name: index",
             Assert.AreEqual(true, ((IJsonLineInfo)a[1]).HasLineInfo());
             Assert.AreEqual(true, ((IJsonLineInfo)a[2]).HasLineInfo());
         }
+
+        [Test]
+        public void ParseAndToStringWithFloatingNumbersAsDouble()
+        {
+            JArray a = JArray.Parse("[1.0,2.00,3.000]");
+            string json = a.ToString(Formatting.None);
+
+            Assert.AreEqual("[1.0,2.0,3.0]", json);
+        }
+
+        [Test]
+        public void ParseAndToStringWithFloatingNumbersAsDecimal()
+        {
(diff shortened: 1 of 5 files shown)


TylerBrinkley:
If this is to be added, overloads should be added to `JObject.Parse` as well.

Also, `DateFormatString` should be added to `JsonTextReaderSettings` and a constructor should be added to `JsonTextReader` that accepts a `JsonTextReaderSettings` object.

JamesNK:
I'm not a fan of this. If you need to customize the reading of JSON from a string then you would create a `JsonTextReader` over the string (`new JsonTextReader(new StringReader(json))`) and pass that to `Load(JsonTextReader)`.

Adding a new settings object introduces new places settings have to be maintained, kept in sync with the reader, and copied between.

It isn't much of an improvement over what you can do today:

```cs
JObject o = (JObject)JToken.Load(
    new JsonTextReader(new StringReader(@"{""x"":100.50,""y"":200.400}"))
    {
        FloatParseHandling = FloatParseHandling.Decimal,
    });
```

augustoproiete:
Thank you gents. Using `JToken.Load` or `JsonConvert.DeserializeObject<JToken>` are good workarounds for setting the `FloatParseHandling`.

PR #2168, opened 2019-09-18
Net Standard 2.1

bigjonroberts:
@JamesNK Since dotnet core 3.0 is approaching release.


JamesNK:
What is the benefit of targeting netstandard2.1? As far as I know the netstandard2.0 assembly will work fine. I don't want to add an extra assembly and increase the package size for no reason.

bigjonroberts:
Most of the improvements would be performance related.

@JamesNK 
Immediately, these code blocks will become active:
https://github.com/JamesNK/Newtonsoft.Json/search?q=HAVE_REFLECTION_EMIT&unscoped_q=HAVE_REFLECTION_EMIT

I also assume that there are potential performance improvements from adopting the updated `Span<>` API's for streams and potentially by leveraging the new `ValueTask<>`.  Though I haven't gone through the code to identify those yet and I don't know if you have any tests that specifically address profiling for performance.

foriequal0:
`aspnetcore` supports `Controllers` to return `IAsyncEnumerable<T>` which is added in 2.1 by buffering it with `AsyncEnumerableReader` (https://github.com/dotnet/aspnetcore/blob/d5fd9fc2faeb079e05e3db31e29dcdea31841eec/src/Mvc/Mvc.NewtonsoftJson/src/NewtonsoftJsonResultExecutor.cs#L138)

A Swagger library [NSwag](https://github.com/RicoSuter/NSwag) and its schema generation library [NSwagSchema](https://github.com/RicoSuter/NJsonSchema) depends on `Newtonsoft.Json`, and it should support `IAsyncEnumerable<T>` too.
Someone made an issue on them about the schema of `IAsyncEnumerable<T>` is not an array-like type. (https://github.com/RicoSuter/NSwag/issues/2582)
The maintainer closed the issue by an patch, but they wrongly believed that the [IContractResolver.ResolveContract](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Serialization/IContractResolver.cs#L44) would resolve it as `JsonArrayContract`, but it wasn't. It resolves `IAsyncEnumerable<T>` as `JsonObjectContract` (https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Serialization/DefaultContractResolver.cs#L1189).
I've corrected the issue by making a workaround on it, but I think `Newtonsoft.Json` should support it properly.

Issue #2176, opened 2019-09-28
Constructor deserialization throws an exception for constructor with more than 256 parameters


horato:
### Source/destination types

```csharp
            var classBuilder = new StringBuilder();
            var jsonBuilder = new StringBuilder();
            var ctorParametersBuilder = new StringBuilder();

            classBuilder.AppendLine("public class Dummy");
            classBuilder.AppendLine("{");
            jsonBuilder.AppendLine("{");
            for (var i = 0; i <= byte.MaxValue + 1; i++)
            {
                classBuilder.AppendLine($"\tpublic int _{i} {{ get; }}");
                jsonBuilder.AppendLine($"\t\"_{i}\" : {i},");
                ctorParametersBuilder.Append($"int p{i},");
            }

            var ctorParameters = ctorParametersBuilder.ToString();
            ctorParameters = ctorParameters.Remove(ctorParameters.Length - 1, 1);
            classBuilder.Append("\tpublic Dummy(");
            classBuilder.Append(ctorParameters);
            classBuilder.AppendLine(")");
            classBuilder.AppendLine("\t{");
            classBuilder.AppendLine();
            classBuilder.AppendLine("\t}");
            classBuilder.AppendLine("}");
            jsonBuilder.AppendLine("}");

            var classContent = classBuilder.ToString();
            var jsonContent = jsonBuilder.ToString();
```

The json and class contents are too large to post here, I've created a code snippet which will generate the class and json for you.

### Expected behavior

Input json deserializes into class correctly

### Actual behavior
InvalidOperationException is thrown inside DynamicReflectionDelegateFactory->ILGenerator. The Ldloc_S OpCode only supports up to 256 local variables when generating the constructor.  Solution is to use Ldloc instead when there are more than 256 ctor parameters.
https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.opcodes.ldloc_s
https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.opcodes.ldloc

### Steps to reproduce

```csharp
var obj = JsonConvert.DeserializeObject<Dummy>(jsonContent);
```

JamesNK:
Cool bug. Thanks for the good repro. Fixed.

PR #2177, opened 2019-09-29
Fix deserializing into constructors with more than 256 parameters

JamesNK:
Fixes https://github.com/JamesNK/Newtonsoft.Json/issues/2176


canton7:
(Note that `ILGenerator` already has the logic to use `ldloc.s`/`ldloc.0`/etc if possible, as long as you use `OpCodes.Ldloc` and not `OpCodes.Ldloc_S`: [source](https://referencesource.microsoft.com/#mscorlib/system/reflection/emit/ilgenerator.cs,861)).

PR #2184, opened 2019-10-03
Implement TypeNameHandling.RootObject

### LINKED ISSUE #2183: Add a new "Root" option for TypeNameHandling
I'd like to use the `TypeNameHandling` feature in one of my projects, unfortunately none of the existing options in the `TypeNameHandling` enum does exactly what I need.

Basically, I want the behavior of `Auto`, except that I want the `$type` property to also be emitted for the root object. I know there's a way to do it, by explicitly specifying the type in the call to `SerializeObject`. However, in my use-case, it's not a very convenient option: I'm using JSON.NET indirectly, through the Cosmos DB SDK, and it doesn't give me the option to specify the type explicitly¹.

I'm aware of #990, and I understand that you don't want to change the current behavior of `Auto`, so it's not what I'm suggesting (it would be a breaking change for too many people anyway).

However, the behavior I want could be supported without breaking anything, by introducing a new `Root` value for the `TypeNameHandling` enum. This value would mean *always emit the `$type` property for the root object*, and could be combined with `Auto` to achieve the behavior I described.

If you're willing to go ahead with this approach, I'll POC it and submit a PR soon.

---

*¹ that's not exactly true: the `CreateItemAsync` method is generic, and the generic type argument is passed down to the serializer. But because of type inference, it's very easy to forget to specify the type, and write `CreateItemAsync(new Derived())` instead of `CreateItemAsync<Base>(new Derived())`, resulting in missing `$type` property. This is too error prone, I can't count on myself or my colleagues to never forget it.*

thomaslevesque:
Fixes #2183


thomaslevesque:
Any feedback on this?

tobias-graf-p:
Any feedback on this from the maintainer of the lib? This PR looks good/ready to me ...

PR #2193, opened 2019-10-15
Add check for NullValueHandling in SerializeDictionary

### LINKED ISSUE #1997: NullValueHandling.Ignore doesn't ignore nulls
### Source/destination types

```csharp
System.Collections.Generic.Dictionary<string, object>
```

### Source/destination JSON

```javascript
{"exists":1,"doesnt":null}
```

### Expected behavior

That the null is not included.

### Actual behavior

The null is included.

### Steps to reproduce

```csharp
using System.Collections.Generic;
using Newtonsoft.Json;

namespace Sample
{
    public class Program
    {
        {
            var obj = new Dictionary<string, object>
            {
                ["exists"] = 1,
                ["doesnt"] = null,
            };

            var json = JsonConvert.SerializeObject(obj, new JsonSerializerSettings
            {
                NullValueHandling = NullValueHandling.Ignore
            });

            System.Console.WriteLine(json);
        }
    }
}
```

mdalepiane:
Add check for NullValueHandling in SerializeDictionary, this fixes #1997.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/NullValueHandlingTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/NullValueHandlingTests.cs
@@ -26,6 +26,7 @@
 using System;
 using System.IO;
 using Newtonsoft.Json.Tests.TestObjects;
+using System.Collections.Generic;
 #if DNXCORE50
 using Xunit;
 using Test = Xunit.FactAttribute;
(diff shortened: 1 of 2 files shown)


mdalepiane:
@JamesNK what do you think about this fix?

mdalepiane:
@JamesNK any thoughts here?

Issue #2201, opened 2019-10-23
JsonContainerAttribute Does Not Respect NamingStrategyParameters for Dictionaries


adamrodger:
### Source/destination types

```csharp
[JsonObject(NamingStrategyType = typeof(SnakeCaseNamingStrategy), NamingStrategyParameters = new object[] { true, true })]
public class UsesOverriddenNamingStrategy
{
    public int Id { get; set; }
    public string FullName { get; set; }
    public IDictionary<string, int> ExtraThings { get; set; }
}
```

### Source/destination JSON

```javascript
// actual value:
{"id":42,"full_name":"John Doe","extra_things":{"ExtraValue":1}}

// expected value:
{"id":42,"full_name":"John Doe","extra_things":{"extra_value":1}}
```

### Steps to reproduce

```csharp
var foo = new UsesOverriddenNamingStrategy
{
    Id = 42,
    FullName = "John Doe",
    ExtraThings = new Dictionary<string, int>
    {
        ["ExtraValue"] = 1
    }
}

string json = JsonConvert.SerializeObject(foo);
```

### Expected behavior

I expected the dictionary keys to be converted to snake case because the `NamingStrategyParameters` were `true` (i.e. `ProcessDictionaryKeys == true`).

### Actual behavior

The property names were overridden (so `NamingStrategyType` was used) but dictionary keys were not converted. They use the naming strategy from the default contract resolver instead of the one specified in either the `JsonObject` or `JsonProperty` override.

adamrodger:
That's fine and all, but then that probably means that `[JsonProperty]` shouldn't allow specifying `NamingStrategyParameters` if they aren't actually going to be used. This is misleading:

```csharp
public class Foo
{
    [JsonProperty(
      NamingStrategyType = typeof(CamelCaseNamingStrategy), // <-- this will be used
      NamingStrategyParameters = new object[] { true, true })] // <-- but these won't
    public IDictionary<string, int> Bar { get; set; }
}
```

In that example, the first of those `NamingStrategyParameters` is specifically related to the handling of dictionary keys, and it's applied to a dictionary, and yet they aren't used. That doesn't really follow the Principle of Least Surprise.

I've added a PR which allows those parameters to be bound to the property so that they will be used during serialization.

[...]

PR #2203, opened 2019-10-23
#2201 Allowed NamingStrategy override on dictionary keys

adamrodger:
See #2201 for detail. Unit tests executed and all passing locally on `netcoreapp3.0`


adamrodger:
...unless of course you accidentally ran the test suite on `master` instead of the PR branch...

PR #2224, opened 2019-11-27
Bugfix: Do not treat ignored field as missing member when deserializing from overriden json constructor

everfor:
If all of the following conditions are true:

- A class has any members marked as "JsonIgnore"
- The class has a custom JsonContructor with parameters that do not include those ignored members
- The SerializationSetting has MissingMemberHandling set to "Error"

Then a bug occurs, where Newtonsoft treats those ignored (but existing) members as missing members, and throws an exception.

Gist to reproduce this bug: https://gist.github.com/everfor/6d5ec290252da9de46c943882f24708d

This PR fixes the bug.

Add UT "DeserializeIgnoredPropertyInConstructorWithoutThrowingMissingMemberError" to confirm the fix works

The bug is also covered in https://stackoverflow.com/questions/59062524/property-with-jsonignore-attribute-not-ignored-when-deserializing-with-missingme


JamesNK:
Yeah, that's a bug. Thanks for the fix and unit tests.

Merged!

Issue #2251, opened 2020-01-02
NamingStrategy: Support numbers as words option


mchandschuh:
I'm integrating an API using kebab casing and views numbers as their own words. For example, for a C# property name of `Street1` (of an `Address` object), the serialized member name needs to be `street-1`. Now, this is easily achievable using the `[JsonProperty]` syntax _if_ you have access to the model type, and if not, a custom `NamingStrategy` to take matters into your own hands. Another option is updating `StringUtils.ToSnakeCase` and `StringUtils.ToKebabCase` to support an optional `numbersAsWords` parameter and adding `SeparatedCaseState.Number` to the enum along with an accompanying case handler in the `ToSeparatedCase` helper method.

### Source/destination types

```csharp
public class Address
{
    public string Street1 { get; set; }
    public string Street2 { get; set; }
}
```

### Source/destination JSON

```javascript
{
    "street-1": "123 Hickory Street",
    "street-2": "Suite 36D"
}
```

### Expected behavior
The json.net provides options to support the above naming strategy out-of-the-box.

### Actual behavior
The json.net API does not support viewing numbers as their own words via the provided `NamingStrategy` implementations.

PR #2252, opened 2020-01-02
Support numbers as words in SnakeCaseNamingStrategy and KebabCaseNamingStrategy

mchandschuh:
Existing implementation does not treat numbers in any special way. They're treated most closely
to lowercase characters. This change provides an options 'numbersAsWords' that permits viewing
numbers as capable of starting and stopping a word.

Closes JamesNK/Newtonsoft.Json#2251


mchandschuh:
As a note, the generated property names for things such as an ISO timestamp with this option enabled is straight up hideous :) -- I didn't make an attempt to make it look pretty, nor do I think it should be changed. As is, the property is explicit as to how it impacts member name generation and doesn't try to do anything fancy to prevent absurd results. Also, providing these cases via the tests provides _tests as documentation_ so at least it's obvious as to what the expected behavior is, despite how unlikely someone would actually desire such an output.

PR #2304, opened 2020-03-25
Fix wrong define is used in StringUtils.ToLower()

Si1ver:
In `StringUtils.ToLower()` call to `char.ToLower(char, CultureInfo)` should be wrapped in `#if HAVE_CHAR_TO_LOWER_WITH_CULTURE` instead of `#if HAVE_CHAR_TO_STRING_WITH_CULTURE`.


JamesNK:
Thanks!

Issue #2222, opened 2019-11-20
DateTimeOffset deserialization works differently for JsonConvert.DeserializeObject and JObject.ToObject


JonathanLydall:
### Source/destination types

```csharp
class Poco
{
    public DateTimeOffset Local { get; set; }
    public DateTimeOffset Utc { get; set; }
}
```

### Source/destination JSON

```javascript
{"Local":"2019-11-20T14:58:14.8197628+02:00","Utc":"2019-11-20T12:58:14.8198451+00:00"}
```

### Expected behavior

The `DateTimeOffset.Offset` of the deserialized properties to be the same as per the serialized properties.

### Actual behavior

The `DateTimeOffset.Offset` of the `LocalProperty` is `00:00:00`.

### Steps to reproduce

```csharp
using System;
using Microsoft.VisualStudio.TestTools.UnitTesting;
using Newtonsoft.Json;
using Newtonsoft.Json.Linq;

namespace UnitTestProject1
{
    [TestClass]
    public class UnitTest1
    {
        [TestMethod]
        public void TestMethod1()
        {
            var original = new Poco
            {
                Local = DateTimeOffset.Now,
                Utc = DateTimeOffset.UtcNow
            };

            var serialized = JsonConvert.SerializeObject(original);
            var deserializedDirectly = JsonConvert.DeserializeObject<Poco>(serialized);
            var deserializedThroughJObject = JObject.Parse(serialized).ToObject<Poco>();

            Assert.AreEqual(original.Local.Offset, deserializedDirectly.Local.Offset);
            Assert.AreEqual(original.Utc.Offset, deserializedDirectly.Utc.Offset);
            Assert.AreEqual(original.Local.Offset, deserializedThroughJObject.Local.Offset);
            Assert.AreEqual(original.Utc.Offset, deserializedThroughJObject.Utc.Offset);
        }
    }

    class Poco
    {
        public DateTimeOffset Local { get; set; }
        public DateTimeOffset Utc { get; set; }
    }
}
```

This seems similar to #1531.

rhyous:
I am experiencing this as well.

I am in MDT time, and both of the below unit tests fail. When I'm in 

```
        [TestMethod]
        public void JObect_Keeps_TimeZone()
        {
            // Arrange
            var testObject = new { EndDate = new DateTimeOffset(2020, 1, 1, 0, 0, 0, 0, TimeSpan.Zero) };
            var expected = testObject.EndDate.ToString();
[...]

PR #2305, opened 2020-03-27
Allow for DateParseHandling to be set in JsonLoadSettings.

rhyous:
This is an option for allowing me to configure JObject to use DateParseHandling.DateTimeOffset.
https://github.com/JamesNK/Newtonsoft.Json/issues/2222


rhyous:
@JamesNK 

This is an option that works. Would you prefer it to be part of JsonLoadSettings, or would you prefer it coded differently?

kirillmyp:
Hello!
I don't get it too about time zone
Is it bug or feature ?
Because if i want to safe a time zone when using JsonConverter, need to add `serializer.DateTimeZoneHandling = DateTimeZoneHandling.Utc;`

PR #2317, opened 2020-04-21
Implement serialization of comments

### LINKED ISSUE #1211: Json5 and comments
Would be nice if there was a way to add comments to fields that also appear in the json file.. like

    [RangeAttribute(0.01f, 50f), JsonComment("Blah Description (Range 0.01f, 50f)")]
    public float GrowIncrement = 0.42f;

You can currently do some sort of dumb crap like JsonProperty("GrowIncrement (Range 0.01f, 50f)")]
but its just ugly to see in json file not to mention.... when a comment line above it would be better.

So the serialized json gets the included comment, so its easier to know more about the property does (like by a third party) if it was game config file!

nikeee:
This draft PR implements the comments feature mentioned in #1211.

This class:
```csharp
class Root
{
    [JsonProperty("test")]
    public float test = 0.42f;
    [JsonProperty("growIncrement", Comment = "Lorem Ipsum")]
    public float GrowIncrement = 0.42f;
}
```
should be serialized to something like this:
```jsonc
{
    "test": 0.42,
    /* Lorem Ipsum */
    "growIncrement": 0.42
}
```

I'm doing a draft pull request to get some opinions and advice on the requested feature and the implementation. Besides adding tests, what can be done better?

This implementation currently emits this JSON (when formatted):
```jsonc
{
  "test": 0.42/* Lorem Ipsum */,
  "growIncrement": 0.42
}
```

How do I prevent the comment being written before the comma of the preceeding property? I suspect the reason being that the AutoClose of the property state is not called before the `WriteComment`. How do I solve this properly?
The comment should also have the same indentation as the property `growIncrement`.

Feature Request: https://stackoverflow.com/questions/9495718
Resolves #1211

### DIFF
--- a/Src/Newtonsoft.Json/Serialization/JsonProperty.cs
+++ b/Src/Newtonsoft.Json/Serialization/JsonProperty.cs
@@ -246,6 +246,12 @@ namespace Newtonsoft.Json.Serialization
         public TypeNameHandling? TypeNameHandling { get; set; }
 
         /// <summary>
+        /// Gets or sets or sets the comment of the property that is written above the property name.
+        /// </summary>
+        /// <value>The contents of the comment.</value>
+        public string? Comment { get; set; }
+
+        /// <summary>
         /// Gets or sets a predicate used to determine whether the property should be serialized.
         /// </summary>
         /// <value>A predicate used to determine whether the property should be serialized.</value>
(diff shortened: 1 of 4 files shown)


Src/Newtonsoft.Json/Serialization/JsonProperty.cs
```diff
@@ -245,6 +245,12 @@ public Required Required
         /// <value>The type name handling.</value>
         public TypeNameHandling? TypeNameHandling { get; set; }
 
+        /// <summary>
+        /// Gets or sets or sets the comment of the property that is written above the property name.
```
almika777, line 249: Description have a reiteration: "or sets or sets"

PR #2337, opened 2020-05-25
Separate StringEnumConverter into abstract base and implementation specific implementation

Brondahl:
TL;DR

Makes it *much* easier to write a Custom StringEnumConverter.
Has no external impact on existing class. (except an Exception message clarification; happy to revert is not wanted)
Tests all pass.

=-=-=-=-=-=-=-=-=-=-=-=-=-=-=

`StringEnumConverter` currently contains 2 separate kinds of code:
 * a Bunch of code that does all of the generalised interaction with the JSON infrastructure; interpretting the JsonReader and tokens, and so-forth
 * a Bunch of code relating to the specific details of how a specific enum value should be converted to a specific string, and _vice versa_.

If one wished to write a Converter that has a different opinion about how to do the actual conversion between a string and an enum, you'd have to copy-paste the two-thirds of the existing class, in order to provide your own implementation of the actual conversion.
_(You'll be shocked to learn that this is something I would like to do :) )_

This PR, introduces an abstract Base class that encapsulates all of the 'boilerplate' and provides 2 abstract methods that are responsible for the actual conversions.
The existing `StringEumConverter` then becomes an implementation of that abstract class, with no changes whatsoever.
But if one wants to create `CustomStringEnumConverter`, then all you have to do is provide implementations of the *actual* string <-> enum conversions ... not worry about all of the other boilerplate.

### DIFF
--- a/Src/Newtonsoft.Json/Converters/StringEnumConverter.cs
+++ b/Src/Newtonsoft.Json/Converters/StringEnumConverter.cs
@@ -27,6 +27,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Reflection;
 using System.Runtime.Serialization;
@@ -44,7 +45,7 @@ namespace Newtonsoft.Json.Converters
     /// <summary>
     /// Converts an <see cref="Enum"/> to and from its name string value.
     /// </summary>
-    public class StringEnumConverter : JsonConverter
+    public class StringEnumConverter : StringEnumConverterBase
     {
         /// <summary>
         /// Gets or sets a value indicating whether the written enum text should be camel case.
@@ -54,7 +55,7 @@ namespace Newtonsoft.Json.Converters
         [Obsolete("StringEnumConverter.CamelCaseText is obsolete. Set StringEnumConverter.NamingStrategy with CamelCaseNamingStrategy instead.")]
         public bool CamelCaseText
         {
-            get => NamingStrategy is CamelCaseNamingStrategy ? true : false;
+            get => (NamingStrategy is CamelCaseNamingStrategy);
             set
             {
                 if (value)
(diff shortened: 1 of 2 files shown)


Brondahl:
@JamesNK I was hoping to make use of the above for a project I'm currently working on. Do you have a feel for when you might be able to look at this PR, and accept it into the codebase?

(It's a pretty small code change, with no functional or NFR effects, and fully covered by existing Unit Tests  :) )

Src/Newtonsoft.Json/Converters/StringEnumConverter.cs
```diff
@@ -54,7 +55,7 @@ public class StringEnumConverter : JsonConverter
         [Obsolete("StringEnumConverter.CamelCaseText is obsolete. Set StringEnumConverter.NamingStrategy with CamelCaseNamingStrategy instead.")]
         public bool CamelCaseText
         {
-            get => NamingStrategy is CamelCaseNamingStrategy ? true : false;
+            get => (NamingStrategy is CamelCaseNamingStrategy);
```
sungam3r, line 58:
```suggestion
            get => NamingStrategy is CamelCaseNamingStrategy;
```

Src/Newtonsoft.Json/Converters/StringEnumConverter.cs
```diff
@@ -161,14 +154,65 @@ public StringEnumConverter(Type namingStrategyType, object[] namingStrategyParam
...
+            Enum e = (Enum)enumObject;
+            return EnumUtils.TryToString(e.GetType(), enumObject, NamingStrategy, out enumStringRepresentation);
+        }
+
+        /// <inheritdoc/>
+        protected override object ConvertStringToEnum(string? inputEnumText, Type destinationPropertyType)
+        {
+            return EnumUtils.ParseEnum(destinationPropertyType, NamingStrategy, inputEnumText!, !AllowIntegerValues);
+        }
+
```
sungam3r, line 176:
```suggestion
```

Brondahl:
I think the check failure is spurious.
The Logs have "Exception: Testhost process exited with error: It was not possible to find any compatible framework version"
and the change that @sungam3r proposed is utterly trivial.

Is there some way to force a re-run?

PR #2388, opened 2020-09-05
Added JsonContract.TypeNameHandling property

TalAloni:
This PR adds JsonContract.TypeNameHandling property.

### Motivation
Having JsonContract.TypeNameHandling property will allow finer control over type name serialization without having to resort to using JsonProperty which is not always possible. (classes outside the developer's control or too many classes)

### Background
During a migration of large project spanning hundreds of DTOs to Newtonsoft.Json, I wished to avoid array types from being serialized with type information, while at the same time have object types serialized with type information when appropriate (TypeNameHandling.Auto for object types, and TypeNameHandling.None for array types).

IMO it is not practical nor desirable to add JsonProperty to hundreds of DTOs, and the suggested property allows finer control at the JsonContract level which I believe is the most desirable approach.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/TypeNameHandlingTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/TypeNameHandlingTests.cs
@@ -767,12 +767,8 @@ namespace Newtonsoft.Json.Tests.Serialization
             string json = JsonConvert.SerializeObject(container, Formatting.Indented,
                 new JsonSerializerSettings
                 {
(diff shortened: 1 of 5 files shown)


TalAloni:
This is still very relevant to me and I would really like to move forward with this.

TalAloni:
This is still very relevant to me and I would really like to move forward with this.

skundlony (approved):

TalAloni:
Thanks @skundlony,
I would really like to move forward with this.

TalAloni:
I would really like to move forward with this.

PR #2396, opened 2020-10-01
Method for serialize to non-strict json

RomanSoloweow:
It's method will be very useful for work with Jquery and for parsing java script objects

### DIFF
--- a/Src/Newtonsoft.Json/JsonConvert.cs
+++ b/Src/Newtonsoft.Json/JsonConvert.cs
@@ -654,6 +654,39 @@ namespace Newtonsoft.Json
             return SerializeObjectInternal(value, type, jsonSerializer);
         }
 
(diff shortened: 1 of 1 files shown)


RomanSoloweow:
@sungam3r

sungam3r:
It seems to me that this PR (as many others) is unlikely to be merged. This is new functionality, not a bugfix. The introduction of new functionality assumes its support in the future, but support for this repository is likely to be phased out in favor of System.Text.Json. Over time, the emphasis has gradually shifted in favor of ystem.Text.Json as a reworked, improved, faster and non-allocating serializer. It will be easier for you to apply this method in your own projects. I think with the release of .NET 5, the System.Text.Json development team will gradually equalize its capabilities with those of Newtonsoft.Json and then this repository will become deprecated officially.

PR #2400, opened 2020-10-06
Simplify `#if` preprocessor statements

damieng:
While the `#if` preprocessor directive is powerful it can play havoc with source analysis tools, syntax highlighters and semantic code analysis when used to build fragments of statements that would otherwise be invalid.

This becomes more important as online analysis tools take off such as GitHub's C# support for semantic code features we've been working on over at https://github.com/tree-sitter/tree-sitter-c-sharp and with the rise of lightweight code editors such as VS Code that do not have a complete compiler at their disposal.

By simply restructuring these`#if` fragments we can make them lightweight-friendly allowing syntax highlighter and analysis tools to deal with them correctly.

There are only three places in the JSON.Net source code where partial-code fragments were in `#if` directives, they are:

1. SerializationErrorHandlingTests - easily solved with a dangling `,` between the sections
2. TextFixtureBase.cs - split the two possible functions out and is now much easier to read as well
3. JsonReader.cs - moved the `#if` outside of the if statement

Here is a screenshot from VSCode showing it struggling with TextFixtureBase:

image


TylerBrinkley (approved):

damieng:
Guessing this isn't going to be merged, closing it out.

Issue #2406, opened 2020-10-15
Cannot create and populate list type System.Data.SqlClient.SqlErrorCollection. Path '$values'


Danthar:
We have been having troubles with deserializing SqlExceptions using newtonsoft.net Version 11.0.2 On .Net 4.6.2 (Full)
We didn't have these issues before when we where on an earlier version of newtonsoft. (10.x)

I have been searching around for how to fix this. But all i can find are solutions in which you decorate the offending property with an attribute. Which obviously is not a option here. Upgrading our .Net version is an option, but is non trivial for us. So im trying to find out if i have other options. 
Upgrading my TestCase to .net 4.8 also did not help.

I believe this to be a Bug in Newtonsoft since version 11.x

Also have a SO question running. https://stackoverflow.com/questions/64308864/cannot-create-and-populate-list-type-system-data-sqlclient-sqlerrorcollection-p

Expected behavior.

Deserializing an System.Data.SqlClient.SqlException from the .net SDK (not .net core) should work.

Actual behavior;

![image](https://user-images.githubusercontent.com/986519/96130756-9f2ae980-0ef8-11eb-9e08-a0a4c4ec5761.png)

Danthar:
For convenience here is a gist with a reproduction case. https://gist.github.com/Danthar/536330ea6ec94874b16a2880769f936a

Danthar:
So i added a TestCase PR, illustrating the issue. I would not mind submitting a PR to fix the issue. But i wouldn't know where to look.

Danthar:
@JamesNK Anyhing i can do help facilitate to get some eyeballs on this issue?

JamesNK:
It only implements `ICollection`, which doesn't have an Add method, and there is no constructor for specifying values. There is no way to deserialize this type.

Danthar:
And yet. It worked in older versions of json.net.

Op di 27 okt. 2020 18:57 schreef James Newton-King <notifications@github.com

Danthar:
@JamesNK Is there anyway i can work around this perhaps? I don't know which change between 10.x and 11.x caused this issue. But is there perhaps a way i can have it handle the SqlException type differently? By for example, having it ignore the SqlErrorCollection ?

johanolofsson:
It seems that as soon as a class is implementing IEnumerable<T>, IReadOnlyList<T> and perhaps others, NewtonsoftJson assumes it should be deserialized as some kind of collection. 

[...]

PR #2412, opened 2020-10-21
Sql exception issue test

Danthar:
Testcase for issue: #2406


Danthar:
Fails due to what seems to be build agent stuff.
Compiles and runs test locally just fine.

PR #2422, opened 2020-11-12
Add [MaybeNull] for return of generic JsonConvert.ReadJson

michsco:
`JsonConverter<T>` has inconsistent nullable reference attribution between the generic and non-generic versions. As a result, custom `JsonConverter<T>` implementations cannot return null values for reference types.

As part of the change, I had to update the .NET Core SDK that is used so that the Azure DevOps pipeline succeeds.


michsco:
Unfortunately, the check appears to fail due to .NET Core 2.2 not being available on the build machine.

JamesNK:
Thanks for the PR.

I've fixed the build issues along with your original change here - https://github.com/JamesNK/Newtonsoft.Json/pull/2424

PR #2438, opened 2020-12-14
Missing nullability annotation

njy:
In the param description it says _"If this is null, default serialization settings will be used"_ but the param is not marked as nullable.


JamesNK (approved):
Good catch. Thanks for the PR!

sungam3r:
Is this change backwards incompatible for already compiled code? The parameter was a ref type, it became a struct.

PR #2454, opened 2021-01-12
jpath regex timeout support

serpentfabric:
from https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices

> When using System.Text.RegularExpressions to process untrusted input, pass a timeout...you should use the timeout feature to ensure that excessive backtracking does not severely degrade regular expression performance...[or causes] a Denial-of-Service attack.

unfortunately timeouts aren't supported in System.Text.RegularExpressions until ~net4.5 or so... so i only really added support in for the frameworks that can accept a timeout.

also because SelectTokens runs the regex against every token that match some path, even if each token completes within some timeout, the sum of regex calls to all tokens may exceed some allotment. therefore i've added a global timeout parameter to SelectTokens to ensure the call completes roughly within the expected timeout.

also, i added support for the compiled regex option/switch, since that's the main use case we have and the speedup is important to us.


JamesNK:
Thanks for the PR. This is something I've been meaning to do for a while.

I've modified your PR with some changes:
* Put the timeout on a new settings object. Will allow for potential future settings.
* Remove global timeout. This is kind of a weird feature. If someone wants to do this then they can implement it themselves when they foreach over the results.

New PR - https://github.com/JamesNK/Newtonsoft.Json/pull/2456

PR #2470, opened 2021-02-18
Ensure that untracked sources are embedded in the PDB

clairernovotny:
Required for reproducible builds.


JamesNK:
Replaced with https://github.com/JamesNK/Newtonsoft.Json/pull/2470. Puts the property in Newtonsoft.Json.csproj.

Issue #2480, opened 2021-02-25
For netfx20 edition cannot format DateTimeOffset correctly


ahdung:
Json.Net version : 8.0.3, 12.0.3 **both for .netfx20**
Project target framework: .netfx2.0, .netfx4.8

I want the right side result, please help.

![微信截图_20210225114154](https://user-images.githubusercontent.com/12094247/109099974-eeaedb00-775e-11eb-8da2-caaf5623b116.png)

ahdung:
for .netfx35+ no problem

ahdung:
I saw https://github.com/ahdung/Newtonsoft.Json/blob/b6dc05be5a0f4808f06ec430f3bb59b24d3fbc3e/Src/Newtonsoft.Json/Newtonsoft.Json.csproj#L59-L62 not defined `HAVE_DATE_TIME_OFFSET`, but .netfx20 have `DateTimeOffset`.

JamesNK:
net20 doesn't have DateOffset because it was added later in a service pack. I don't want to add it now because it is potentially a breaking change.

I recommend either don't use net20, or don't use DateTimeOffset.

ahdung:
I see. but i think the number of people using rtm less than using service pack, there have a reference is the results of google ".net framework 2.0" are almost about 2.0SP2, so maybe there is another choice for rtm users: staying at current json.net version, or update .net with sp.

PR #2481, opened 2021-02-25
Add DateTimeOffset support for NetFx20 edition

ahdung:
`.netfx20` obviously have `DateTimeOffset` class, but the netfx20 edition doesn't support it, so I did this little changing.


ahdung:
#2480

Issue #2488, opened 2021-03-09
Target type not exposed in DefaultContractResolver.CreateProperty method when property is declared in a base class


YuKitsune:
When inheriting from the `DefaultContractResolver` and overriding the `CreateProperty(MemberInfo, MemberSerialization)` method, the `MemberInfo` made available provides the wrong `memberInfo.ReflectedType` when the member is declared on a base class.

I've narrowed the issue down to this block `ReflectionUtilities.GetProperties(Type, BindingFlags)`:
https://github.com/JamesNK/Newtonsoft.Json/blob/b6dc05be5a0f4808f06ec430f3bb59b24d3fbc3e/Src/Newtonsoft.Json/Utilities/ReflectionUtils.cs#L945-L954

I totally get why this is necessary, but this results in the `MemberInfo.ReflectedType` being incorrect later down the track, so when overriding methods such as `DefaultContractResolver.CreateProperty(MemberInfo, MemberSerialization)`, you can't figure out what the target `Type` is being resolved.

### Expected behavior
Without having looked at the code linked above, I would've assumed that the `ReflectedType` is the `Type` which would've been passed into the `Resolve(Type)` method.

### Actual behavior
The `ReflectedType` is the same as the `DeclaringType`, and no reference to the target `Type` is exposed to the `CreateProperty(MemberInfo, MemberSerialization)` method, leaving you unable to accurately determine the `Type` being resolved.

### Steps to reproduce

```csharp
public class Vehicle
{
  public string Registration { get; set; }
}

public class Car
{
}

public class MyContractResolver : DefaultContractResolver
{
  protected override JsonProperty CreateProperty(
    MemberInfo member,
    MemberSerialization memberSerialization)
  {
    // When the `MemberInfo` for the `Registration` property has been passed in,
    //  the `DeclaringType` and the `ReflectedType` will be `Vehicle`, even if a `Car` instance was provided.
    Console.WriteLine($"DeclaringType: {member.DeclaringType}");
    Console.WriteLine($"ReflectedType: {member.ReflectedType}");
  }
}
```

## Temporary Solutions
A temporary solution I've found is to override the `CreateProperties` method (which still has a reference to the original type as a parameter), then create a separate method which takes the `Type` as a parameter, and behaves similarly to the original `CreateProperty` method.
This might not work in every scenario, and could end up being hard to maintain, since the base `CreateProperties` method isn't as straightforward as my below example.
```csharp
[...]

PR #2489, opened 2021-03-09
Added `CreateProperty` method exposing the target `Type` to the `DefaultContractResolver`

YuKitsune:
Attempts to address #2488

## Summary
When inheriting from the `DefaultContractResolver` and overriding the `CreateProperty(MemberInfo, MemberSerialization)` method, the `MemberInfo` made available provides the wrong `memberInfo.ReflectedType` when the member is declared on a base class.

This is because private getters and setters are inaccessible unless the property was gotten from the base type:
https://github.com/JamesNK/Newtonsoft.Json/blob/b6dc05be5a0f4808f06ec430f3bb59b24d3fbc3e/Src/Newtonsoft.Json/Utilities/ReflectionUtils.cs#L945-L954

This unfortunately has a side-affect where the target `Type` being resolved can be lost when creating the `JsonProperty` for a property defined on a base class, leaving the `CreateProperty` method with no context of what `Type` is being resolved.

While this isn't an issue internally, it could pose an issue for anyone overriding this method who needs to know the type being resolved (I've run into this issue and have implemented a temporary solution as shown in the linked issue).

This PR adds an extra method between `CreateProperties` and `CreateProperty`, which exposes the target `Type` as a parameter. 

### Before
`CreateProeprties(Type, MemberSerialization)` -> `CreateProperty(MemberInfo, MemberSerialization)`

### After
`CreateProeprties(Type, MemberSerialization)` -> `CreateProperty(Type, MemberInfo, MemberSerialization)` -> `CreateProperty(MemberInfo, MemberSerialization)`

This intermediate method serves no purpose but to expose the `Type` if necessary.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/ContractResolverTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/ContractResolverTests.cs
@@ -797,6 +797,18 @@ namespace Newtonsoft.Json.Tests.Serialization
         }
 
         [Test]
+        public void DefaultContractResolverExposesCorrectTypeToCreatePropertyMethod()
+        {
+            VehicleContractResolver resolver = new VehicleContractResolver();
+
+            JsonObjectContract trainContract = (JsonObjectContract)resolver.ResolveContract(typeof(Train));            
+            JsonObjectContract aircraftContract = (JsonObjectContract)resolver.ResolveContract(typeof(Aircraft));            
+
+            Assert.IsTrue(trainContract.Properties[nameof(Vehicle.Registration)].Ignored);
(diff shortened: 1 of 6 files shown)


YuKitsune:
Having thought about it for a bit longer, think the changes here could be confusing to people attempting to override the `CreateProperty()` method, as the relationships between `CreateProperty(Type, MemberInfo, MemberSerialization)` and `CreateProperty(MemberInfo, MemberSerialization)` is not clear without first looking at the source code.

Given that this isn't a _huge_ issue, I think an appropriate workaround is to override the `CreateProperties(Type, MemberSerialization)`.

Issue #2501, opened 2021-03-24
MaxDepth default value of `null` or `100` is not respected


Smaug123:
### Source/destination types

Code expressed in F#, sorry. Serialisation is fine; deserialisation is not.

```fsharp
type Foo =
| Foo of Foo
| Bar of int
```

### Source/destination JSON

This is very, very long - sorry.

```javascript
{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Foo","Fields":[{"Case":"Bar","Fields":[3]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}]}
```

### Expected behavior

As documented, the `MaxDepth` property being null should mean there is no max depth, and we should get a deserialisation.

### Actual behavior

On Newtonsoft.Json 13.0.1 only (I checked 12.0.3 and it was fine), the following exception is thrown:
```
[...]

PR #2504, opened 2021-03-25
Fix MaxDepth problems introduced between 12.0.3 and 13.0.1

mcon:
Fixes https://github.com/JamesNK/Newtonsoft.Json/issues/2501 - have verified that the example provided in F# fails with no change and is resolved with the change.

Between 12.0.3 and 13.0.1 looks like `_maxDepth = 64;` was set in the `JsonReader` constructor - any inheritors which use this default constructor end up with a max depth of 64 as a result - from the existing tests on max depth, it seems that this setting is not actually required in order to ensure the max depth is respected.

I've also updated a few doc comments which erroneously claim the max depth as 128.

Update: as much as the initial fix (https://github.com/JamesNK/Newtonsoft.Json/pull/2462) prevents a DOS attack, it does create other problems, such as 2501 - max depth is a bit tricky throughout the library, one option would be to make it mandatory on the `JsonReader` perhaps.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
@@ -7996,14 +7996,10 @@ This is just junk, though.";
         public void SetMaxDepth_DepthExceeded()
         {
             JsonTextReader reader = new JsonTextReader(new StringReader("[[['text']]]"));
-            Assert.AreEqual(64, reader.MaxDepth);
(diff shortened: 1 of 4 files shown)


Smaug123:
Suggest adding MaxDepth tests to verify that you can extend MaxDepth to be bigger than the default?

JamesNK:
Thanks for looking at this.

I've fixed this in a different way here https://github.com/JamesNK/Newtonsoft.Json/pull/2505

Issue #2529, opened 2021-05-14
Exception deserializing JArray on JToken-typed property pre-initialized with JValue


FstTesla:
### Destination type

```csharp
class Something
{
    public JToken Foo { get; set; } = JValue.CreateNull();
}
```

### Source JSON

```javascript
{ "foo": [] }
```

### Expected behavior

Successfully deserialized a `Something` object with property `Foo` equal to an empty `JArray`.

### Actual behavior

Exception `JsonSerializationException`: Deserialized JSON type 'Newtonsoft.Json.Linq.JArray' is not compatible with expected type 'Newtonsoft.Json.Linq.JValue'. 

### Steps to reproduce

```csharp
JsonConvert.DeserializeObject<Something>("{ \"foo\": [] }");
```

### Observations

Issue is new in version 13.0.1 and did not affect version 12.0.3.
Probably this is an unwanted side effect of fixing issue #2484.

The error comes from the difference between the types of the pre-initialized value of the property (`null` :: `JValue`) and the one provided in the input (`[]` :: `JArray`).
Therefore any of the following changes does not raise the exception:
- removing the property initializer of `Foo`;
- replacing the property initializer of `Foo` with an instance of `JArray` (even a non-empty one);
- removing the `foo` property in the input;
- replacing the `foo` property in the input with a `JValue`-compatible JSON snippet.

PR #2530, opened 2021-05-14
Fix deserializing mistmatched JToken types in properties

JamesNK:
Fixes https://github.com/JamesNK/Newtonsoft.Json/issues/2529


ShaydeNofziger:
@JamesNK Hey James - thanks for all your hard work on this and the Newtonsoft.Json library in general.

Do you know if/when a new release version will be cut for this fix? Running into the issue now while updating from 12.0.3 to 13.0.1 and would prefer the fix vs the workarounds mentioned in #2529

ejball:
@JamesNK Will this regression fix ever be deployed? It is especially critical now that older versions of the package are all alerted as potential security vulnerabilities.

JamesNK:
It's on NuGet

PR #2556, opened 2021-07-12
Epsilon for double value comparison can now be set

santiagoIT:
This commit allows the epsilon value for double comparison to be set.
The previously hard coded value is too tight for many of our use cases.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Linq/JTokenTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Linq/JTokenTests.cs
@@ -1221,6 +1221,29 @@ namespace Newtonsoft.Json.Tests.Linq
 
             Assert.IsTrue(v1.DeepEquals(v2));
         }
+        
+        [Test]
+        public void DoubleDeepEqualsWithCustomEpsilon()
+        {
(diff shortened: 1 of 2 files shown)


TylerBrinkley:
@santiagoIT `MathUtils` is an internal class and thus cannot be used directly outside of Newtonsoft.Json itself and it's unit test project thus this PR will have no value to you if implemented.

santiagoIT:
@TylerBrinkley Thx! Did not notice that the MathUtils class was internal. I added a public class JsonCompareOptions that holds the Epsilon value.

JamesNK:
Static configuration isn't a good idea. What happens if two libraries set it to different values?

The right way to do this would be to have a settings object that is passed to DeepEquals and then it gets passed to all its children. Like https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Linq/JsonSelectSettings.cs

That is a much more invasive change and might not be one I want to make.

santiagoIT:
@JamesNK 
I have added a JsonCompareSettings class.
This is indeed the better approach.
Thx!

PR #2557, opened 2021-07-13
Use latest SDK for reproducible builds

clairernovotny:
Needs the 16.10 or later MSBuild for the proper metadata.


Src/global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "5.0.200"
+    "version": "5.0.3 00"
```
campersau, line 3:
```suggestion
    "version": "5.0.300"
```

JamesNK:
Thanks!

PR #2584, opened 2021-09-10
Add ShallowCopy method for settings

MatthewSteeples:
Start of the fix for https://github.com/dotnet/aspnetcore/issues/29532

There are some hidden side effects for 2 of the properties (MaxDepth and DateFormatString). If you were to create a new JsonSerializerSettings object and copy all of the properties across the behaviour would be different (MaxDepth would be null on your source but as the "set" bool is false you'd have a value of 64. If you set your destination's property to be null then the value will be unlimited

This ShallowCopy method will create a copy of the object but at a field level (where possible) to avoid these side effects

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
@@ -8081,7 +8081,7 @@ This is just junk, though.";
         {
             var settings = new JsonSerializerSettings();
 
-            var clone = settings.ShallowCopy();
+            var clone = new JsonSerializerSettings(settings);
 
             Assert.AreEqual(false, settings._dateFormatStringSet);
             Assert.AreEqual(false, clone._dateFormatStringSet);
--- a/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
@@ -8079,12 +8079,109 @@ This is just junk, though.";
         [Test]
         public void ShallowCopy_DoesntSetDateFormatStringSet()
         {
+            var propertyNames = typeof(JsonSerializerSettings).GetProperties().Select(property => property.Name).ToList();
+
             var settings = new JsonSerializerSettings();
 
             var clone = new JsonSerializerSettings(settings);
 
-            Assert.AreEqual(false, settings._dateFormatStringSet);
-            Assert.AreEqual(false, clone._dateFormatStringSet);
+            Assert.AreEqual(settings.DateFormatString, clone.DateFormatString);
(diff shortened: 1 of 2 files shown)


TylerBrinkley:
I've wanted something like this also but was thinking of a copy constructor instead. What do you think?

MatthewSteeples:
That's probably a better idea actually. More discoverable and clearer on the intent. I'll make the changes on this branch

Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
```diff
@@ -8076,6 +8076,17 @@ public void SetMaxDepth_MaxValueDepthNotExceeded()
...
 
+        [Test]
+        public void ShallowCopy_DoesntSetDateFormatStringSet()
+        {
+            var settings = new JsonSerializerSettings();
+
+            var clone = new JsonSerializerSettings(settings);
+
+            Assert.AreEqual(false, settings._dateFormatStringSet);
+            Assert.AreEqual(false, clone._dateFormatStringSet);
```
halter73, line 8087: Can we use reflection to verify all fields/properties are copied? Similar to [this CloneSslOptionsClonesAllProperties test](https://github.com/dotnet/aspnetcore/blob/8b30d862de6c9146f466061d51aa3f1414ee2337/src/Servers/Kestrel/Core/test/SniOptionsSelectorTests.cs#L658-L749) in Kestrel. The basic idea is to use GetFields()/GetProperties() to get every property and then remove each property from the list after verifying that a non-default value was copied. Then at the end of the test you verify the list is empty. The test will need updated any time a property is added, but at least the test should fail if a property is added without being copied by the constructor.

MatthewSteeples: @halter73 Done (better late than never... :))

MatthewSteeples:
Build looks like it's failing because the version of NET6 being downloaded (6.0.200) is too new for global.json to use

As a workaround for this, it might be worth having a `UseDotNet` task that reads the value from global.json (so that it's always up to date)

Src/Newtonsoft.Json/JsonSerializerSettings.cs
```diff
@@ -452,5 +453,43 @@ public JsonSerializerSettings()
         {
             Converters = new List<JsonConverter>();
         }
+
+        /// <summary>
+        /// Create a copy of the passed in <see cref="JsonSerializer"/>.
```
JamesNK, line 458:
```suggestion
        /// Initializes a new instance of the <see cref="JsonSerializerSettings"/> class using values copied from the passed in <see cref="JsonSerializerSettings"/>.
```

JamesNK:
I've merged this in PR https://github.com/JamesNK/Newtonsoft.Json/pull/2676 because I couldn't edit this one.

Thanks for the changes.

Issue #2591, opened 2021-10-04
Type interpretation error for arrays of more than 2 dimensions in a serialized class.


Perpete:
Hello,

In a VB project with .Net 5 in WPF using Newtonsoft.json (13.0.1) I noticed the problem below.

In a class, I'm using type name handling for an Object property that contains an array.
 (**<Newtonsoft.Json.JsonProperty (TypeNameHandling: = Newtonsoft.Json.TypeNameHandling.All)>**)

When I deserialize this class after serialization, I get an error :
**Unexpected character encountered while parsing value: [. Path '[2] .Value. $ Values ​​[0]', line 1, position 340**.

While researching, I noticed an issue with arrays that have dimensions greater than 2.
The **$type ":" System.xxxxx [,], System.Private.CoreLib** flag always keeps a comma between square brackets for arrays with more than 2 dimensions.
For 3 dimensions, we should have **$type ":" System.xxxxx [,,], System.Private.CoreLib**, which is not the case.

It is this misinterpretation of the type that causes the error.
If we add the missing commas in a serialized file, the deserialization of the file goes correctly.

3D array with wrong type in file
![2021-10-04 12_25_15-IOCenterOPC - Microsoft Visual Studio](https://user-images.githubusercontent.com/66659162/135836060-ad0bef9e-413a-4493-828d-c2dc226c5a4a.png)

After correction the deserialization is correct
![2021-10-04 12_35_07-IOCenterOPC - Microsoft Visual Studio](https://user-images.githubusercontent.com/66659162/135837121-36c1b4ac-ca75-4149-b67c-143336e75861.png)

Here are screenshots of a test program.

Serialization and deserialization with 2 Array (dimension <= 2)
![2021-10-04 12_04_27-TestJson (Exécution) - Microsoft Visual Studio](https://user-images.githubusercontent.com/66659162/135832961-073489ed-6d07-4a2c-8cfc-52ec77bca4a5.png)

Serialization and deserialization with 4 Array (dimension> 2)
![2021-10-04 12_08_30-TestJson (Exécution) - Microsoft Visual Studio](https://user-images.githubusercontent.com/66659162/135833416-585559b3-240c-4a50-8874-8cf2a2c4ee2a.png)

Here is my code in vb for the test.

`Imports Newtonsoft.Json

Class MainWindow

	Dim output As String

	Private Sub btDeSerialize_Click(sender As Object, e As RoutedEventArgs) Handles btDeSerialize.Click

[...]

PR #2605, opened 2021-10-28
Correction for arrays of more than 2 dimensions with $type

Perpete:
When using the type interpretation for arrays of more than 2 dimensions, on deserialization, an error is encountered.
There is a problem in the switch case that does not allow adding more than one comma between the brackets of System.xxxx [].
View issue #2591


JamesNK:
Hi! Thanks for the fix.

I've made some small changes and added some tests. This PR is replaced by https://github.com/JamesNK/Newtonsoft.Json/pull/2610 which builds on your changes.

PR #2614, opened 2021-11-17
Add support to pre-epoch Unix time

### LINKED ISSUE #2292: UnixDateTimeConverter throws exception if date is before 1970.
UnixDateTimeConverter is only able to convert dates after 1970, while there is apparently no reason to throw exception for dates previous to 1970. 
In several algorithm explanations, dates before 1970 can be converted to negative values in unix timestamp.

MaxKot:
Fixes #2292

The existing implementation contains code specifically to disallow dates before Unix epoch. It can be assumed that some users of the library depend on the converter throwing an exception if such date is serialized or deserialized. To maintain backwards compatibility, a constructor parameter and a property are introduced to explicitly enable the new behavior. A static default setting for allowing pre-epoch dates was also introduced to help library users to switch to the new behavior with minimal changes of the client code.

### Rationale

Many implementations of Unix time use signed integer values with negative values representing dates before Unix epoch. Examples of such implementations are some C standard libraries and DateTimeOffset.ToUnixTimeSeconds and DateTimeOffset.FromUnixTimeSeconds methods in .NET basic class library. JavaScript built-in Date.prototype.getTime() method type uses milliseconds instead of seconds but it still handles pre-epoch dates with negative numbers. In fact the only implementation disallowing pre-epoch dates I could find was MySQL. Hence negative timestamps are likely to appear in JSON documents in any problem domain that handles dates before 1970-01-01 (e.g. people birthdays).

### DIFF
--- a/Src/Newtonsoft.Json/Converters/UnixDateTimeConverter.cs
+++ b/Src/Newtonsoft.Json/Converters/UnixDateTimeConverter.cs
@@ -39,6 +39,20 @@ namespace Newtonsoft.Json.Converters
         private bool _allowPreEpoch;
 
         /// <summary>
+        /// Gets or sets the default value of the <see cref="AllowPreEpoch"/>
+        /// property for new instances of the <see cref="UnixDateTimeConverter"/>
+        /// class.
+        /// </summary>
+        /// <value>
+        /// <c>true</c> to allow converting dates before Unix epoch to and from
+        /// JSON;
+        /// <c>false</c> to throw an exception when a date being converted to or
+        /// from JSON occurred before Unix epoch.
+        /// The default is <c>false</c>.
+        /// </value>
+        public static bool DefaultAllowPreEpoch { get; set; } = false;
+
+        /// <summary>
         /// Gets or sets a value indicating whether the dates before Unix epoch
         /// should converted to and from JSON.
         /// </summary>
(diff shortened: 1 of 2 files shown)


MaxKot:
@JamesNK could you please take look at this?

Src/Newtonsoft.Json/Converters/UnixDateTimeConverter.cs
```diff
@@ -36,6 +36,64 @@ public class UnixDateTimeConverter : DateTimeConverterBase
...
+        /// class.
+        /// </summary>
+        /// <value>
+        /// <c>true</c> to allow converting dates before Unix epoch to and from
+        /// JSON;
+        /// <c>false</c> to throw an exception when a date being converted to or
+        /// from JSON occurred before Unix epoch.
+        /// The default is <c>false</c>.
+        /// </value>
+        public static bool DefaultAllowPreEpoch { get; set; } = false;
```
JamesNK, line 53: Static configuration like this isn't desirable. It is global and different apps/libraries can override each other. Please remove.

JamesNK (changes requested):
This looks fine apart from static configuration.

MaxKot:
Thanks for the quick reply!

> Static configuration like this isn't desirable. It is global and different apps/libraries can override each other. Please remove.

Done.

BrandonDusseau:
@JamesNK would you mind taking another look at this? Having the ability to parse pre-1970 dates from epoch time would be immensely helpful. I would even go as far as to enable it without an option.

JamesNK (approved):

JamesNK:
Thanks!

PR #2616, opened 2021-11-20
Fix ContractResolver to also consider non-public ShouldSerialize methods

matkoch:
The [documentation](https://www.newtonsoft.com/json/help/html/ConditionalProperties.htm) doesn't say whether this works with public or non-public methods, but it does claim that it works _similar_ to [XmlSerializer](http://msdn.microsoft.com/en-us/library/53b8022e.aspx), which in fact accepts private methods.

I think this would be good to allow not spoiling the public API of the types.

### DIFF
--- a/Src/Newtonsoft.Json/Serialization/DefaultContractResolver.cs
+++ b/Src/Newtonsoft.Json/Serialization/DefaultContractResolver.cs
@@ -1635,7 +1635,7 @@ namespace Newtonsoft.Json.Serialization
 
         private Predicate<object>? CreateShouldSerializeTest(MemberInfo member)
         {
-            MethodInfo shouldSerializeMethod = member.DeclaringType.GetMethod(JsonTypeReflector.ShouldSerializePrefix + member.Name, ReflectionUtils.EmptyTypes);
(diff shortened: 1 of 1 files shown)


sungam3r:
Should be treated as a breaking change in behavior.

matkoch:
@sungam3r could you elaborate why? Are you assuming people might have methods declared exactly according to that naming pattern but private?

sungam3r:
Sort of it. In fact, it does not matter to consider these changes as breaking or not. The likelihood that this PR will be accepted is close to zero.

JamesNK:
> @sungam3r could you elaborate why? Are you assuming people might have methods declared exactly according to that naming pattern but private?

Yes. Someone will upgrade to a newer version of Newtonsoft.Json (or have a dependency do it for them) and then properties will stop being serialized.

ShouldSerialize is a feature I'm not very fond of because when it happens accidentally, it is surprising and unexpected. Having private methods do it would make the issue worse.

matkoch:
@JamesNK  I’m confident you know better what people run into the most. My thinking here was that changing accessibility without compile issues but with changing behavior is a bit of a red flag, particularly for analyzers. Also it’s not “stop being serialized” but whatever the method returns.

Would you see any possibility to have another `JsonPropertyAttribute` property that can reference any member by name including non-public? That would also solve the “unused member” issue.

Re “not fond”: maybe you have a better suggestion. What I’m trying to do is to prevent serialization of empty collections but at the type declaration. The code is generated, so I wouldn’t have an issue with mentioned magic methods. Any other solution I’ve tried is either more complex or requires custom serialization settings.

PR #2618, opened 2021-12-01
Enabling NuGet package validation

joperezr:
This PR adds the [new NuGet package validation](https://docs.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview) into Newtonsoft.Json's package build. This will ensure that checks are made on each PR to ensure compatibility within the package, as well as against previously shipped package (this PR will add comparison against 13.0.2 which is the latest stable version). As part of this PR, I'm also baselining the existing compatibility issues. @JamesNK could you please take a look at those to see if all of them are by design? If so, we can live them suppressed via the baseline, if not, then we can address the issues and remove them from the baseline.

FYI: @ericstj @safern


Src/Newtonsoft.Json/Newtonsoft.Json.csproj
```diff
@@ -33,6 +33,13 @@
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
     <CodeAnalysisRuleset>Newtonsoft.Json.ruleset</CodeAnalysisRuleset>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
+	<!--Package Validation-->
+	<EnablePackageValidation>true</EnablePackageValidation>
+	<PackageValidationBaselineVersion>13.0.1</PackageValidationBaselineVersion>
+	<!--CP0003 will be logged when the assembly identity between baseline package and current don't match.
+	Disable that warning since local builds do not strong-name sign the assembly, causing a mistmatch when comparing
+	the previously shipped assemblies and current.-->
+	<NoWarn>$(NoWarn);CP0003</NoWarn>
```
safern, line 42: Nit, indentation is weird.

Src/Newtonsoft.Json/Newtonsoft.Json.csproj
```diff
@@ -33,6 +33,13 @@
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
     <CodeAnalysisRuleset>Newtonsoft.Json.ruleset</CodeAnalysisRuleset>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <!--Package Validation-->
+    <EnablePackageValidation>true</EnablePackageValidation>
+    <PackageValidationBaselineVersion>13.0.1</PackageValidationBaselineVersion>
+    <!--CP0003 will be logged when the assembly identity between baseline package and current don't match.
+    Disable that warning since local builds do not strong-name sign the assembly, causing a mistmatch when comparing
+    the previously shipped assemblies and current.-->
+    <NoWarn>$(NoWarn);CP0003</NoWarn>
```
ericstj, line 42: Can we apply a condition here so that it's not permitted during an official build?  Maybe `Condition='$(SignAssembly)' != 'true'"`

joperezr: Oh good suggestion, sure thing, will do.

Src/Newtonsoft.Json/CompatibilitySuppressions.xml
```diff
@@ -0,0 +1,104 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Suppressions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
```
joperezr, line 2: @JamesNK Thanks for fixing the build by updating VS. I added some comments to explain he breaking changes that were found in the package as it is today. Could you please take a look and confirm all of them are by design?

Src/Newtonsoft.Json/Newtonsoft.Json.csproj
safern, line 42: It seems like the condition for this hasn't been added.

joperezr: Oops looks like I never pushed the change.

Src/Newtonsoft.Json/CompatibilitySuppressions.xml
JamesNK: They look correct.

JamesNK:
Thanks!

PR #2624, opened 2021-12-15
[FixBug] Error converting value "" to Int64

SpringHgui:
# Solve the following problems
When property type is int32, The following code OK
```c#
    public class Program
    {
        public class TestModel
        {
            public int from { get; set; }
        }

        {
            var json = "{ \"from\": \"\"}";
            var ss = JsonConvert.DeserializeObject<TestModel>(json, new JsonSerializerSettings { NullValueHandling = NullValueHandling.Ignore });
            Console.WriteLine(ss);
        }
    }
```

When property type is int64, The following code throw exception
```c#
    public class Program
    {
        public class TestModel
        {
            public long from { get; set; }
        }

        {
            var json = "{ \"from\": \"\"}";
            var ss = JsonConvert.DeserializeObject<TestModel>(json, new JsonSerializerSettings { NullValueHandling = NullValueHandling.Ignore });
            Console.WriteLine(ss);
        }
    }
```

set `NullValueHandling = NullValueHandling.Ignore` they not have the same effect

### DIFF
--- a/Src/Newtonsoft.Json/JsonReader.cs
+++ b/Src/Newtonsoft.Json/JsonReader.cs
@@ -396,6 +396,57 @@ namespace Newtonsoft.Json
         public abstract bool Read();
 
         /// <summary>
+        /// Reads the next JSON token from the source as a <see cref="Nullable{T}"/> of <see cref="Int64"/>.
+        /// </summary>
+        /// <returns>A <see cref="Nullable{T}"/> of <see cref="Int64"/>. This method will return <c>null</c> at the end of an array.</returns>
+        public virtual long? ReadAsInt64()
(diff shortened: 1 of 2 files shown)


happysammy:
I'm experencing the same issue, Deserialization works fine for properties of type int, decimal, or string, but it throws an exception for properties of type long. The library version is '13.0.3', and the exception is as follows:
```
CLR/Newtonsoft.Json.JsonSerializationException
An exception of type 'Newtonsoft.Json.JsonSerializationException' occurred in Newtonsoft.Json.dll but was not handled in user code: 'Error converting value "" to type 'System.Int64'. 

```

PR #2625, opened 2021-12-16
Added Validation to String Buffer ToString Method

ilkererkek:
I added validation to StringBuilder.ToString function to prevent exceptions. 
If the parameters don't validate function now returns the ToString function with no parameters.


jrmoreno1:
Negative length seems likely to mean "get length characters before start"....

PR #2636, opened 2022-01-11
Make JPath public to allow more basic operations on JSONPath expressi…

rh78:
Making JPath public has a lot of benefits. Users of the Newtonsoft JSON library can then parse JSON expressions and validate them, without having to actually supply a data object (which sometimes is not available at validation or preparation).

This functionality is currently missing except using SelectTokens, which is a very high level approach.


rh78:
This issue is not implemented/merged. This seems to be an error.

PR #2639, opened 2022-01-14
implement box avoidance

### LINKED ISSUE #2638: Memory usage suggestion : reduce unnecessary boxes
Currently, if you deserialize a model with lots of, say, boolean values in it: you get a *lot* of boxed booleans on the heap, from multiple sources, but mostly:

- `JsonTextReader.SetToken`
- `new JValue(object, [JTokenType])`

Suggestion:

- create `static readonly` boxed objects for a range of likely primitive values
- use the shared boxed copies in preference to vanilla boxing
- avoid naïve and unnecessary `new JValue(object)` (side benefit: avoid some extra type testing)

This would significantly help (based on real testing with dotMemory etc) both the temporary gen-zero allocations of in-flight objects *and* the cost of retained `JToken` etc trees.

Happy to submit a PR if useful.

mgravell:
fix #2638

1. adds internal `BoxedPrimitives` helper that holds boxed instances for common primitive values and has helper APIs
2. use `BoxedPrimitives` in preference to vanilla boxing in `JValue` and `JsonTextReader`
3. avoid `new JValue(object)` in key paths

Disclosure: if someone uses `Unsafe` methods to mutate the inside of one of these boxes: bad things. But that is self-inflicted.

### DIFF
--- a/Src/Newtonsoft.Json/Utilities/BoxedPrimitives.cs
+++ b/Src/Newtonsoft.Json/Utilities/BoxedPrimitives.cs
@@ -112,11 +112,5 @@ namespace Newtonsoft.Json.Utilities
         internal static readonly object DoublePositiveInfinity = double.PositiveInfinity;
         internal static readonly object DoubleNegativeInfinity = double.NegativeInfinity;
         internal static readonly object DoubleZero = (double)0;
-
-#if HAVE_BIG_INTEGER
-        internal static object Get(System.Numerics.BigInteger value) => value == System.Numerics.BigInteger.Zero ? BigIntegerZero : value;
-
-        private static readonly object BigIntegerZero = System.Numerics.BigInteger.Zero;
-#endif
     }
 }
--- a/Src/Newtonsoft.Json/Utilities/BoxedPrimitives.cs
+++ b/Src/Newtonsoft.Json/Utilities/BoxedPrimitives.cs
@@ -1,0 +1,83 @@
+﻿#region License
+// Copyright (c) 2022 James Newton-King
+//
+// Permission is hereby granted, free of charge, to any person
+// obtaining a copy of this software and associated documentation
+// files (the "Software"), to deal in the Software without
+// restriction, including without limitation the rights to use,
+// copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following
+// conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+// OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+// HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+// WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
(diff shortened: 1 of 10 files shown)


Src/Newtonsoft.Json/Utilities/BoxedPrimitives.cs
```diff
@@ -0,0 +1,83 @@
...
+                case 0:
+                case 1:
+                case 2:
+                case 3:
+                    return Int32Small[value + 1];
+                default:
+                    return value;
+            }
+        }
+        private static readonly object[] Int32Small = new object[] { -1, 0, 1, 2, 3 };
```
JamesNK, line 48: How come there is a range of cached 32/64bit numbers, but only zero for decimal? Why up to 3?

Is there a reason to use an array here rather than separate fields? Fields would avoid array access.

Src/Newtonsoft.Json/Utilities/BoxedPrimitives.cs
```diff
@@ -0,0 +1,83 @@
...
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+// OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+// HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+// WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+// OTHER DEALINGS IN THE SOFTWARE.
+#endregion
+
+
```
JamesNK, line 26:
```suggestion
```

Src/Newtonsoft.Json/Utilities/BoxedPrimitives.cs
```diff
@@ -0,0 +1,83 @@
...
+            }
+        }
+        private static readonly object[] Int64Small = new object[] { -1L, 0L, 1L, 2L, 3L };
+
+        internal static object Get(decimal value) => value == decimal.Zero ? DecimalZero : value;
+        private static readonly object DecimalZero = decimal.Zero;
+
+        internal static object Get(double value)
+        {
+            if (value == 0.0d) return DoubleZero;
```
JamesNK, line 71: Add braces plus new lines for consistency

Src/Newtonsoft.Json/Utilities/BoxedPrimitives.cs
mgravell: Fairly arbitrary based on observed use. Although in my "real" case, it was actually 0-5 that dominated. I can switch to fields, and: "pick a range" :)

mgravell: For decimal: as a gross over simplification - there don't *tend* to be as many common hits with decimals. Sure, we could pick a few integer values, but...

I'm open to suggestions

mgravell:
@JamesNK nits resolved

JamesNK:
Does this have any impact on comparisons when the value is still boxed? I wonder about people doing something dumb like:

```csharp
if (JValue1.Value == JValue2.Value)
{
}
```

Before it would be false but now it is true (if values are -1 to 8 and share a box)

mgravell:
@JamesNK  *yes*, technically the change is observable; however, the code as shown is already incorrect for most definitions of "correct"/"incorrect"; when dealing with boxed values, what they almost certainly *mean* is something like:

``` c#
if (Equals(JValue1.Value, JValue2.Value)
{
}
```

or

``` c#
if ((JValue1.Value.Equals(JValue2.Value))
{
}
```

The `==` usage is simply ... wrong. It will already behave very inconsistently. For example, if the values are strings, it would currently report `true` for two empty strings but *only* two empty strings.

I'm not sure we should be too concerned about obviously-broken code.

JamesNK:
Thanks @mgravell

PR #2643, opened 2022-01-24
Add async serialization and deserialization support

atykhyy:
This PR adds async serialization and deserialization support to Newtonsoft.Json. It uses a Roslyn source generator to generate async methods from their synchronous counterparts at build time. There is therefore no double code maintenance burden. Generated source is checked in for reference, easier debugging and PDB source links.

Async code paths are tested by the existing code suite by means of substituting calls to async methods at three key points: `JsonSerializer.SerializeInternal`, `JsonSerializer.PopulateInternal` and `JsonSerializer.DeserializeInternal`. Completeness of async code generation coverage is ensured by wrapping `JsonReader` and `JsonWriter` objects passed into these methods with wrappers which throw exceptions from all sync I/O methods.

At the moment the PR is in draft stage. The main project is set up to always test the async code paths (`TEST_ASYNC_AS_SYNC` preprocessor constant). Over 90% of the test suite passes in this mode, but tests that rely on custom converters in `Newtonsoft.Json.Converters` namespace fail because I haven't yet extended them with async implementations. I will add these if this PR has a chance of being merged. When `TEST_ASYNC_AS_SYNC` is not defined, tests use existing (synchronous) code paths. All tests pass in this mode.


sungam3r:
👍  I am amazed by people who continue attempts to invest in this project! On the other hand, it's a shame that the work performed with a probability is close to zero will ever get to the release.

Issue #2664, opened 2022-04-05
Only failes on parsing to floating point to int not long


gaotter:
Hi

In Newtonsoft.Json version 13.0.1

If one tries to deserialize a Json property that is a floating point to a int i C# one gets as expected an exception. But not when doing the same with a long value then is just get the casts it to a long, loosing it decimals. 

Should it also cast an exception when trying to deserialize a floating point value to long?

```
using Newtonsoft.Json;

var data = JsonConvert.DeserializeObject<SomData>(@"{""aInt"":1, ""ALong"":2.2}"); 

Console.WriteLine(data.ALong); // output 2

var data2 = JsonConvert.DeserializeObject<SomData>(@"{""aInt"":1.2, ""ALong"":2}"); // exception

Console.WriteLine(data2.AInt);


internal class SomData
{
    public int AInt { get; set; }
    public long ALong { get; set; }
}
```

colton-manville:
Hello,

This doesn't just impact `long` it also impacts `uint` and `ulong`. 

So `int` is the only type that throws an exception when deserializing JSON that contains a decimal value for an integral property. 

I tested this in Newtonsoft.Json version 13.0.3

Updated test code from the original problem statement:

```
using Newtonsoft.Json;

var data = JsonConvert.DeserializeObject<SomData>(@"{""aInt"":1, ""ALong"":2.2}"); // no exception

Console.WriteLine(data.ALong); // output 2

var data2 = JsonConvert.DeserializeObject<SomData>(@"{""aInt"":1, ""AUint"":2.2}"); // no exception

Console.WriteLine(data2.AUint); // output 2

var data3 = JsonConvert.DeserializeObject<SomData>(@"{""aInt"":1, ""AUlong"":2.2}"); // no exception

Console.WriteLine(data3.AUlong); // output 2

var data4 = JsonConvert.DeserializeObject<SomData>(@"{""aInt"":1.2, ""ALong"":2}"); // exception

Console.WriteLine(data4.AInt);


internal class SomData
{
    public int AInt { get; set; }
    public uint AUint {get;set;}
    public long ALong { get; set; }
    public ulong AUlong {get;set;}
}
```

Issue #654, opened 2015-09-18
Error converting float value to integer in Web API


pareshnawale:
Hi,

I am using Newtonsoft.Json NuGet package version 6.0.8. When I pass float value for the integer property of the model, model binding fails.

Here is the controller method , model and sample input 

Controller: 
```
 // POST api/values
        public string Post(MyModel model)
        {
               //some code here
        }
```
Model:
```
public class MyModel
    {
        public string ThisIsString { get; set; }
        public int ThisIsInt { get; set; }
        public decimal ThisIsDecimal { get; set; }
    }
```
Sample input:
```
{
    "ThisIsString": "string",
    "ThisIsInt": 2.20,  <-- Property is set to float number
    "ThisIsDecimal": 0
  }
```
When I tried above sample input, instead of receiving 0 for "ThisIsInt" property entire model received in controller method is null.

if I try same code using 6.0.3 version then I receive model with "ThisIsInt" set to 0 as expected.

Could you please help me resolve this issue? I have already tried upgrading  to latest version of NuGet package.

Thanks,
Paresh

JamesNK:
Later versions of Json.NET throw an error when deserializing a floating point value onto an integer property.

pareshnawale:
@JamesNK  Thanks for your reply. 
Is this by design or it is a bug? Is there any way to allow floating point value conversation into integer?

Thanks,
Paresh

JamesNK:
By design.

PR #2668, opened 2022-04-07
Fix Only failes on parsing from floating point to int not long 

isdaniel:
This MR will let  `int`  & `long` have the same behaviour for throwing an error when deserializing a floating-point value onto an integer property.

As before discussed in #654 Error converting float value to integer in Web API

But now that only worked in `int` but not in `long`

I had added some unit tests to make sure  `long` behaviour will be as same as `int`

I also saw there is another issue #2664, this MR might solve this problem.


isdaniel:
Hi @thomas-oliver-cko 

I am glad you like this MR.

I saw there were some conflicts created by the new modification.

Now, I have fixed it.

isdaniel:
Hi @JamesNK Could you please take look at this? Thanks.

Issue #2521, opened 2021-05-06
Support new DateOnly and TimeOnly structs in .NET 6


mattjohnsonpint:
.NET 6 (starting with preview 4) will support two new primitives: `System.DateOnly` and `System.TimeOnly`.
Refer to: https://github.com/dotnet/runtime/issues/49036

These are similar to `System.DateTime`, but split into subcomponents.  They should serialize/deserialize in their ISO 8601 formats, such as `"2021-12-31"` and `"23:59:59.9999999"`, respectively.

It would be great if these were supported out of the box by Json.NET.  However, that would require targeting `net6.0`, and presently Json.NET is built for `netstandard` targets only.  Alternatively, a separate library could be built to add custom `JsonConverter` objects for these types, though it seems a pitfall that users won't know they will need that bit of glue.

Unfortunately, the types do not implement `[Serializable]` (See [comment](https://github.com/dotnet/runtime/issues/49036#issuecomment-789453258)).  So I believe they will need to have support added directly.

Any thoughts on this would be appreciated.  Thanks.

mattjohnsonpint:
Converters would be like so:

```csharp
public class DateOnlyJsonConverter : JsonConverter<DateOnly>
{
    private const string DateFormat = "yyyy-MM-dd";

    public override DateOnly ReadJson(JsonReader reader, Type objectType, DateOnly existingValue, bool hasExistingValue, JsonSerializer serializer)
    {
        return DateOnly.ParseExact((string)reader.Value, DateFormat, CultureInfo.InvariantCulture);
    }

    public override void WriteJson(JsonWriter writer, DateOnly value, JsonSerializer serializer)
    {
        writer.WriteValue(value.ToString(DateFormat, CultureInfo.InvariantCulture));
    }
}

public class TimeOnlyJsonConverter : JsonConverter<TimeOnly>
{
    private const string TimeFormat = "HH:mm:ss.FFFFFFF";

    public override TimeOnly ReadJson(JsonReader reader, Type objectType, TimeOnly existingValue, bool hasExistingValue, JsonSerializer serializer)
    {
        return TimeOnly.ParseExact((string)reader.Value, TimeFormat, CultureInfo.InvariantCulture);
    }

    public override void WriteJson(JsonWriter writer, TimeOnly value, JsonSerializer serializer)
    {
        writer.WriteValue(value.ToString(TimeFormat, CultureInfo.InvariantCulture));
    }
}
```

[...]

PR #2673, opened 2022-05-14
Add DateOnly and TimeOnly converters

karl-sjogren:
Adds `DateOnlyConverter` and `TimeOnlyConverter` since this was the way STJ went with it in https://github.com/dotnet/runtime/pull/69160.

Solves #2521 if merged.

Adding this as a draft PR for now to get feedback about if this is the correct direction to go.

I've also lowered the CA1307 rule to a warning instead of an error. When building for net6.0 this one triggers over 400 errors about misssing `StringComparison` settings all over the place if left in place. If this one should be kept as error this should be fixed in a seperate PR before this one is merged.


Src/Newtonsoft.Json.Tests/Newtonsoft.Json.Tests.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks Condition="'$(TestFrameworks)'==''">net46;net40;net35;net20;net5.0;netcoreapp3.1;netcoreapp2.1</TargetFrameworks>
+    <!--TargetFrameworks Condition="'$(TestFrameworks)'==''">net6.0;net46;net40;net35;net20;net5.0;netcoreapp3.1;netcoreapp2.1</TargetFrameworks-->
+    <TargetFrameworks Condition="'$(TestFrameworks)'==''">net6.0;net5.0;netcoreapp3.1;netcoreapp2.1</TargetFrameworks>
```
karl-sjogren, line 4: These will of course need to be restored before merging, I didn't want to install all the targeting packs all the way back to net2.0 though.

Src/Newtonsoft.Json/Converters/TimeOnlyConverter.cs
```diff
@@ -0,0 +1,70 @@
...
+        public override TimeOnly ReadJson(JsonReader reader, Type objectType, TimeOnly existingValue, bool hasExistingValue, JsonSerializer serializer)
+        {
+            TimeSpan timeSpan;
+            if(!StringUtils.IsNullOrEmpty(_timeFormat)) {
+                timeSpan = TimeSpan.ParseExact((string)reader.Value, _timeFormat, CultureInfo.InvariantCulture);
+            } else {
+                timeSpan = TimeSpan.Parse((string)reader.Value, CultureInfo.InvariantCulture);
+            }
+
+            return TimeOnly.FromTimeSpan(timeSpan);
```
karl-sjogren, line 60: This is how they did it in STJ, parsing using a `TimeSpan` seems to accept a wider range of values then just `TimeOnly.Parse`.

JamesNK:
Thanks for this but I've gone for a different approach and added support without using converters.

PR here - https://github.com/JamesNK/Newtonsoft.Json/pull/2679

PR #2679, opened 2022-05-16
Add support for DateOnly and TimeOnly

JamesNK:
Fixes https://github.com/JamesNK/Newtonsoft.Json/issues/2521


VinzSpring:
When will this be released in the nuget package? We kinda need this asap... 🚀

JamesNK:
5 days ago

Gnimuc:
Hi @JamesNK, thank for fixing this! Would you mind releasing a new version so we don't need to patch this manually in our project?

JamesNK:
It’s on nuget

Gnimuc:
We are using the pre-released version 13.0.2-beta1 and it works well, but we're not allowed to use a beta lib in our products. :)

JamesNK:
You’ll need to build your apartment an version then

nesc58:
Is the next release already scheduled?

sjd2021:
This is working great for me @JamesNK. Thank you! 
I am running into some conflicts with using other packages like NSwag that aren't targeting any beta versions. Do you know when the next major release will be so that we could revisit using any packages that might have these problems?

goldenc:
@JamesNK what about adding support for `DateOnly` and `TimeOnly` in JValue ?

Issue #2681, opened 2022-05-20
JsonConvert.SerializeXNode: Global namespace on child nodes are lost and implicitly replaced with parent elements namespace


tkvalvik:
Emnpty namespaces are lost during serialization if parent-element has a non-empty namespace.
### Source/destination types

```csharp
var xml = new XElement("{x}a", new XElement("b", ""));
var json = JsonConvert.SerializeXNode(xml);
Console.WriteLine(json); 
```

### Source/destination JSON
```javascript
{
  "a": {
    "@xmlns": "x",
    "b": ""
  }
}
```

### Expected behavior
xmlns should be present on an element when it differs from its parent, even when it is the global namespace.

```javascript
{
  "a": {
    "@xmlns": "x",
    "b": {
      "@xmlns": ""
    }
  }
}
```
### Actual behavior
<b>-element has no xmlns-attribute, even though its namespace differs its parent.

This also means that empty namespaces are dropped in a serialize/deserialize round trip.

### Steps to reproduce

```csharp
var xml = new XElement("{x}a", new XElement("b", ""));
Console.WriteLine($"Before: \n:{xml}"); 
var json = JsonConvert.SerializeXNode(xml);
var deserializedXml = JsonConvert.DeserializeXNode(json);
Console.WriteLine($"After: \n:{deserializedXml}");
```

PR #2685, opened 2022-05-25
Fixed XElement serialization with explicit empty namespace.

MagMar94:
Fixes issue #2681

When an `XElement` child node has an explicit empty namespace, it should not be serialized with the namespace of the parent node. The child should have `"@xmlns": ""` .


Src/Newtonsoft.Json/Converters/XmlNodeConverter.cs
```diff
@@ -860,16 +860,21 @@ public override List<IXmlNode> Attributes
                     }
                     else
                     {
+                        bool namespaceAdded = false;
                         _attributes = new List<IXmlNode>();
                         foreach (XAttribute attribute in Element.Attributes())
                         {
+                            if(attribute.Name == "xmlns")
```
sungam3r, line 867:
```suggestion
                            if (attribute.Name == "xmlns")
```

Issue #2694, opened 2022-06-16
Add IAsyncDisposable to JsonReader and JsonWriter


JamesNK:
Calling Dispose can result in blocking IO. The fix is to add an async dispose option to clean up these types.

tomkerkhove:
I'm happy to work on this if you want.

PR #2695, opened 2022-06-16
Fix blocking IO in JsonTextWriter.CloseAsync and add IAsyncDisposable

JamesNK:
Fixes https://github.com/JamesNK/Newtonsoft.Json/issues/2694


Src/Newtonsoft.Json/JsonWriter.Async.cs
```diff
@@ -37,7 +37,20 @@
 namespace Newtonsoft.Json
 {
     public abstract partial class JsonWriter
+#if HAVE_ASYNC_DISPOABLE
```
sungam3r, line 40: HAVE_ASYNC_DISPOABLE -> HAVE_ASYNC_DISPOSABLE

Issue #815, opened 2016-02-11
Provide a way to do case-sensitive property deserialization


MuiBienCarlota:
Could you provides a way to opt-in to case-sensitive property deserialization?

It could take the form of a property of JsonSerializerSettings or JsonSerializer. This property could be a boolean or an enum type (permits a fine grain choice: in ctor and/or properties).

JsonPropertyCollection.GetClosestMatchProperty is used in:

- JsonSerializerInternalReader where is already present a property Serializer of type JsonSerializer,
- DefaultContractResolver whose instance is provided by JsonSerializer.

Emyr42:
Can you explain the use case for this feature please?

The C# style guides/standards I've read in the past have all proscribed having members differentiated solely by case.

MuiBienCarlota:
It's quite complexe but it can try to detailed my problem.

I use Json.Net to managed settings files used by my applications. These files should be mainly modiffied by my Apps but user can edit them manually.

I've one machine settings file and several user settings files. Machine settings file is not supposed to be editable. Only an administrator can distribute new machine settings files. Each modifications made by user are stores in user settings file and this file only contains "what is different"  from machine file. User data takes precedence over machine ones.

To implement this, I've used JSon.Net JContainer.Merge and a custom umnmerged algo to separate "what is different". Merge/unmerge is done at json DOM level (ie tree of JObjects). On load, after merging machine and user data, I use Json.Net deserialization (JObject.ToObject) to get a strongly typed .Net settings object. On save, I get a JObject tree fromJToken.FromObject, do an unmerge (a deep clone and delete "what is identical") to only store differences in user settings. And all was good and well.

I've just had a subtle bug when customer a have manually edit settings file but have made a casing error (mimetype vs Mimetype). During merge/unmerge works on JObject tree, comparaisons are made case sensitives but deserialization not (GetClosestMatchProperty). This difference of treatment break my diff algo.

YacovGold:
Also I need this feature.
I get sometimes very complex jsons with double names with only case sensitive difference (MyName is different from myName).
The thing is, I have no control over the stuff I get, and I only know what I need, so...
When I get the data, its get overridden by the data I don't care about :(

steevcoco:
**Wow**, I definitely want to try and chime in here. Myself, I feel that this should be very important to change. I feel a little worry. (Like people are getting a case of "facades in your eyes" over JSon.)

[...]

PR #2709, opened 2022-07-21
Implement case sensitive property matching

SakuraIsayeki:
#### This PR aims to address #815, by implementing a new serializer option, named `PropertyCaseSensitivityHandling`: 
```cs
public enum PropertyCaseSensitivityHandling
{
    // Case-insensitive matching (default).
    CaseInsensitive = 0,

    // Case-sensitive matching.
    CaseSensitive = 1
}
```

To preserve continuity with prior implementations, **default is set to `CaseInsensitive`**. 
With the new behaviour provided by `CaseSensitive`, we can now have differentiation and validation for casing, which is required by several API specs (Binance & xAPI, to name a few examples).

As this behaviour has an extremely lightweight implementation ([only one extra check on each Property matching](https://github.com/JamesNK/Newtonsoft.Json/blob/911809352f66368b66ffae6fae0b12467b99811a/Src/Newtonsoft.Json/Serialization/JsonPropertyCollection.cs#L143-L147)), **it does what it says on the tin, and only that**.  

Most of the matching with convention-following PascalCase C# properties will fail if you don't add `[JsonProperty("name")]` to each property, or write some esoteric conversion between JSON standard camelCase (or whichever casing is used) and C# Properties.

<hr /> 

##### A few footnotes...

This PR was inspired by the stale #601, where I factually stripped the Casing implementation to form my own.
Furthermore, as this is my first ever PR to Newtonsoft.Json *(hello guys!)*, I might've inadvertently broken a few long-standing conventions along the way, so any feedback will be most appreciated.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/CaseSensitivityHandlingTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/CaseSensitivityHandlingTests.cs
@@ -1,0 +1,74 @@
+#region License
+// Copyright (c) 2007 James Newton-King
+//
+// Permission is hereby granted, free of charge, to any person
+// obtaining a copy of this software and associated documentation
+// files (the "Software"), to deal in the Software without
+// restriction, including without limitation the rights to use,
+// copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the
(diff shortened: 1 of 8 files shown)


pierregarcia:
Really glad to see a PR about that. We had to fork the repo to make these changes. It would be nice to see them integrated in Newtonsoft

PR #2715, opened 2022-08-12
Add nullable type reflection cache

israellot:
Deserializing a class like the one has a performance hit due presumably due to calls to `Nullable.GetUnderlyingType`
This is aggravated when using converters like `StringEnumConverter ` which add a nullable type check at `CanConvert` method

```
public class NullablePropertiesTestClass
{
    public int? a { get; set; }            
    public TestEnum? b { get; set; }
    public bool? c { get; set; }

    public enum TestEnum
    {
        a,
        b,
        c
    }
}
``` 

This PR opens up the discussion of adding a simple cache for these expensive reflection calls. 
On this particular example, around 30% gains at the expense of the memory for holding the cache.
Benchmark code is included.

Screenshot 2022-08-12 132917
Screenshot 2022-08-12 133803


Src/Newtonsoft.Json/Utilities/ReflectionUtils.cs
```diff
@@ -1109,4 +1116,61 @@ public static bool IsMethodOverridden(Type currentType, Type methodDeclaringType
...
+
+        public static NullableTypeReflectionInfo GetTypeInfo(Type t)
+        {
+            ValidationUtils.ArgumentNotNull(t, nameof(t));
+
+            if (!t.IsGenericType() || !t.IsValueType())
+                return new NullableTypeReflectionInfo(false, t, null);
+
+#if !HAVE_CONCURRENT_DICTIONARY
+            lock(NullableInfoCache){
```
sungam3r, line 1155:
```suggestion
            lock(NullableInfoCache) {
```

Src/Newtonsoft.Json/Utilities/ReflectionUtils.cs
```diff
@@ -1109,4 +1116,61 @@ public static bool IsMethodOverridden(Type currentType, Type methodDeclaringType
...
+#endif
+            if (!NullableInfoCache.TryGetValue(t, out var info))
+            {
+                var genericTypeDefinition = t.GetGenericTypeDefinition();
+                var isNullable = (genericTypeDefinition == typeof(Nullable<>));
```
sungam3r, line 1160:
```suggestion
                var isNullable = genericTypeDefinition == typeof(Nullable<>);
```

israellot:
https://github.com/dotnet/runtime/issues/73860

Src/Newtonsoft.Json.Tests/Benchmarks/DeserializeNullableBenchmarks.cs
```diff
@@ -0,0 +1,92 @@
+﻿#region License
```
MazeMonster, line 1: License

PR #2741, opened 2022-09-22
Added async serialize and deserialize methods to JsonSerializer

hagenson:
I've added SerializeAsync and DeserializeAsync methods to JsonSerializer, rippling the required down the JsonConverter sublasses.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Converters/BinaryConverterTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Converters/BinaryConverterTests.cs
@@ -37,6 +37,9 @@ using Newtonsoft.Json.Converters;
 using Xunit;
 using Test = Xunit.FactAttribute;
 using Assert = Newtonsoft.Json.Tests.XUnitAssert;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Tests.TestObjects;
+using System.Linq;
 #else
 using NUnit.Framework;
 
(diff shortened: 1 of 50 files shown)


zhenomansh:
@JamesNK, We are waiting for Newtonsoft.Json asynchronous support for already 3 years. :)

Such a functionality is highly important for my company, as we are using json serialization over gRPC with passing data of hundreds gigabytes. For the memory economy we are currently using StreamWriter functionality, only available in Newtonsoft.Json.

Something like:
```
        var serializer = new JsonSerializer();
        var writeStream = new StreamInterceptor(serializeChunkAction);
        using (var writer = new StreamWriter(writeStream, Encoding.UTF8, chunkSize))
          serializer.Serialize(writer, o, parameterType);

```
With the StreamInterceptor class we are able to intercept the stream.Write methods with chunks of serialized data and pass the chunks over network and deserialize them using the similar approach on the server side.

The only problem now is - due to the absence of async support in Newtonsoft.Json, we have to hold an extra thread on a server while passing all chunks per a procedure call, and cannot use asynchronous network calls, which makes the functionality very limited.

I've spent 1 day to add the serialization support to Newtonsoft.Json, and it looks similar to what @hagenson wanted to merge in 2022, and basically, it is quite a simple change.
Basically, the JsonSerializer class should have SerializeAsync method, which calls for WriteAsync methods of the passing StreamWriter stream.

Sure we want to stay with the main Newtonsoft.Json release channel, and have all upcoming updates.
So, the question is - are you considering to accept the async commits in the nearest future? 
If you are not agree with some concepts from the commit, can I prepare another commit, based on the latest release, to successfully include it into Newtonsoft.Json master branch?

hagenson:
@zhenomansh I have given up on this change making into Newtonsoft.Json.
I feel like this is a legacy code base now.
It's only going to fall further behind if the moderators can't take new PRs in.
I've started work on porting my stuff over to the Microsoft JSON serialiser - for my use case, I need to implement polymorphic serialisation/deserialisation as the MS support for this is weak.
You may need to consider doing the same.

PR #2747, opened 2022-10-06
Add JsonCloneSettings to disable copy annotations

snshivakumar:


Src/Newtonsoft.Json/Linq/JArray.cs
```diff
@@ -93,7 +93,7 @@ internal override bool DeepEquals(JToken node)
             return (node is JArray t && ContentsEqual(t));
         }
 
-        internal override JToken CloneToken()
+        internal override JToken CloneToken(JsonCloneSettings? settings = null)
         {
             return new JArray(this);
         }
```
laveeshb, line 99: It seems odd to me that even when `settings` were provided, this method will not be using it. 
Is it because the clone settings won't apply to `JArray` or because they don't apply to the scenario we are trying to solve which seems specific to `JValue`?

laveeshb: If we think this is something that should be defined at `JToken` level, then I would imagine every implementation of `JToken` should make use of it. Otherwise, we should just introduce an new overload in `JValue`.

laveeshb (changes requested):

Src/Newtonsoft.Json/Linq/JArray.cs
snshivakumar: Thanks for catching this. I included the new settings object within all JToken implementations.

Src/Newtonsoft.Json.Tests/Linq/AnnotationsTests.cs
```diff
@@ -308,6 +308,21 @@ public void MultipleAnnotationsAreCopied()
             Assert.AreEqual(annotation, t.DeepClone().Annotation<T>());
         }
 
+        [Test]
+        public void MultipleAnnotationsAreNotCopiedWithSetting()
```
laveeshb, line 312: the code changes look good.
As we are updating all the other derivates of JToken, we should add case for each one in the test. I think a single unit test for all should be sufficient, let's just clone each type in the test and verify.

snshivakumar: Added. Also, the test passes.

laveeshb (approved):

Issue #2768, opened 2022-12-06
Parsing 0.0 (decimal) different in 13.0.2


marcovisserFurore:
After using Newtonsoft.Json version 13.0.2, I encountered an unwanted behavior.

### Source/destination types

```csharp
        [TestMethod]
        public void ParseJsonDecimal()
        {
            var json = @"{ ""property"": 0.0}";

            var reader = new JsonTextReader(new StringReader(json))
            {
                FloatParseHandling = FloatParseHandling.Decimal
            };

            while (reader.Read())
            {
                if (reader.TokenType == JsonToken.Float)
                {
                    if (reader.Value is decimal d)
                    {
                        d.ToString(NumberFormatInfo.InvariantInfo).Should().Be("0.0");

                    }
                    else throw new AssertFailedException("Value is not decimal");
                }
            }
        }
```

### Expected behavior

I would expect that the decimal `0.0` contains still the trailing zero after parsing.

### Actual behavior

Decimal looses precision after parsing. The trailing zero (`0.0`) is gone.

### Steps to reproduce
See above unit test.

elgonzo:
Nice catch.

The cause of the issue is in commit https://github.com/JamesNK/Newtonsoft.Json/commit/13717cfc452ee9c1ea0b4fc368b8e9922d562028#diff-188cdcd35a8220c2477c864944a945fa67539c0beac51d3909424ff07dea7fe4 based on this pull request https://github.com/JamesNK/Newtonsoft.Json/pull/2639

Specifically, the `BoxedPrimitives.Get(decimal)` method: https://github.com/JamesNK/Newtonsoft.Json/blob/4fba53a324c445f06ee08e45a015c346000a7ef2/Src/Newtonsoft.Json/Utilities/BoxedPrimitives.cs#L90-L92

For any numerically zero value with trailing zeroes like 0.0, this method unfortunately returns `decimal.zero` which is just 0 without any trailing zeros. I don't know why this had to be hardcoded instead of making it a configurable choice in the JsonTextReader and JsonSerializerSettings...

JamesNK:
Should fix this.

Is there a way to detect whether a decimal has trailing zeroes without ToString? If there is then decimals with trailing zeroes can skip the cache. If there isn't then remove caching of decimals altogether.

elgonzo:
I know of two ways. However, both are available only beginning with .NET 7. 

[...]

PR #2769, opened 2022-12-07
Fix parsed decimal losing trailing zeroes

JamesNK:
Fixes https://github.com/JamesNK/Newtonsoft.Json/issues/2768

### DIFF
--- a/Src/Newtonsoft.Json/Utilities/BoxedPrimitives.cs
+++ b/Src/Newtonsoft.Json/Utilities/BoxedPrimitives.cs
@@ -23,6 +23,9 @@
 // OTHER DEALINGS IN THE SOFTWARE.
 #endregion
 
+using System;
+using System.Diagnostics;
+
 namespace Newtonsoft.Json.Utilities
 {
     internal static class BoxedPrimitives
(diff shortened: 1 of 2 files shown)


Src/Newtonsoft.Json/Utilities/BoxedPrimitives.cs
```diff
@@ -87,9 +90,42 @@ internal static class BoxedPrimitives
         internal static readonly object Int64_7 = 7L;
         internal static readonly object Int64_8 = 8L;
 
-        internal static object Get(decimal value) => value == decimal.Zero ? DecimalZero : value;
+        internal static object Get(decimal value)
+        {
+            // Decimals can contain trailing zeros. For example 1 vs 1.0. Unfortunatly, Equals doesn't check for trailing zeros.
+            // There isn't a way to find out if a decimal has trailing zeros in older frameworks without calling ToString.
```
sungam3r, line 96: Check this  https://github.com/graphql-dotnet/graphql-dotnet/search?q=DecimalData&type=
Should work for all platforms.

JamesNK: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafe.as?view=net-7.0

This would only be available for the net6.0 target, which already has `decimal.GetBits`. Anyway, I prefer to use public API for getting these values.

sungam3r: ?
https://www.nuget.org/packages/System.Runtime.CompilerServices.Unsafe/7.0.0-preview.2.22152.2#supportedframeworks-body-tab

netstandard2.0 compatible

JamesNK: I won't introduce a dependency for a small perf optimization

sins921 (approved):

PR #2823, opened 2023-02-22
Use 17 digits of precision with Double.ToString

oleks:
Using 15 digits ("R") leads to failure of roundtrip serialization/deserialization.

See also https://learn.microsoft.com/en-us/dotnet/api/system.double.tostring

Another example is 0.6822871999174


oleks:
I suppose one could make the argument that it is better to parameterize the library by the format that it should use for `Double.ToString`, since G17 is also broken, just in a different way.

```
Serialize(Deserialize<float>("1.1")) != "1.1"
```

Hence, it is probably best to let the users choose their poison.

PR #2852, opened 2023-04-30
Small fixes

rinatxp:
- converting some switch statements to switch expressions;
- combining several Write() into one;
- other small fixes


Src/Newtonsoft.Json/JsonTextWriter.cs
```diff
@@ -785,9 +775,7 @@ public override void WriteComment(string? text)
         {
             InternalWriteComment();
 
-            _writer.Write("/*");
-            _writer.Write(text);
-            _writer.Write("*/");
```
sungam3r, line 790: Such calls here are intentionally to avoid string allocations when doing string interpolation.

Src/Newtonsoft.Json/JsonWriter.cs
```diff
@@ -380,10 +369,10 @@ private JsonContainerType Pop()
         {
             JsonPosition oldPosition = _currentPosition;
 
-            if (_stack != null && _stack.Count > 0)
+            if (_stack?.Count > 0)
             {
-                _currentPosition = _stack[_stack.Count - 1];
-                _stack.RemoveAt(_stack.Count - 1);
+                _currentPosition = _stack.Last();
+                _stack.Remove(_stack.Last());
```
sungam3r, line 375: ` _stack.RemoveAt(_stack.Count - 1);` is not the same as  `_stack.Remove(_stack.Last());`

Src/Newtonsoft.Json/JsonTextWriter.cs
rinatxp: Cancelled all such commits

Src/Newtonsoft.Json/JsonWriter.cs
rinatxp: Cancelled this too

rinatxp:
Reverted bad commits and added some new

Issue #1295, opened 2017-04-25
Different de-serialization behavior for immutable struct and class when DataMember attributes are present


illinar:
De-serialization of immutable struct requires argument name in the constructor to match the field name as specified by the `DataMember` attribute and not the .NET property name. _This behavior is different from class de-serialization._

Here is the test case:
```
[DataContract]
public class ImmutableClass
{
    public ImmutableClass (int value1, int value2)
    {
        Value1 = value1;
        Value2 = value2;
    }

    [DataMember (Name = "value1")]
    public int Value1 { get; }

    [DataMember (Name = "xyz")]
    public int Value2 { get; }
}

[DataContract]
public struct ImmutableStruct
{
    public ImmutableStruct (int value1, int value2)
    {
        Value1 = value1;
        Value2 = value2;
    }

    [DataMember (Name = "value1")]
    public int Value1 { get; }

    [DataMember (Name = "xyz")]
    public int Value2 { get; }
}
```

Both serialize correctly into:
```
{"value1":...,"xyz":...}
```

However deserialization works for `ImmutableClass` but not for `ImmutableStruct` unless the second argument in the latter is changed to `xyz`. `ImmutableClass` is deserialized correctly using either option.

JamesNK:
There is different behavior because unlike classes, even a struct with a parameterized constructor will still have a default constructor, and so it will only be called if every parameter has a matching property.

dbc2:
There does seem to be an inconsistency here between contract generation and deserialization, that leads to the observed behavior.

When selecting a parameterized constructor for a value type, [`DefaultContractResolver.GetImmutableConstructor()`](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Serialization/DefaultContractResolver.cs#L586) walks through the parameterized constructors to see if there is exactly one whose arguments match the serialized properties by name, by calling [`MatchProperty()`](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Serialization/DefaultContractResolver.cs#L670) on the constructor argument name.  This method looks for a `JsonProperty` among the list of serialized properties whose serialized name is the same as the argument name, and fails.

[...]

PR #2860, opened 2023-05-17
Enable roundtrip for F# struct record with datamember attribute

T-Gro:
This addresses #1295 , specifically the part which started to regress after F#7 added the "readonly" modified to compiler generated records.

The solution is to adjust the heuristic in order to find a good fitting parametrized constructor, and avoid fallback to default one.
Why did this work before? Because a default constructor was used, and private property/field injected later. However, with the "readonly" modifier, field injection no longer works.

This was always working fine without using mapping attributes, because newtonsoft found a constructor argument by its name.
The change in this PR just extends this to also look for constructor argument using the underlying name, and not just the name of the JSON property.

### DIFF
--- a/Src/Newtonsoft.Json/Serialization/DefaultContractResolver.cs
+++ b/Src/Newtonsoft.Json/Serialization/DefaultContractResolver.cs
@@ -727,7 +727,7 @@ namespace Newtonsoft.Json.Serialization
 
             JsonProperty? property = 
                 properties.GetClosestMatchProperty(name) ??
-                // In case of name-napping attributes such as DataMember, match the underlying name as a constructor param.
+                // In case of name-mapping attributes such as DataMember, match the underlying name as a constructor param.
                 // This is needed for types with readonly properties, such as F# records.
                 properties.GetByUnderlyingName(name);
             // must match type as well as name


Src/Newtonsoft.Json.Tests/Issues/Issue1295.cs
```diff
@@ -0,0 +1,34 @@
+﻿using Newtonsoft.Json.Tests.TestObjects;
```
JamesNK, line 1: Add the standard license header

Src/Newtonsoft.Json.Tests/TestObjects/FsharpStructRecordWithDataMember.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
```
JamesNK, line 1: Add the standard license header

Src/Newtonsoft.Json/Serialization/JsonPropertyCollection.cs
```diff
@@ -135,7 +135,7 @@ public void AddProperty(JsonProperty property)
         /// <param name="propertyName">Name of the property.</param>
         /// <returns>A matching property if found.</returns>
         public JsonProperty? GetClosestMatchProperty(string propertyName)
-        {
+        {
```
JamesNK, line 138:
```suggestion
        {
```

Src/Newtonsoft.Json/Serialization/JsonPropertyCollection.cs
```diff
@@ -186,5 +186,22 @@ private bool TryGetProperty(string key, [NotNullWhen(true)]out JsonProperty? ite
...
+        /// </summary>
+        /// <param name="propertyName">The name of the property to get.</param>   
+        /// <returns>A matching property if found.</returns>
+        public JsonProperty? GetByUnderlyingName(string propertyName)
+        {
+            for (int i = 0; i < _list.Count; i++)
+            {
+                JsonProperty property = _list[i];
+                if (string.Equals(propertyName, property.UnderlyingName, StringComparison.OrdinalIgnoreCase))
+                    return property;
```
JamesNK, line 201:
```suggestion
                if (string.Equals(propertyName, property.UnderlyingName, StringComparison.OrdinalIgnoreCase))
                {
                    return property;
                }
```

Src/Newtonsoft.Json/Serialization/DefaultContractResolver.cs
```diff
@@ -724,7 +725,9 @@ protected virtual IList<JsonProperty> CreateConstructorParameters(ConstructorInf
                 return null;
             }
 
-            JsonProperty? property = properties.GetClosestMatchProperty(name);
+            JsonProperty? property =
```
JamesNK, line 728: Add a comment here about why to fallback to GetByUnderlyingName

Src/Newtonsoft.Json.Tests/Issues/Issue1295.cs
```diff
@@ -0,0 +1,34 @@
...
+            Assert.AreEqual(fsharpStructRecord.Foo, deserialized.Foo);
+        }
+
+        [Test]
+        public void TestJsonSanity()
+        {
+            var fsharpStructRecord = new FSharpStructRecordWithDataMember("42");
+            var asJson = JsonConvert.SerializeObject(fsharpStructRecord);         
+
+            Assert.AreEqual(@"{""foo_field"":""42""}", asJson);
```
JamesNK, line 56: This test fails https://github.com/JamesNK/Newtonsoft.Json/pull/2860/checks?check_run_id=13544125215

T-Gro: Thanks, have to #ifdef the value for net2.0

T-Gro:
Hi all, is there anything else I should improve here, or is this ready to be merged now?

JamesNK:
I need time to think this through and understand what breaking changes it could cause. I'm really busy right now, so I don't know when I'll get to that.

T-Gro:
@JamesNK : Hi James, would now be a better time to merge this in please?

PR #2877, opened 2023-07-17
Fixes #2876 Allow the contents of a JsonToken.Raw token to be returne…

### LINKED ISSUE #2876: JsonReader.ReadAsString() doesn't return the contents of JsonToken.Raw tokens
### Source/destination types and Repro Steps

```csharp
[JsonConverter(typeof(DecimalLikeConverter))]
public class DecimalLikeThing
{
    public decimal D { get; set; }
}

public class DecimalLikeConverter : JsonConverter
{
    public override bool CanConvert(Type objectType) => true;

    public override object? ReadJson(JsonReader reader, Type objectType, object? existingValue, JsonSerializer serializer)
    {
        reader.Read();
        var value = reader.ReadAsString();
        var result = new DecimalLikeThing
        {
            D = decimal.Parse(value ?? "0"),
        };

        reader.Read();
        return result;
    }
    
    public override void WriteJson(JsonWriter writer, object? value, JsonSerializer serializer)
    {
        var x = (DecimalLikeThing)value!;
        writer.WriteStartObject();
        writer.WritePropertyName("D");
        writer.WriteRawValue(x.D.ToString());
        writer.WriteEndObject();
    }
}

public void Run()
{
    var x = new DecimalLikeThing
    {
       D = 100.300M,
    };

    var res = JToken.FromObject(x);
    res.ToObject<DecimalLikeThing>(); //<-- This throws the exception noted below

    var text = res.ToString(Formatting.None);
    var text = res.ToString(Formatting.None);
    var f = JsonConvert.DeserializeObject(text, typeof(DecimalLikeThing)); // <-- This succeeds because the JToken doesn't contain a token noted to by of type Raw
}
```

### Source/destination JSON

```javascript
{
  "D": 100.300
}
```

### Expected behavior

The `ReadAsString()` call in the JsonConverter above should have returned the text captured in the Raw type token

### Actual behavior

`Newtonsoft.Json.JsonReaderException: 'Error reading string. Unexpected token: Raw. Path 'D'.'`
   at Newtonsoft.Json.JsonReader.ReadAsString() in /_/Src/Newtonsoft.Json/JsonReader.cs:line 507

mlorbetske:
…d when calling JsonReader.ReadAsString()

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/JsonSerializerTest.cs
@@ -8213,5 +8213,18 @@ This is just junk, though.";
 
             return depth;
         }
+
+        [Test]
+        public void RoundTripDecimalContainerWithConverter()
+        {
+            DecimalContainer container = new DecimalContainer
+            {
+                Value = 200.100M
+            };
+
+            JToken wrapped = JToken.FromObject(container);
+            DecimalContainer newContainer = wrapped.ToObject<DecimalContainer>();
+            Assert.Equals(container.Value, newContainer.Value);
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


mlorbetske:
I'll fix the build error for using nullability annotations in a non-nullable context shortly

Issue #2887, opened 2023-08-22
AssemblyVersion only contains Major version causes DLL conflicts


manuelmainz:
### Environment
I am using Nuget package **13.0.3** and my application has target framework **net48**.

On a customer machine any other tool has registered an older version (probably 13.0.1) of Newtonsoft.Json in GAC (Global Assembly Cache).
Both versions of Newtonsoft have Assembly version 13.0.0.0.

Even though our application does not register Newtonsoft in GAC, we still break because another application, which is out of our control, (not really sure which one it is) and installed on client machine registers in GAC.
So you get the complains by the customer, even though it is not even your fault.

In my opinion the Assembly version should always contain the Major and Release version number. In this way the application is able to determine if there is the same version in GAC or not.

### Source code

```csharp
var a = new JsonSerializerSettings();
a.Formatting = Formatting.Indented;

var b = new JsonSerializerSettings(a);
```

### Expected behavior
The assembly Newtonsoft v 13.0.3.0 out of my binary directory should get loaded.

### Actual behavior
The assembly gets loaded from GAC and causes following exception to be thrown, because this constructor has been introduced with version 13.0.2.
`System.MissingMethodException: 'Method not found: 'Void Newtonsoft.Json.JsonSerializerSettings..ctor(Newtonsoft.Json.JsonSerializerSettings)'.'`

### Steps to reproduce

1. download Newtonsoft version 13.0.1.0 somewhere locally on your machine
2. register  Newtonsoft.Json.dll in GAC by executing 
`gacutil.exe /i "C:\temp\Json130r1\Bin\net45\Newtonsoft.Json.dll"`
3. Run your application which contains above C# code including the copy constructor.

### For reference to cleanup your local machine again:
To unregister the assembly from GAC run `gacutil.exe /u Newtonsoft.Json`

manuelmainz:
Similar issues:
- `https://github.com/JamesNK/Newtonsoft.Json/issues/615 `
- `https://github.com/JamesNK/Newtonsoft.Json/issues/2662`

readyjoseph:
We are having similar issue. The azure build pipeline artefact throws 
"Could not load file or assembly 'Newtonsoft.Json, Version=13.0.1.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)"
But locally published artefact works without issue. 
[...]

PR #2888, opened 2023-08-22
issue #2887 set release number in assembly version AssemblyVersion

manuelmainz:
AssemblyVersion should always be unique specified for each release.
This is necessary to avoid version mismatch with dlls in GAC
See [AssemblyVersion only contains Major version causes DLL conflicts #2887](https://github.com/JamesNK/Newtonsoft.Json/issues/2887)


LanceFHadBiomerieux:
This problem is breaking our code deployment. Is it possible to get this merged and released at some point soon?

JamesNK:
Not updating the assembly version every release is an intentional decision to reduce problems around binding redirects. Newtonsoft.Json has had this behavior for years.

Issue #2900, opened 2023-09-20
Please allow setting the max depth in JObject and JArray's Parse method 


tradercentric:
Hi James,

For Avro message deserialization, I am in need for Confluent to support parsing Avro schema that is more than 64 depth levels where Confluent library is using Newtonsoft library to parse into Avro schema object.

Currently, there is a issue logged with with Apache Avro where Confluent is not able parse Avro schema where the json is over 64 depth level.   

The references to the issue are logged at

https://issues.apache.org/jira/browse/AVRO-3856
https://github.com/apache/avro/blob/41b3c08ca5da192786c2b08546e691b3126e1856/lang/csharp/src/apache/main/Schema/Schema.cs#L250

I think we would need your help to add the feature to set the max depth for JObject and JArray Parse method.  Here is what I am thinking if that is feasible if you agree.
   
In Newtonsoft.Json.Linq.JsonLoadSettings, 

public int? MaxDepth { get; set; }       // add

In Newtonsoft.Json.Linq.JObject

        public new static JObject Parse(string json, JsonLoadSettings? settings)
        {
            using (JsonReader reader = new JsonTextReader(new StringReader(json)))
            {
                var maxDepth = settings?.MaxDepth;         //add
                if (maxDepth != null)                                   //add
                {
                    reader.MaxDepth = maxDepth;               //add
                }

In Newtonsoft.Json.Linq.JArray

        public new static JArray Parse(string json, JsonLoadSettings? settings)
        {
            using (JsonReader reader = new JsonTextReader(new StringReader(json)))
            {
                var maxDepth = settings?.MaxDepth;        //add
                if (maxDepth != null)                                  //add
                {
                    reader.MaxDepth = maxDepth;             //add
                }
                
I attached a schame.txt which contain a json more than 64 depth level for replicate the issue.

Thanks

Edmond Wong

P.S. I noticed Microsoft.Text.Json.Nodes support the passing of max depth, so I hope Newtonsoft can support the feature as well.

using System.Text.Json;
using System.Text.Json.Nodes;

var schema = File.ReadAllText("Schemas/schema.txt");

JsonDocumentOptions options = new()
{
    MaxDepth = 128
};

JsonNode.Parse(schema, null, options);
[schema.txt](https://github.com/JamesNK/Newtonsoft.Json/files/12669094/schema.txt)

tradercentric:
[...]

PR #2904, opened 2023-09-21
Added MaxDepth property to JsonLoadSettings to override the maxDepth …

tradercentric:
…of 64 levels limit in JsonReader when the reader is called by JObject.Parse, or JArray.Parse.


tradercentric:
Hi, this is a fix for https://github.com/JamesNK/Newtonsoft.Json/issues/2900

Src/Newtonsoft.Json.Tests/Issues/Issue2900.cs
```diff
@@ -0,0 +1,330 @@
...
+    public class Issue2900
+    {
+        [Test]
+        //https://github.com/JamesNK/Newtonsoft.Json/issues/2900
+        public void Override64DepthLevelOnJObjectParse()
+        {
+
+            var json = @"
+{
+  ""level1"": {
```
sungam3r, line 24: You may use raw string literals to avoid `""` in json.

Src/Newtonsoft.Json/Linq/JsonLoadSettings.cs
```diff
@@ -36,6 +36,11 @@ public class JsonLoadSettings
         private LineInfoHandling _lineInfoHandling;
         private DuplicatePropertyNameHandling _duplicatePropertyNameHandling;
 
+        /// <summary>
+        /// Setting MaxDepth to override JsonReader if necessary.
```
sungam3r, line 40:
```suggestion
        /// Setting MaxDepth to override <see cref="JsonReader"/> if necessary.
```

Src/Newtonsoft.Json.Tests/Issues/Issue2900.cs
tradercentric: I got this error when after I change to string literal,: 
  Invalid option '11' for /langversion. Use '/langversion:?' to list supported values.
Not sure if I can do that.

tradercentric:
I really need the issue fixed soon because  I do have an Avro schema from the source team who cannot reduce the depth level the document from a vendor.   I am open to alternative if my fix is not good enough.  Please help.

Src/Newtonsoft.Json.Tests/Issues/Issue2900.cs
sungam3r: You can, but nevermind, not an issue for this pr.

sungam3r:
@tradercentric if you in a hurry I really advise you to fork and compile your version. You may wait for applying/releasing for ages.

JamesNK:
I don't think this feature is valuable. There is already a `JObject.Load(JsonReader reader, JsonLoadSettings? settings)` method that lets you customize the reader.

Having a setting on JsonLoadSettings for max depth creates confusion about which is used. And then the question comes up whether all the other settings on JsonReader should be added to JsonLoadSettings.

tradercentric:
Thanks for the response.  I am an user of the Apache Avro project which is using Newtonsoft's library, I will fork the Apache Avro project and see if I can instantiate JsonReader first before (to customize) before using JObject.Parse and JArray.Parse.  That is to resolve the issue in https://github.com/apache/avro/blob/41b3c08ca5da192786c2b08546e691b3126e1856/lang/csharp/src/apache/main/Schema/Schema.cs#L250.

tradercentric:
Hi @JamesNK , JsonReader seems to severely overcount the depth levels of Avro schema, the following Avro schema has depth level  4.  However in JsonReader's Push(JsonContainerType value), when I put a break point on the line "if (_maxDepth != null && Depth + 1 > _maxDepth && !_hasExceededMaxDepth)", it counted 11. 

Here is the sample Avro schema for the test:

var json = @"{
  ""type"": ""record"",
  ""name"": ""Level1"",
  ""fields"": [
    {
      ""name"": ""field1"",
      ""type"": ""string""
    },
    {
      ""name"": ""field2"",
      ""type"": ""int""
    },
    {
      ""name"": ""level2"",
      ""type"": {
        ""type"": ""record"",
        ""name"": ""Level2"",
        ""fields"": [
          {
            ""name"": ""field3"",
            ""type"": ""boolean""
          },
          {
            ""name"": ""field4"",
            ""type"": ""double""
          },
          {
            ""name"": ""level3"",
            ""type"": {
              ""type"": ""record"",
              ""name"": ""Level3"",
              ""fields"": [
                {
                  ""name"": ""field5"",
                  ""type"": ""string""
                },
                {
                  ""name"": ""field6"",
                  ""type"": ""int""
                },
                {
                  ""name"": ""level4"",
                  ""type"": {
                    ""type"": ""record"",
                    ""name"": ""Level4"",
                    ""fields"": [
                      {
                        ""name"": ""field7"",
                        ""type"": ""boolean""
                      },
                      {
                        ""name"": ""field8"",
                        ""type"": ""double""
                      }
                    ]
                  }
                }
              ]
            }
          }
        ]
      }
    }
  ]
}";

Issue #1947, opened 2018-12-25
JsonConvert.DeserializeObject can throw several unexpected exceptions


Metalnem:
**JsonConvert.DeserializeObject** can throw several unexpected exceptions (**JsonReaderException** is the one that is usually expected). These are:

- **ArgumentException**
- **JsonSerializationException**
- **JsonWriterException**
- **NullReferenceException**

Documentation doesn't say anything about this, but deserialization is definitely weird place for a **JsonWriterException**. **ArgumentException** and **NullReferenceException** should not be possible in my opinion (if the input strings are non-empty). Here is the complete program to reproduce all of this:

```csharp
using System;

namespace Newtonsoft.Json.Run
{
  public class Program
  {
    public static void Main(string[] args)
    {
      try
      {
        JsonConvert.DeserializeObject("[{\"click\":\"X\",\"Close()\"://[{\"click\":\"Kpen()\"},{\"cOOOOOOOOOO\u007f]}}}");
      }
      catch (ArgumentException) { }

      try
      {
        JsonConvert.DeserializeObject("{\"menu\":{\"id\":1,\"val\":\"X\",\"pop\":{\"a\":[{\"click\":\"Open()\"},{\"click\":\"Close()\"}]}}");
      }
      catch (JsonSerializationException) { }

      try
      {
        JsonConvert.DeserializeObject("{\"menu\":{\"id\":1,\"val\":\"X\",\"pop\":{\"a\":[{\"clicT\":\"Open()\"},");
      }
      catch (JsonWriterException) { }

      try
      {
        JsonConvert.DeserializeObject("[new\0([");
      }
      catch (NullReferenceException) { }
      }
    }
}

```

Found via [SharpFuzz].

[SharpFuzz]: https://github.com/Metalnem/sharpfuzz

AaronJThompson:
The last 3 seem to have been fixed. However, the first one now throws the error:
`Can not add Newtonsoft.Json.Linq.JValue to Newtonsoft.Json.Linq.JObject.`
Which I presume is still a bug

PR #2922, opened 2023-12-07
Fix handling of empty constructor names

### LINKED ISSUE #2921: Unexpected exceptions during deserialization due to missing constructor name
### Sample code

```csharp
using Newtonsoft.Json;

try {
    Console.WriteLine(JsonConvert.DeserializeObject("[new \0("));
} catch (JsonException) {
    Console.WriteLine("invalid JSON");
}
```

### Expected/actual behavior

Should print "invalid JSON". Instead crashes with a `NullReferenceException`.

### Analysis

The crash happens on the expression [`_parent!.Parent`](https://github.com/JamesNK/Newtonsoft.Json/blob/01e1759cac40d8154e47ed0e11c12a9d42d2d0ff/Src/Newtonsoft.Json/Linq/JTokenWriter.cs#L135), so the _damnit_ operator here is inaccurate.

The sequence of events is:
- `JTokenWriter.WriteStartConstructor` invokes [`AddParent(new JConstructor(name))`](https://github.com/JamesNK/Newtonsoft.Json/blob/01e1759cac40d8154e47ed0e11c12a9d42d2d0ff/Src/Newtonsoft.Json/Linq/JTokenWriter.cs#L161) with an empty `name`. This throws an `ArgumentException`, so `AddParent` is not invoked, even though `WriteStartConstructor` has successfully been called
- `JsonWriter.Dispose` is invoked, which does `Close → AutoCompleteAll → WriteEnd → WriteEndConstructor → InternalWriteEnd → AutoCompleteClose → JTokenWriter.WriteEnd → JTokenWriter.RemoveParent`, removing a level of parenting that doesn't exist

### Proposed fixes?

Moving the `new JConstructor` call to the start of `WriteStartConstructor` validates the name before `WriteStartConstructor` is invoked. This stops the NRE but still raises an `ArgumentException` instead of a `JsonException`. Probably the validation of the constructor name should happen much earlier, during reading not writing.

Also, should `JsonWriter` be completing its work if there is an exception in-flight?

## ArgumentException

It is also possible to surface the `ArgumentException` itself via:

```csharp
using Newtonsoft.Json;

try {
    Console.WriteLine(JsonConvert.DeserializeObject("{' ':new \0("));
} catch (JsonException) {
    Console.WriteLine("invalid JSON");
}
```

### Expected/actual behavior

Should print "invalid JSON". Instead crashes with a `ArgumentException` ("Constructor name cannot be empty").

### Analysis

Underlying cause is the same as previous but this one avoids the call to `Dispose` so it doesn’t trigger the NRE.

----

**Credit**: Found via [LibFuzzer](https://llvm.org/docs/LibFuzzer.html) & [SharpFuzz](https://github.com/Metalnem/sharpfuzz).

Porges:
See #2921 for details.

Closes #2921 & some of #1947.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/JsonConvertTest.cs
+++ b/Src/Newtonsoft.Json.Tests/JsonConvertTest.cs
@@ -119,6 +119,17 @@ namespace Newtonsoft.Json.Tests
         }
 
         [Test]
+        public void NoConstructorName() {
+            ExceptionAssert.Throws<JsonException>(
+                () => JsonConvert.DeserializeObject("[new \0("),
+                "Empty constructor name. Path '', line 1, position 6.");
+
+            ExceptionAssert.Throws<JsonException>(
+                () => JsonConvert.DeserializeObject("{'x':new \0("),
+                "Empty constructor name. Path 'x', line 1, position 10.");
+        }
+
+        [Test]
         public void PopulateObjectWithOnlyComment()
         {
             var ex = ExceptionAssert.Throws<JsonSerializationException>(() =>
--- a/Src/Newtonsoft.Json/JsonTextReader.cs
+++ b/Src/Newtonsoft.Json/JsonTextReader.cs
@@ -1938,7 +1938,8 @@ namespace Newtonsoft.Json
                     }
                 }
                 
-                if (initialPosition == endPosition) {
+                if (initialPosition == endPosition)
+                {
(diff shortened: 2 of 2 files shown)


JamesNK (approved):

JamesNK:
Thanks. Looks good to me.

Issue #2927, opened 2024-01-10
Multiple JsonSerializerSettings are sharing same CamelCasePropertyNamesContractResolver cache


DanStory:
### Source types

```csharp
var data = new Dictionary<string, object>(){
    { "camelCase", "value"},
    { "PascalCase", "Value" },
    { "snake_case", "v_lue" },
    {"YELLING", "VALUE"}
};
```

### destination JSON

```javascript
{
  "camelCase": "value",
  "pascalCase": "Value",
  "snake_case": "v_lue",
  "yelling": "VALUE"
}
// or
 {
  "camelCase": "value",
  "PascalCase": "Value",
  "snake_case": "v_lue",
  "YELLING": "VALUE"
}
```

### Expected behavior

Contract of type is isolated between two different JsonSerializerSettings

### Actual behavior

Contract of type is being shared (with a static cache) across JsonSerializerSettings and serialization 

### Steps to reproduce

```csharp
using System.Collections;
using System.Reflection;
using Newtonsoft.Json;
using Newtonsoft.Json.Serialization;

var data = new Dictionary<string, object>(){
    { "camelCase", "value"},
    { "PascalCase", "Value" },
    { "snake_case", "v_lue" },
    {"YELLING", "VALUE"}
};

var camelSettings = new JsonSerializerSettings()
{
    Formatting = Formatting.Indented,
    ContractResolver = new CamelCasePropertyNamesContractResolver()
        { NamingStrategy = new CamelCaseNamingStrategy() { ProcessDictionaryKeys = true } }
};
var unprocessedSettings = new JsonSerializerSettings()
{
    Formatting = Formatting.Indented,
    ContractResolver = new CamelCasePropertyNamesContractResolver()
        { NamingStrategy = new CamelCaseNamingStrategy() { ProcessDictionaryKeys = false } }
};


Console.WriteLine($"Camel Settings: {JsonConvert.SerializeObject(data, camelSettings)}\n");
Console.WriteLine($"Unprocessed Settings: {JsonConvert.SerializeObject(data, unprocessedSettings)}");

((IDictionary)typeof(CamelCasePropertyNamesContractResolver).GetField("_contractCache", BindingFlags.Static | BindingFlags.NonPublic)?.GetValue(null)!).Clear();
Console.WriteLine("\n-- Forced Clearing of CamelCasePropertyNamesContractResolver internal cache --\n");

[...]

PR #2930, opened 2024-01-23
Ensure CamelCasePropertyNamesContractResolver considers NamingStrategy in the cache

mahdi-ninja:
`CamelCasePropertyNamesContractResolver` uses a shared contract cache across all instances but it doesn't consider the fact that sometimes different instances use different naming strategies. https://github.com/JamesNK/Newtonsoft.Json/issues/2927

In our case, we were using `CamelCasePropertyNamesContractResolver` in our ASP.NET app and then added [Elsa Workflow](https://github.com/elsa-workflows/elsa-core) which was using the same resolver with another naming strategy and broke our existing code. It is now fixed in Elsa Workflow (with this PR https://github.com/elsa-workflows/elsa-core/pull/4521) but it took us a long time to understand what was wrong.


mahdi-ninja:
Hey @JamesNK 
Do you plan to merge this? It feels like a serious bug to me

narmontas:
We are having issues as well and would like this to be merged. @JamesNK

PR #2935, opened 2024-02-15
Author fuzz tests and fix a comment-related issue

Porges:
Add some basic fuzz tests:

- `FuzzDeserialization` tests deserialization
- `FuzzSerialization` tests serialization (using deserialization to create an arbitrary object)
- `FuzzIdempotent` ensures that serialize → deserialize gives back the same object (using deserialization to create an arbitrary object initially). This is more of a "semantic" or property-based test than a pure fuzzer, but it is useful.

The latter test found an issue regarding comments where the line comment `// with a */` would end up being serialized as `/* with a */*/` which broke things. The fix I've implemented for this is to detect this case and serialize it as a line comment. This changes the existing behaviour in another existing unit test.

### DIFF
--- a/Build/build.ps1
+++ b/Build/build.ps1
@@ -133,6 +133,10 @@ task Package -depends Build {
   robocopy $sourceDir $workingDir\Package\Source\Src /MIR /NFL /NDL /NJS /NC /NS /NP /XD bin obj TestResults AppPackages .vs artifacts /XF *.suo *.user *.lock.json | Out-Default
   robocopy $buildDir $workingDir\Package\Source\Build /MIR /NFL /NDL /NJS /NC /NS /NP /XD Temp /XF runbuild.txt | Out-Default
   robocopy $docDir $workingDir\Package\Source\Doc /MIR /NFL /NDL /NJS /NC /NS /NP | Out-Default
+  
+  # include fuzz tests in ADO pipeline artifacts
+  mkdir $workingDir\FuzzTests
+  Copy-Item -Path $sourceDir\Newtonsoft.Json.FuzzTests\bin\Release\net6.0\* -Destination $workingDir\FuzzTests
 
   Compress-Archive -Path $workingDir\Package\* -DestinationPath $workingDir\$zipFileName
 }
--- a/Src/Newtonsoft.Json.FuzzTests/FuzzTests.cs
+++ b/Src/Newtonsoft.Json.FuzzTests/FuzzTests.cs
@@ -1,0 +1,86 @@
+using System;
+using System.IO;
+
+namespace Newtonsoft.Json.FuzzTests {
+
+public static class Fuzzers
+{
+    static readonly JsonSerializer jsonSerializer = new();
+
+    public static void FuzzDeserialization(ReadOnlySpan<byte> buffer)
+    {
+        try
+        {
+            using var ms = new MemoryStream(buffer.ToArray());
+            using var sr = new StreamReader(ms);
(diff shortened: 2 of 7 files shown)


Build/build.ps1
```diff
@@ -133,6 +133,10 @@ task Package -depends Build {
   robocopy $sourceDir $workingDir\Package\Source\Src /MIR /NFL /NDL /NJS /NC /NS /NP /XD bin obj TestResults AppPackages .vs artifacts /XF *.suo *.user *.lock.json | Out-Default
   robocopy $buildDir $workingDir\Package\Source\Build /MIR /NFL /NDL /NJS /NC /NS /NP /XD Temp /XF runbuild.txt | Out-Default
   robocopy $docDir $workingDir\Package\Source\Doc /MIR /NFL /NDL /NJS /NC /NS /NP | Out-Default
+  
+  # include fuzz tests in ADO pipeline artifacts
+  mkdir $workingDir\FuzzTests
```
nimengan, line 138: also, curious where the pipeline changes are made for enabling fuzzing ? are there any yaml file changes somewhere ?

Src/Newtonsoft.Json.Tests/FuzzRegressionTests.cs
```diff
@@ -0,0 +1,127 @@
+#region License
+// Copyright (c) 2007 James Newton-King
```
nimengan, line 2: here as well we have 2007

Build/build.ps1
Porges: This is defined in a separate pipeline which consumes the output of this build pipeline, it makes it easier if we need to change any configuration settings.

Src/Newtonsoft.Json.FuzzTests/FuzzTests.cs
```diff
@@ -0,0 +1,86 @@
+using System;
+using System.IO;
+
+namespace Newtonsoft.Json.FuzzTests {
```
JamesNK, line 4: Standard C# indenting here.

Src/Newtonsoft.Json.Tests/FuzzRegressionTests.cs
```diff
@@ -0,0 +1,127 @@
...
+#endif
+
+#pragma warning disable xUnit1013
+
+namespace Newtonsoft.Json.Tests
+{
+    [TestFixture]
+    public class FuzzRegressionTests : TestFixtureBase
+    {
+        static readonly JsonSerializer jsonSerializer = new();
```
JamesNK, line 43:
```suggestion
        private static readonly JsonSerializer jsonSerializer = new();
```

Src/Newtonsoft.Json/JsonTextWriter.cs
```diff
@@ -784,10 +784,20 @@ public override void WriteValue(Uri? value)
...
-
-            _writer.Write("/*");
-            _writer.Write(text);
-            _writer.Write("*/");
+            
+            // if text contains "*/" then it must have been a line comment
+            if (text != null && text.IndexOf("*/", StringComparison.Ordinal) > -1)
+            {
+                _writer.Write("//");
+                _writer.Write(text);
```
JamesNK, line 792: What happens if text has `*/` and a new line?

Porges: This shouldn't be possible via parsing but if you instantiate a comment type directly you can do it. I'll add some explicit test cases for this.

PR #2956, opened 2024-05-23
Fix comment on JsonConvert.SerializeObject

ev-yu:
Fixes a small mistake in a function comment for JsonConvert.SerializeObject.
I suspect the comment incorrectly copied from an alternative version of the function and not updated to reflect the new parameters.


JamesNK (approved):

JamesNK:
Thanks

PR #2960, opened 2024-06-13
Converter can be null

anjankumar-aptean:
In my case, the converters list had a null value, which caused the application not to continue.
image


anjankumar-aptean:
Please review this PR.

bartelink:
Did you investigate what would be involved in blocking the addition of null converters in the first instance? (I'm thinking that there's probably similar situations where the lib internally walks the converters list assuming there can't be nulls)

PR #2980, opened 2024-09-10
Annotate for trim/AOT (in)compatibility

sbomer:
Adds annotations to propagate trim and AOT compatibility warnings to callers. This way, a trimmed/AOT-compiled app referencing Newtonsoft.Json will get warnings at the places where it calls into Newtonsoft.Json, instead of warnings that point at Newtonsoft.Json internals.

Adds a net8.0 target, because the AOT analyzer requires net7.0+.

I'm not setting up a test app because I don't want to be responsible for some ad-hoc infrastructure or script that does the validation - but I did check this change against a test app locally. I think relying on the analyzer is good enough in this case because the goal isn't to make the code trim-compatible, but to surface existing trim incompatibilities earlier in the development cycle for consumers. If we do miss some warnings it's not a big problem since the problems will still be caught when publishing an app, as they are today.

@eerhardt @agocke @michalstrehovsky @jtschuster


Src/Newtonsoft.Json/JsonTextReader.cs
```diff
@@ -54,6 +56,8 @@ internal enum ReadType
     /// <summary>
     /// Represents a reader that provides fast, non-cached, forward-only access to JSON text data.
     /// </summary>
+    [RequiresUnreferencedCode(MiscellaneousUtils.TrimWarning)]
+    [RequiresDynamicCode(MiscellaneousUtils.AotWarning)]
```
JamesNK, line 60: Why add these annotations to types like JsonTextReader? I don't think it uses reflection or other problem APIs.

Same with reader, Newtonsoft.Json.Linq types, etc.

Src/Newtonsoft.Json/Newtonsoft.Json.csproj
```diff
@@ -40,6 +40,8 @@
     Disable that warning since local builds do not strong-name sign the assembly, causing a mistmatch when comparing
     the previously shipped assemblies and current.-->
     <NoWarn Condition="'$(SignAssembly)' != 'true'">$(NoWarn);CP0003</NoWarn>
+    <IsTrimmable Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))">true</IsTrimmable>
+    <IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">true</IsAotCompatible>
```
eerhardt, line 44: Thinking about this a bit deeper - we really only want the analyzers turned on. We don't want the `IsTrimmable` assembly metadata attribute to be put into the assembly. That is going to break existing apps that work today - like Blazor WASM.

```suggestion
    <EnableSingleFileAnalyzer Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))">true</EnableSingleFileAnalyzer >
    <EnableTrimAnalyzer Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))">true</EnableTrimAnalyzer >
    <EnableAotAnalyzerCondition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">true</EnableAotAnalyzer>
```

Src/Newtonsoft.Json/Newtonsoft.Json.csproj
```diff
@@ -51,6 +53,10 @@
...
..._CONTRACTS;HAVE_DATE_TIME_OFFSET;HAVE_DB_NULL_TYPE_CODE;HAVE_DYNAMIC;HAVE_EMPTY_TYPES;HAVE_ENTITY_FRAMEWORK;HAVE_EXPRESSIONS;HAVE_FAST_REVERSE;HAVE_FSHARP_TYPES;HAVE_FULL_REFLECTION;HAVE_GUID_TRY_PARSE;HAVE_HASH_SET;HAVE_ICLONEABLE;HAVE_ICONVERTIBLE;HAVE_IGNORE_DATA_MEMBER_ATTRIBUTE;HAVE_INOTIFY_COLLECTION_CHANGED;HAVE_INOTIFY_PROPERTY_CHANGING;HAVE_ISET;HAVE_LINQ;HAVE_MEMORY_BARRIER;HAVE_METHOD_IMPL_ATTRIBUTE;HAVE_NON_SERIALIZED_ATTRIBUTE;HAVE_READ_ONLY_COLLECTIONS;HAVE_REFLECTION_EMIT;HAVE_REGEX_TIMEOUTS;HAVE_SECURITY_SAFE_CRITICAL_ATTRIBUTE;HAVE_SERIALIZATION_BINDER_BIND_TO_NAME;HAVE_STREAM_READER_WRITER_CLOSE;HAVE_STRING_JOIN_WITH_ENUMERABLE;HAVE_TIME_SPAN_PARSE_WITH_CULTURE;HAVE_TIME_SPAN_TO_STRING_WITH_CULTURE;HAVE_TIME_ZONE_INFO;HAVE_TRACE_WRITER;HAVE_TYPE_DESCRIPTOR;HAVE_UNICODE_SURROGATE_DETECTION;HAVE_VARIANT_TYPE_PARAMETERS;HAVE_VERSION_TRY_PARSE;HAVE_XLINQ;HAVE_XML_DOCUMENT;HAVE_XML_DOCUMENT_TYPE;HAVE_CONCURRENT_DICTIONARY;HAVE_INDEXOF_STRING_COMPARISON;HAVE_REPLACE_STRING_COMPARISON;HAVE_REPLACE_STRING_COMPARISON;HAVE_GETHASHCODE_STRING_COMPARISON;HAVE_NULLABLE_ATTRIBUTES;HAVE_DYNAMIC_CODE_COMPILED;HAS_ARRAY_EMPTY;HAVE_DATE_ONLY;$(AdditionalConstants)</DefineConstants>
```
eerhardt, line 58: This is a lot of duplication. Can we put this in a single spot and reuse it between `net8` and `net6`?

Src/Newtonsoft.Json/JsonTextReader.cs
JamesNK: I would like to tell people not to use anything to do with the serializer in AOT. But some Newtonsoft.Json things work and don't need warnings on them.

* Reader/writer should be fine in AOT.
* Almost all of JObject, JArray, JValue, etc should almost completely work. Those types have an `ToObject<T>()` method that invokes the serializer and should be annotated individually.

Src/Newtonsoft.Json/Newtonsoft.Json.csproj
JamesNK: Just keep the current standard. I'm not sure I like it anymore, but it would be annoying to change it.

Src/Newtonsoft.Json/Newtonsoft.Json.csproj
eerhardt: > I would have expected trimming the serialization library not to make the problem much worse than it already is.

In a perfect world, this would be true. But I've seen crazier things. I would error on the side of not changing the behavior at all with this change.

Src/Newtonsoft.Json/Utilities/MiscellaneousUtils.cs
```diff
@@ -40,6 +40,37 @@ namespace Newtonsoft.Json.Utilities
...
     {
+        internal const string TrimWarning = "Newtonsoft.Json relies on reflection over types that may be removed when trimming.";
+
+        internal const string AotWarning = "Newtonsoft.Json relies on dynamically creating types that may not be available with Ahead of Time compilation.";
+
+#if HAVE_APPCONTEXT
+        [FeatureSwitchDefinition("Newtonsoft.Json.SerializationIsSupported")]
+        [FeatureGuard(typeof(RequiresUnreferencedCodeAttribute))]
+        [FeatureGuard(typeof(RequiresDynamicCodeAttribute))]
+        internal static bool SerializationIsSupported => AppContext.TryGetSwitch("Newtonsoft.Json.SerializationIsSupported", out bool isSupported) ? isSupported : true;
```
eerhardt, line 51: My suggestion on these names is to name them like public .NET APIs. That way if we ever needed to expose this as a public property, we could match the signature to the name.

For example https://github.com/dotnet/runtime/blob/128fdfd1fd982e277e969d5c7ca1710345f4f649/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializer.Helpers.cs#L27

Src/Newtonsoft.Json/Linq/JObject.cs
```diff
@@ -828,7 +841,15 @@ EventDescriptorCollection ICustomTypeDescriptor.GetEvents()
         /// </returns>
         protected override DynamicMetaObject GetMetaObject(Expression parameter)
         {
+#if HAVE_APPCONTEXT
+            if (!MiscellaneousUtils.DynamicIsSupported)
```
eerhardt, line 845: Should this (and similar) be:

```suggestion
#if HAVE_DYNAMIC
            if (!MiscellaneousUtils.DynamicIsSupported)
```

?

Src/Newtonsoft.Json/Linq/JContainer.cs
```diff
@@ -959,9 +959,17 @@ string ITypedList.GetListName(PropertyDescriptor[] listAccessors)
...
         {
+#if HAVE_APPCONTEXT
+            if (!MiscellaneousUtils.ComponentModelIsSupported)
+            {
+                throw new NotSupportedException(MiscellaneousUtils.ComponentModelNotSupportedMessage);
+            }
+#endif
             ICustomTypeDescriptor? d = First as ICustomTypeDescriptor;
 
+#pragma warning disable IL2026
```
eerhardt, line 970: I don't think these feature switches are going to work like you want them to. What you are doing is disabling the static analysis at build time, so users don't get warnings up front. But they will still get warnings at publish time. And their app will fail at runtime becuase no one defaults `ComponentModelIsSupported = false` in the build.

Src/Newtonsoft.Json/JsonTextReader.cs
JamesNK: Sorry, it's been a while since I've thought about trimming and AOT. I don't know what the feature switches mean. Maybe we should have a call sometime and you can talk through this PR and show me.

Src/Newtonsoft.Json/Utilities/TypeExtensions.cs
```diff
@@ -593,27 +617,16 @@ public static bool AssignableToTypeName(this Type type, string fullTypeName, boo
...
-            {
-                foreach (Type i in type.GetInterfaces())
-                {
-                    if (string.Equals(i.Name, fullTypeName, StringComparison.Ordinal))
-                    {
-                        match = type;
-                        return true;
-                    }
-                }
-            }
```
JamesNK, line 606: Did you remove this because it caused trouble and `searchInterfaces` was always false?

Src/Newtonsoft.Json/Newtonsoft.Json.csproj
```diff
@@ -40,6 +40,11 @@
     Disable that warning since local builds do not strong-name sign the assembly, causing a mistmatch when comparing
     the previously shipped assemblies and current.-->
     <NoWarn Condition="'$(SignAssembly)' != 'true'">$(NoWarn);CP0003</NoWarn>
+    <!-- Enable the analyzers directly instead of setting IsTrimmable or IsAotCompatible, to avoid adding IsTrimmable assembly metadata.
+         This should help avoid breaking consumers who use partial TrimMode. -->
+    <EnableSingleFileAnalyzer Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))">true</EnableSingleFileAnalyzer>
+    <EnableTrimAnalyzer Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))">true</EnableTrimAnalyzer>
+    <EnableAotAnalyzer Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">true</EnableAotAnalyzer>
```
JamesNK, line 47: Are these automatically enabled after .NET 8? It would be useful to have a comment about when they're neccessary so I know what to do when adding new targets.

JamesNK:
@sbomer I pushed some changes to improve JToken writing and ToString.

It looks like they're safe as long as there are no converters. I added new overloads without converters and used from ToString.

JamesNK (approved):

JamesNK:
Looks good to me. Thanks!

PR #2993, opened 2024-10-26
Add support for list patterns

zms9110750:
Add processing for` Index `and `Range` to `JArray.this [object]`


sungam3r:
Please keep indentation to ease review.

JamesNK:
I removed `Range`. Creating a JArray for the result of the range isn't a good pattern. I left in `Index` and added some unit tests.

Thanks for the PR.

Issue #2366, opened 2020-07-20
In which cases JsonConvert.DeserializeObject<T>(string, JsonSerializerSettings) actually returns null?


EnricoMassone:
I'm working with .NET core 3.1, C#8 and nullable reference types enabled.

From the class library I'm writing, I'm referencing the version **12.0.3** of Newtonsoft JSON.

I noticed that by calling [JsonConvert.DeserializeObject\<T\>](https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_JsonConvert_DeserializeObject__1_2.htm) I can get a null reference (Visual Studio analyzers detect a possible dereferencing of a null reference). 

Notice that I'm calling the overload which takes a string and an instance of `JsonSerializerSettings`. I'm only using the `JsonSerializerSettings` in order to handle the possible deserialization errors (via the [Error property](https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonSerializerSettings_Error.htm)).

The github source code confirms that the overload I'm calling can possible return a null reference, via the `MaybeNull` attribute: take a look [here](https://github.com/JamesNK/Newtonsoft.Json/blob/666d9760719e5ec5b2a50046f7dbd6a1267c01c6/Src/Newtonsoft.Json/JsonConvert.cs#L789) for a confirmation.

My question is: in which cases newtonsoft JSON returns a `null` reference when deserializing a JSON string to a .NET type ? 

Usually it returns an object of the given type populated or having its properties at the default value for their type, I have never encountered a case where `null` is returned instead.

TylerBrinkley:
`JsonConvert.DeserializeObject<string>("null")` will return `null`.

EnricoMassone:
@TylerBrinkley thanks for the contibution. 

I noticed that your code actually does not raise any warning inside of visual studio 2019, even with nullable reference types enabled.

The only way to get a warning is calling the other overload, the one taking a second parameter of type `JsonSerializerSetting`.

So, in my opinion, one of the followings is true:

- apart from the mentioned `"null"` input case, there is some other subtle corner case for a `null` reutrn value, involving `JsonSerializerSettings` in some way
- the type annotation for the library is not complete and even the overload of `JsonConvert.DeserializeObject\<T\>` taking only a `string `parameter should be annoted as possibly returning `null`

TylerBrinkley:
[...]

PR #2994, opened 2024-10-30
Added explanation for null in `JsonConvert.DeserializeObject`

rgarlik:
Hello,

As mentioned [here](https://github.com/JamesNK/Newtonsoft.Json/issues/2366#issuecomment-2446719183), the [documentation](https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_JsonConvert_DeserializeObject__1.htm) does not have the correct return type for `DeserializeObject` as it implies that the method can only return a non-nullable type.

Other than adding the explanation of when the method actually returns null, I'd like to ask someone to regenerate the documentation so that the return type is `T?` and not `T`.

### DIFF
--- a/Src/Newtonsoft.Json/JsonConvert.cs
+++ b/Src/Newtonsoft.Json/JsonConvert.cs
@@ -692,7 +692,7 @@ namespace Newtonsoft.Json
         /// Deserializes the JSON to a .NET object.
         /// </summary>
         /// <param name="value">The JSON to deserialize.</param>
-        /// <returns>The deserialized object from the JSON string.</returns>
+        /// <returns>The deserialized object from the JSON string. Null if the provided JSON is valid but represents a null value.</returns>
         [DebuggerStepThrough]
         [RequiresUnreferencedCode(MiscellaneousUtils.TrimWarning)]
         [RequiresDynamicCode(MiscellaneousUtils.AotWarning)]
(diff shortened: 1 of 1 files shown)


Src/Newtonsoft.Json/JsonConvert.cs
```diff
@@ -812,7 +812,7 @@ private static string SerializeObjectInternal(object? value, Type? type, JsonSer
         /// The <see cref="JsonSerializerSettings"/> used to deserialize the object.
         /// If this is <c>null</c>, default serialization settings will be used.
         /// </param>
-        /// <returns>The deserialized object from the JSON string.</returns>
+        /// <returns>The deserialized object from the JSON string. A <see cref="null"/> value is returned if the provided JSON is valid but represents a null value.</returns>
```
JamesNK, line 815:
```suggestion
        /// <returns>The deserialized object from the JSON string. A <see langword="null"/> value is returned if the provided JSON is valid but represents a null value.</returns>
```

Src/Newtonsoft.Json/JsonConvert.cs
```diff
@@ -827,7 +827,7 @@ private static string SerializeObjectInternal(object? value, Type? type, JsonSer
         /// <param name="value">The JSON to deserialize.</param>
         /// <param name="type">The type of the object to deserialize.</param>
         /// <param name="converters">Converters to use while deserializing.</param>
-        /// <returns>The deserialized object from the JSON string.</returns>
+        /// <returns>The deserialized object from the JSON string. A <see cref="null"/> value is returned if the provided JSON is valid but represents a null value.</returns>
```
JamesNK, line 830:
```suggestion
        /// <returns>The deserialized object from the JSON string. A <see langword="null"/> value is returned if the provided JSON is valid but represents a null value.</returns>
```

Src/Newtonsoft.Json/JsonConvert.cs
```diff
@@ -849,7 +849,7 @@ private static string SerializeObjectInternal(object? value, Type? type, JsonSer
         /// The <see cref="JsonSerializerSettings"/> used to deserialize the object.
         /// If this is <c>null</c>, default serialization settings will be used.
         /// </param>
-        /// <returns>The deserialized object from the JSON string.</returns>
+        /// <returns>The deserialized object from the JSON string. A <see cref="null"/> value is returned if the provided JSON is valid but represents a null value.</returns>
```
JamesNK, line 852:
```suggestion
        /// <returns>The deserialized object from the JSON string. A <see langword="null"/> value is returned if the provided JSON is valid but represents a null value.</returns>
```

Issue #2054, opened 2019-04-22
It seems a memory leak when using Newtonsoft.Json for dotnet core.


dickence:
I have published a dotnet core webapi project to IIS. But the memory is increasing fast. I checked the memory by diagnostic tools in Visual Studio. This is the result:

![屏幕快照 2019-04-22 22 09 21](https://user-images.githubusercontent.com/9348495/56504531-5e445580-654b-11e9-88e2-2cf177e4e48b.png)

I called the same api twice with the same parameters, the results of memory increasement are similar. It seems that memory about newtonsoft.json is increasing fastest, and never goes down. Is it a bug? Or do I use newtonsoft.json in a wrong way?

fschlaef:
I have the exact same issue, but it doesn't seem that the problem lies within Newtonsoft.Json, but rather how the .Net Core runtime handles large objects in your case, probably `retlist`.

Many objects stay in memory with no way to release them until the process memory passes a threshold.

I would say nothing to worry here, unless you get OutOfMemoryException.

joetherod:
nothing to worry here? If its a memory leak issue in core, it should be resolved. Seems that core has very poor memory management. I ported  over an existing mvc app to core and have these ' memory issues' that dont seem to get resolved.

dcoen:
We are having this issue too.  I suspect it is from the StringWriter used in SerializeObjectInternal NOT being disposed.  That implements IDisposable and should be placed in a using() block.

![image](https://user-images.githubusercontent.com/36668631/74948901-efce5f00-53ca-11ea-8df1-3259bd9ff57b.png)
![image](https://user-images.githubusercontent.com/36668631/74948925-f8269a00-53ca-11ea-857c-b08c07d91759.png)
![image](https://user-images.githubusercontent.com/36668631/74948996-18eeef80-53cb-11ea-9d9b-6866cfcb1475.png)

dcoen:
You all can probably ignore my earlier comment....i think this instead an issue with Microsoft's Durable Task Framework trying to serialize an Exception (presumably with a self referencing loop) using Newtonsoft.  Specifically, DurableTask.Core.Serializing.JsonDataConverter.

AlexanderKot:
Hello
Seems having the same problem. 
.net core SDK version: V6.0.7
newtonsoft.json version: V13.0.1

I also trying serialize object containing **inside List < dynamic >**, and this can be source of problem.
doing this - have many leaked object related too newtonsoft.

For example
[...]

PR #3022, opened 2025-02-24
Fix StringWriter dispose #2054

bars229:
Add disposing pattern for StringWriter to fix issue #2054 or at least improve memory usage.


JamesNK:
Dispose doesn't do anything in these types.

bars229:
Sorry, you're right. I've checked the dispose methods, they aren't do anything meaningful.

Issue #3055, opened 2025-05-27
Newtonsoft.Json version 13.0.3 missing json:Array attribute when using JsonConvert.DeserializeXNode() on properties with special characters


drpdrp:
Newtonsoft.Json version 13.0.3 (current)
.NET 9 (current)

BUG:
When using DeserializeXNode with arguments **writeArrayAttribute** and **encodeSpecialCharacters** set to **true**, the array attribute is not set for properties with special characters.

SAMPLE:
[DotNetFiddeSample](https://dotnetfiddle.net/gzvPjO)

CODE:
`		var json = """
			{
				"test": ["CORRECT - contains jsonArray attribute"],
				"test#specialchar": ["BUG - missing jsonArray attribute"],
				"test$specialchar": ["BUG - missing jsonArray attribute"],
				"test%specialchar": ["BUG - missing jsonArray attribute"],
				"test:specialchar": ["BUG - missing jsonArray attribute"],
			}
			""";
		
        var xDocument = JsonConvert.DeserializeXNode(json, "root", writeArrayAttribute: true, encodeSpecialCharacters: true);
		Console.WriteLine(xDocument.ToString());`

RESULT:
`<root>
  <test json:Array="true" xmlns:json="http://james.newtonking.com/projects/json">CORRECT - contains jsonArray attribute</test>
  <test_x0023_specialchar>BUG - missing jsonArray attribute</test_x0023_specialchar>
  <test_x0024_specialchar>BUG - missing jsonArray attribute</test_x0024_specialchar>
  <test_x0025_specialchar>BUG - missing jsonArray attribute</test_x0025_specialchar>
  <test_x003A_specialchar>BUG - missing jsonArray attribute</test_x003A_specialchar>
</root>`

PR #3060, opened 2025-05-30
Add json:Array attribute on properties with special characters (JSON to XML conversion) #3055

drpdrp:
Please refer to https://github.com/JamesNK/Newtonsoft.Json/issues/3055


JamesNK:
Thanks for the fix and test!

Issue #3056, opened 2025-05-28
Duplicate attribute error on nested arrays with writeArrayAttribute set to true


drpdrp:
Newtonsoft.Json version 13.0.3 (current)
.NET 9 (current)

BUG:
When using DeserializeXNode with arguments **writeArrayAttribute** set to **true** and nested arrays, the conversion fails on System.InvalidOperationException: Duplicate attribute.

SAMPLE:
[DotNetFiddleSample](https://dotnetfiddle.net/haZfJs)

CODE:
```
var json = """
	{
		"test": [["nested array"]],
	}
	""";

var xDocument = JsonConvert.DeserializeXNode(json, "root", true);
Console.WriteLine(xDocument.ToString());
```

EXPECTED:
```
<root>
  <test json:Array="true" xmlns:json="http://james.newtonking.com/projects/json">
    <test json:Array="true" xmlns:json="http://james.newtonking.com/projects/json">nested array</test>
  </test>
</root>
```

ACTUAL:
```
Unhandled exception. System.InvalidOperationException: Duplicate attribute.
   at System.Xml.Linq.XElement.AddAttributeSkipNotify(XAttribute a)
   at Newtonsoft.Json.Converters.XElementWrapper.SetAttributeNode(IXmlNode attribute)
   at Newtonsoft.Json.Converters.XmlNodeConverter.AddJsonArrayAttribute(IXmlElement element, IXmlDocument document)
   at Newtonsoft.Json.Converters.XmlNodeConverter.DeserializeNode(JsonReader reader, IXmlDocument document, XmlNamespaceManager manager, IXmlNode currentNode)
   at Newtonsoft.Json.Converters.XmlNodeConverter.CreateElement(JsonReader reader, IXmlDocument document, IXmlNode currentNode, String elementName, XmlNamespaceManager manager, String elementPrefix, Dictionary`2 attributeNameValues)
   at Newtonsoft.Json.Converters.XmlNodeConverter.ReadElement(JsonReader reader, IXmlDocument document, IXmlNode currentNode, String propertyName, XmlNamespaceManager manager)
   at Newtonsoft.Json.Converters.XmlNodeConverter.ReadJson(JsonReader reader, Type objectType, Object existingValue, JsonSerializer serializer)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.DeserializeConvertable(JsonConverter converter, JsonReader reader, Type objectType, Object existingValue)
[...]

PR #3062, opened 2025-05-30
Alternate fix duplicate attribute exception on nested arrays with writeArrayAttribute set to true

drpdrp:
Please refer to https://github.com/JamesNK/Newtonsoft.Json/issues/3056

This alternate fix is less verbose but also less performant than #3061 version


JamesNK:
Thanks for the fix and tests!

Issue #972, opened 2016-07-26
Type name info is not saved for F# unions with TypeNameHandling.All


alfonsogarciacaro:
Hi there. When I try to serialise an F# union with settings `TypeNameHandling=TypeNameHandling.All` the type name info is not saved as with other types:

```fsharp
type Tree =
    | Leaf of int
    | Branch of Tree[]

let x = Leaf 5
let json = JsonConvert.SerializeObject(x,
                   JsonSerializerSettings(TypeNameHandling=TypeNameHandling.All))

// json = """{"Case":"Leaf","Fields":[5]}"""
```

With F# records and classes, the type name is correctly saved.

I'm using Newtonsoft.Json v9.0.1, please tell me if you need any other info.

Cheers!

PR #3063, opened 2025-06-04
Support $type property for F# discriminated unions

NatElkins:
Fixes https://github.com/JamesNK/Newtonsoft.Json/issues/972 for both reading and writing.

Does the following:

1. When writing F# discriminated unions and when `TypeNameHandling` is set to `All`, it will write `$type`.
2. However, if `PreserveReferencesHandling` is set to `All`, it won't write it.
3. A modification has been made to `JsonSerializerInternalReader` where if there a `$type`, we check to see if there's a converter for it.  The existing guidance is that converters must handle all properties (as detailed [here](https://stackoverflow.com/a/79149083)).

Tests have been added to show that both `JsonConvert.DeserializeObject<Shape>` and `JsonConvert.DeserializeObject` work (the latter only when $type has been filled in).

This is strictly an improvement on the existing logic.


NatElkins:
Hm, the tests pass locally, not sure why it fails in CI.  Wonder if it's possible to trigger CI again.

PR #3081, opened 2025-09-22
Fix error message displayed when a property marked as DisallowNull is set to null

whippet71:
The docs state that if a property is *not* required but also cannot be a null value, then we can mark it with (Required = Required.DisallowNull). However if the property is set to null, the error message displayed is "Required property '{propName}' expects a non-null value", which is incorrect, as the property is not required.

The fix is to change the error message in JsonSerializerInternalReader.cs, and to update JsonSerializerTests.cs accordingly.


JamesNK (approved):

JamesNK:
Thanks!

PR #3093, opened 2025-12-30
Reduce allocations in ToMultidimensionalArray

alexzzzs:
Optimizes the `ToMultidimensionalArray` method to reduce memory allocations when deserializing multidimensional arrays.

### The Issue
The previous implementation of `CopyFromJaggedToMultidimensionalArray` created a new `int[]` array at every step of the recursion to store indices. For large multidimensional arrays (e.g., 3D grids), this resulted in O(N) intermediate array allocations, creating unnecessary GC pressure.

### The Fix
- Replaced the recursive array creation with a single `int[]` buffer allocated once at the entry point.
- This buffer is passed down the recursion stack and updated in-place.
- Updated `JaggedArrayGetValue` to accept a length parameter, allowing it to read from the shared buffer without needing a trimmed copy.

### DIFF
--- a/Src/Newtonsoft.Json/Utilities/CollectionUtils.cs
+++ b/Src/Newtonsoft.Json/Utilities/CollectionUtils.cs
@@ -314,43 +314,36 @@ namespace Newtonsoft.Json.Utilities
             return dimensions;
         }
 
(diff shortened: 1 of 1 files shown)


JamesNK:
Thanks!

Issue #2869, opened 2023-06-28
Divide by zero exception in Newtonsoft.Json.Utilities.BoxedPrimitives.Get() method


aurax:
### Source/destination types

```csharp
[Serializable]
internal class ImportData {
        public string FilePath { get; set; }
        public double Scale { get; set; }
        public Point Origin { get; set; }
        public string Name { get; set; }
}
[Serializable]
public sealed class Point {
        [JsonConstructor]
        public Point(double x, double y, double z) {
            _x = x;
            _y = y;
            _z = z;
        }
        private readonly double _x;
        private readonly double _y;
        private readonly double _z;

        public double X => _x;
        public double Y => _y;
        public double Z => _z;
}
```

### Source/destination JSON

```javascript
{ "aaa": {
        "FilePath":"file:///C:/temp/aaa.txt",
        "Scale":2.0,
        "Origin": {
            "X":0.0,
            "Y":0.0,
            "Z":0.0
        },
        "Name":"aaa"
    }
}
```

### Expected behavior

Version 13.0.2 correctly deserializes this.

### Actual behavior

System.DivideByZeroException : Attempted to divide by zero in Newtonsoft.Json\Utilities\BoxedPrimitives.cs, line 135

![image](https://github.com/JamesNK/Newtonsoft.Json/assets/1779282/8cb4c863-ab84-4401-9f1e-2792cf6ef2b2)

JamesNK:
I couldn't replicate this in a test. See below.

Also, I don't see how this error is possible. You can't get `DivideByZeroException` when dividing by a zero double. The result is either positive or negative infinity.
```cs
        [Test]
        public void DeserializeZeroIssue()
        {
            var json = @"{ ""aaa"": {
        ""FilePath"":""file:///C:/temp/aaa.txt"",
        ""Scale"":2.0,
        ""Origin"": {
            ""X"":0.0,
            ""Y"":0.0,
            ""Z"":0.0
        },
        ""Name"":""aaa""
    }
}";

            JsonConvert.DeserializeObject<ImportData>(json);
        }

        [Serializable]
        internal class ImportData
        {
            public string FilePath { get; set; }
            public double Scale { get; set; }
            public Point Origin { get; set; }
            public string Name { get; set; }
        }
        [Serializable]
        public sealed class Point
        {
            [JsonConstructor]
            public Point(double x, double y, double z)
            {
                _x = x;
                _y = y;
                _z = z;
            }
[...]

PR #3094, opened 2026-01-19
Use a safer method for detecting negative zero

daveorourke:
Fixes https://github.com/JamesNK/Newtonsoft.Json/issues/2869.

The prior method would throw a DivideByZeroException on systems where the `_EM_ZERODIVIDE` floating point exception was enabled.

The changes in this PR are just a safer implementation of the changes in https://github.com/JamesNK/Newtonsoft.Json/pull/2777 and are covered by those same tests.

Note: I tried to add a test for https://github.com/JamesNK/Newtonsoft.Json/issues/2869, but changing the `_EM_ZERODIVIDE` flag with [_controlfp](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/control87-controlfp-control87-2?view=msvc-170) was affecting other tests due to the changing of global state and the parallel execution of the tests.  I tried several different test methods before giving up on using [_controlfp](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/control87-controlfp-control87-2?view=msvc-170).  I also didn't see any other use of interop (P/Invoke) in the tests, so I wasn't sure you'd accept that anyway.

Even without the additional test, I think these changes are very straightforward and would resolve the above issue, preserving all existing functionality.

### DIFF
--- a/Src/Newtonsoft.Json/Utilities/BoxedPrimitives.cs
+++ b/Src/Newtonsoft.Json/Utilities/BoxedPrimitives.cs
@@ -99,7 +99,7 @@ namespace Newtonsoft.Json.Utilities
 #if NET6_0_OR_GREATER
             // Number of bits scale is shifted by.
             const int ScaleShift = 16;
-        
+
             if (value == decimal.Zero)
             {
                 Span<int> bits = stackalloc int[4];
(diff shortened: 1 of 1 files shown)


JamesNK:
Thanks!

PR #3111, opened 2026-05-31
Validate $type metadata values during deserialization

### LINKED ISSUE #3109: NullReferenceException in JsonSerializerInternalReader.ReadMetadataProperties for $type value that isn't a string or null
## Description

When Json.NET deserializes a JSON object whose `$type` metadata property is followed by any token whose `reader.Value` is `null`, the deserializer throws `System.NullReferenceException` instead of a `JsonSerializationException`. In practice this covers structural tokens (`[`, `{`, malformed placeholders like a bare `,`) **and also the literal JSON `null`** (`JsonToken.Null` has `Value == null` by design). Non-null values such as `42` or `true` parse fine - `ToString()` succeeds and the code proceeds to resolve the string as a type name.

The `$ref` path in the same method has a token-type guard that produces a clean exception; the `$id` path uses a null-conditional `?.`; only the `$type` path dereferences `reader.Value!` directly and NREs.

The sibling `$ref` handling (line 724) correctly guards with:

```csharp
if (reader.TokenType != JsonToken.String && reader.TokenType != JsonToken.Null)
{
    throw JsonSerializationException.Create(reader, "JSON reference {0} property must have a string or null value.".FormatWith(...));
}
```

The `$type` handling (line 757) has no such guard and assumes `reader.Value` is non-null.

## JSON that reproduces the issue

Any of the following small inputs trigger the NRE:

```
{"$type":,            (11 bytes - malformed, comma after colon)
{"$type":[            (10 bytes - StartArray token, Value=null)
{"$type":{            (10 bytes - StartObject token, Value=null)
{"$type":null}        (14 bytes - well-formed JSON, Null token with Value=null)
```

All four land on `reader.Value!.ToString()!` at `JsonSerializerInternalReader.cs:line 757` (confirmed via a stack trace with source-link line numbers) and NRE because `reader.Value` is null. The `{"$type":null}` case is the notable one: it is well-formed JSON and a user could legitimately serialize it, yet it crashes the deserializer.

Non-null primitive values (`{"$type":42}`, `{"$type":true}`) do not trigger - `ToString()` succeeds and the call proceeds to `ResolveTypeName` with the stringified value.

## Json.NET code that reproduces it

```csharp
using Newtonsoft.Json;

class Target { }

try
{
    JsonConvert.DeserializeObject<Target>("{\"$type\":,");
}
catch (Exception ex)
{
    Console.WriteLine(ex.GetType().FullName + ": " + ex.Message);
    // Prints: System.NullReferenceException: Object reference not set to an instance of an object.
}
```

The bug reproduces for both typed (`DeserializeObject<Target>`) and weakly-typed (`DeserializeObject<Dictionary<string,object>>`, `DeserializeObject(input)`) deserialization, because `ReadMetadataProperties` runs in `CreateObject` regardless of the target type. `JToken.Parse` is not affected - the DOM parser doesn't process metadata properties. The bug is specific to `MetadataPropertyHandling.Default` (the library default): setting `MetadataPropertyHandling.ReadAhead` surfaces a clean `JsonReaderException`, and `MetadataPropertyHandling.Ignore` surfaces a `JsonSerializationException`. Both are viable workarounds for callers who can't wait for a fix.

## Exception message and stack trace

```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.ReadMetadataProperties(
        JsonReader reader, Type& objectType, JsonContract& contract, JsonProperty member,
        JsonContainerContract containerContract, JsonProperty containerMember,
        Object existingValue, Object& newValue, String& id)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(
        JsonReader reader, Type objectType, JsonContract contract, JsonProperty member,
        JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(...)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader, Type, Boolean)
   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader, Type)
   at Newtonsoft.Json.JsonSerializer.Deserialize(JsonReader, Type)
   at Newtonsoft.Json.JsonConvert.DeserializeObject(String, Type, JsonSerializerSettings)
```

## Affected versions

`Newtonsoft.Json` 13.0.4 (latest stable NuGet release), `13.0.5-beta1` (pre-release - tested), and `master` at the most recent commit. All three reproduce. The source at `Src/Newtonsoft.Json/Serialization/JsonSerializerInternalReader.cs` line 757 is `string qualifiedTypeName = reader.Value!.ToString()!;` and is unchanged since 13.0.4.

## Expected behavior

Match the existing behavior of the `$ref` branch - throw a `JsonSerializationException` with a message like *`"$type" property must have a string or null value.`* rather than a raw `NullReferenceException`. Callers that already wrap deserialization in `catch (JsonException)` / `catch (JsonSerializationException)` would then handle the bad input cleanly instead of an uncategorized NRE.

## Discovery

Coverage-guided fuzzing (AFL++ 4.00c + SharpFuzz 2.2.0) against `Newtonsoft.Json 13.0.4` on .NET 10.

BikcromBiswas:
Fixes  https://github.com/JamesNK/Newtonsoft.Json/issues/3109
 Summary
- Validate `$type` metadata values before resolving type names during deserialization.
- Throw a `JsonSerializationException` when `$type` is not a string.
- Add unit tests for null, array, object, and integer `$type` values.
- 
Tests
- `dotnet test Src/Newtonsoft.Json.Tests/Newtonsoft.Json.Tests.csproj -f net8.0 --filter FullyQualifiedName~TypeNameHandlingTests`

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Serialization/TypeNameHandlingTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Serialization/TypeNameHandlingTests.cs
@@ -724,6 +724,42 @@ namespace Newtonsoft.Json.Tests.Serialization
         }
 
         [Test]
+        public void DeserializeWithNullTypeProperty_ThrowsJsonSerializationException()
+        {
+            string json = "{\"$type\":null,\"Name\":\"test\"}";
+
+            ExceptionAssert.Throws<JsonSerializationException>(() =>
+                JsonConvert.DeserializeObject<Dictionary<string, object>>(json));
+        }
+
+        [Test]
+        public void DeserializeWithArrayTypeProperty_ThrowsJsonSerializationException()
+        {
+            string json = "{\"$type\":[\"foo\"],\"Name\":\"test\"}";
+
+            ExceptionAssert.Throws<JsonSerializationException>(() =>
+                JsonConvert.DeserializeObject<Dictionary<string, object>>(json));
+        }
+
+        [Test]
+        public void DeserializeWithObjectTypeProperty_ThrowsJsonSerializationException()
+        {
+            string json = "{\"$type\":{},\"Name\":\"test\"}";
+
+            ExceptionAssert.Throws<JsonSerializationException>(() =>
+                JsonConvert.DeserializeObject<Dictionary<string, object>>(json));
+        }
+
+        [Test]
+        public void DeserializeWithIntegerTypeProperty_ThrowsJsonSerializationException()
+        {
+            string json = "{\"$type\":42,\"Name\":\"test\"}";
+
+            ExceptionAssert.Throws<JsonSerializationException>(() =>
+                JsonConvert.DeserializeObject<Dictionary<string, object>>(json));
+        }
+
+        [Test]
         public void DeserializeTypeNameWithNoTypeNameHandling()
         {
             string employeeRef = typeof(EmployeeReference).AssemblyQualifiedName;
--- a/Src/Newtonsoft.Json/Serialization/JsonSerializerInternalReader.cs
+++ b/Src/Newtonsoft.Json/Serialization/JsonSerializerInternalReader.cs
@@ -489,106 +489,106 @@ namespace Newtonsoft.Json.Serialization
             switch (contract.ContractType)
             {
                 case JsonContractType.Object:
+                {
+                    bool createdFromNonDefaultCreator = false;
+                    JsonObjectContract objectContract = (JsonObjectContract)contract;
+                    object targetObject;
+                    // check that if type name handling is being used that the existing value is compatible with the specified type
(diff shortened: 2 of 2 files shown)


BikcromBiswas:
Hi @JamesNK 

I've submitted this PR to fix a NullReferenceException that occurs when the $type metadata property has a null 
or non-string value.

Could you please review this when you get a chance? I'd really appreciate any feedback or suggestions if 
something needs to be changed or improved.

Thank you!

JamesNK:
Thanks!

PR #3114, opened 2026-07-06
Check conversion operators to determine if `objectType` is assignable from `byte[]`

olfek:
https://learn.microsoft.com/dotnet/csharp/language-reference/operators/user-defined-conversion-operators

For when `objectType` is `Memory<byte>` for example.


olfek:
@JamesNK --- any good?

Issue #1199, opened 2017-02-01
DefaultValueHandling.Populate doesn't set default value when deserializing


ychuzevi:
While migrating from JSON.Net 6 to latest version (9.0.1), I noticed a regression regarding how DefaultValueHandling.Populate & DefaultValueHandling.IgnoreAndPopulate are handled during deserialization. With latest version, the Json property is initialized to null instead of its default value after deserialization.

Here is a simple test to reproduce the issue: 

```
        private class MyTestClass
        {
            public const string DefaultText = "...";

            [DefaultValue(DefaultText)]
            [JsonProperty(PropertyName = "myText", DefaultValueHandling = DefaultValueHandling.Populate)]
            public readonly string Text;

            public MyTestClass(string text = DefaultText)
            {
                Text = text;
            }
        }

        [Test]
        public void DumbTest()
        {
            MyTestClass myObject = JsonConvert.DeserializeObject<MyTestClass>("{}"); // Fail with version 9.0.1
            Assert.AreEqual(MyTestClass.DefaultText, myObject.Text);
        }
```

I tried with an int property and the default value was also ignored.

ychuzevi:
It seems the issue is that if the JsonConverter finds a parameter's name in the constructor which matches one of the property, then it will consider it initialized by the constructor and will not apply the property's "default value" rules.

For example if I rename either MyTestClass's constructor "test" parameter or MyTestClass "Text" property then it will work fine.

It is often useful to have a constructor if you want to create the object from another source than JSON. And when initializing the properties from the constructor it is a good practice to use as argument's name the property name in lower case.

dbc2:
There's some discussion of the issue on stackoverflow, [here](http://stackoverflow.com/a/42009289/3744182).  

 It appears that [`JsonSerializerInternalReader.CreateObjectUsingCreatorWithParameters()`](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Serialization/JsonSerializerInternalReader.cs#L1979) now checks for the presence of the `[DefaultValue(DefaultText)]` and `[JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]` attributes on the *constructor parameter* rather than the corresponding property.  This check is made around line 1979:

    if (HasFlag(constructorProperty.DefaultValueHandling.GetValueOrDefault(Serializer._defaultValueHandling), DefaultValueHandling.Populate))
    {
        context.Value = EnsureType(
[...]

PR #3120, opened 2026-08-12
Fix default value population for renamed constructor parameters

### LINKED ISSUE #3119: [DefaultValue] applied to a constructor parameter does not work when the parameter has also been renamed by setting JsonPropertyAttribute.PropertyName
**Description of bug**

Consider the following two immutable types:

	public class Test1
	{
		[JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate), DefaultValue("default value")]
		public string Value { get; }

		public Test1([JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate), DefaultValue("default value")]string value) => 
			this.Value = value;
	}

	public class Test2
	{
		[JsonProperty("renamedValue", DefaultValueHandling = DefaultValueHandling.Populate), DefaultValue("default value")]
		public string Value { get; }

		public Test2([JsonProperty("renamedValue", DefaultValueHandling = DefaultValueHandling.Populate), DefaultValue("default value")]string value) => 
			this.Value = value;
	}

Both types have a single read-only property `Value` that is set in their single constructor.  And both types have a default value applied to the constructor parameter as well as to the property.  (I am aware that, as explained in [Issue 1199](https://github.com/JamesNK/Newtonsoft.Json/issues/1199), the default value attribute must needs be applied to the constructor parameter.)  But in addition the second type renames the property (and constructor parameter, for consistency) by setting `JsonPropertyAttribute.PropertyName`.

Now imagine I try to deserialize an empty JSON object `{}` to both types like so, with the expectation that the default value will be supplied to the constructor:

	var test1 = JsonConvert.DeserializeObject<Test1>("{}");
	Console.WriteLine("De-serialized and re-serialized {0}: {1}", test1, JsonConvert.SerializeObject(test1)); // Prints {"Value":"default value"}

	var test2 = JsonConvert.DeserializeObject<Test2>("{}");
	Console.WriteLine("De-serialized and re-serialized {0}: {1}\n", test2, JsonConvert.SerializeObject(test2)); // Prints {"renamedValue":null}

	Assert.That(test1?.Value == "default value"); // Passes
	Assert.That(test2?.Value == "default value"); // Fails!

`test1?.Value` is set to `"default value"` as expected -- but `test2?.Value` is null, and the assertion fails.

Since the only difference between the two types is that the property and constructor parameter were renamed, setting the default value should work in both cases.

Demo Json.NET 13.0.4 fiddle [here](https://dotnetfiddle.net/49pQL0).

**Analysis**

The problem seems to be caused (at least partially) by a bug in `JsonSerializerInternalReader.CreateObjectUsingCreatorWithParameters()` around [line 2060](https://github.com/JamesNK/Newtonsoft.Json/blob/4f73e74372445108d2c1bda37b36e6f5e43402e0/Src/Newtonsoft.Json/Serialization/JsonSerializerInternalReader.cs#L2060):

    if (constructorProperty == null && context.Property != null)
    {
        constructorProperty = contract.CreatorParameters.ForgivingCaseSensitiveFind(p => p.PropertyName!, context.Property.UnderlyingName!);
    }

What this code tries to do is to match `p.PropertyName` with `context.Property.UnderlyingName` -- but this looks wrong.  `PropertyName` is the **renamed** name where as `UnderlyingName` is the **reflected** name.  It should use one or the other consistently:

      constructorProperty = contract.CreatorParameters.ForgivingCaseSensitiveFind(p => p.UnderlyingName!, context.Property.UnderlyingName!);

or

      constructorProperty = contract.CreatorParameters.ForgivingCaseSensitiveFind(p => p.PropertyName!, context.Property.PropertyName!);
  
I haven't forked Json.NET and rebuilt it for confirmation, but I suspect that either fix should resolve this issue.  However, I'm not sure which fix would be correct.

**Update**

If I only set `JsonPropertyAttribute.PropertyName` on the property and not the constructor (while still setting `JsonPropertyAttribute.DefaultValueHandling = DefaultValueHandling.Populate` on both) then deserialization seems to work:

	public class Test3
	{
		[JsonProperty("renamedValue", DefaultValueHandling = DefaultValueHandling.Populate), DefaultValue("default value")]
		public string Value { get; }

		public Test3([JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate), DefaultValue("default value")]string value) => 
			this.Value = value;
	}

	var test3= JsonConvert.DeserializeObject<Test3>("{}");
	Assert.That(test3?.Value == "default value"); // Also Passes

Is this expected?  It does contradict what @JamesNK  wrote in [DefaultValueHandling.Populate doesn't set default value when deserializing #1199](https://github.com/JamesNK/Newtonsoft.Json/issues/1199#issuecomment-277602750)

> That is intended behavior. The constructor param name is `text` and the property name is `myText`. Place an attribute on the param to change its name so the two are linked up.

If I modify the example from that comment to use a read-only property rather than a `readonly` field, then deserialization fails:

```
var test1 = JsonConvert.DeserializeObject<MyTestClass>("{}");
Console.WriteLine("De-serialized and re-serialized {0}: {1}", test1, JsonConvert.SerializeObject(test1)); // Prints {"Value":"default value"}

Assert.That(test1?.Text == MyTestClass.DefaultText); // Fails

class MyTestClass
{
    public const string DefaultText = "...";

    [DefaultValue(DefaultText)]
    [JsonProperty(PropertyName = "myText", DefaultValueHandling = DefaultValueHandling.Populate)]
    public string Text { get; }

    public MyTestClass([JsonProperty(PropertyName = "myText")]string text = DefaultText)
    {
        Text = text;
    }
}
```

Demo fiddle 2 [here](https://dotnetfiddle.net/FuDt6r).

HengYpinn:
Fixes #3119.

Related context: #1199.

## Summary

- Resolve an explicitly named constructor parameter with the same naming rules used for properties before matching it to a member.
- Match missing creator values by resolved JSON name so default-value population reaches the linked constructor parameter.
- Add regression coverage for get-only properties and for the intentionally unmatched behavior documented in #1199.

## Root cause

Constructor parameters were initially matched to members before the parameter's `JsonPropertyAttribute.PropertyName` was resolved. Later, missing-value processing compared a resolved constructor JSON name with an underlying CLR member name. Consequently, explicitly assigning the same renamed JSON name to a member and constructor parameter did not reliably link them or inherit the member's default settings.

## Compatibility

This preserves the behavior documented in #1199: a constructor parameter named `text` is not linked to a member serialized as `myText` unless the parameter explicitly uses `[JsonProperty("myText")]` (or is itself named `myText`). The fix does not add a broad CLR-name fallback.

## User impact

Immutable models using explicit `JsonProperty` renames with `DefaultValueHandling.Populate` now receive their configured constructor default during deserialization instead of `null`.

## Validation

- `dotnet test Src/Newtonsoft.Json.Tests/Newtonsoft.Json.Tests.csproj -f net8.0` (3,519 passed)
- `dotnet test Src/Newtonsoft.Json.Tests/Newtonsoft.Json.Tests.csproj -f net6.0` (3,515 passed)
- `dotnet build Src/Newtonsoft.Json/Newtonsoft.Json.csproj -c Release` (all library target frameworks built successfully)


JamesNK:
Thanks!

PR #3123, opened 2026-08-12
Fix quadratic JObject duplicate property handling

JamesNK:
## Summary

- use lazy, operation-scoped property indexes when replacing duplicate properties during `JObject` loading
- apply the same indexed replacement to `JTokenWriter`, including nested-object cleanup and stale-index recovery for externally mutated target containers
- add deterministic sync and async regressions that fail if duplicate replacement searches the property list, plus nested, ordering, and serializer coverage

Duplicate-heavy input previously performed a linear property-list search for each replacement. The new path builds an index only after the first duplicate and replaces subsequent properties by their known position, preserving last-value-wins behavior and the original property position.

## Validation

- `dotnet test .\Src\Newtonsoft.Json.Tests\Newtonsoft.Json.Tests.csproj -f net8.0 --no-build --no-restore`
  - 3,524 passed, 0 failed
- focused duplicate loading tests: 10 passed
- focused duplicate writer tests: 4 passed
- `git diff --check`

Freshly loaded rebuilt DLL, duplicate attack versus same-token-count unique control:

| Path | 15K attack | 15K control | 30K attack | 30K control |
|---|---:|---:|---:|---:|
| `JObject.Parse` | 28-31 ms | 26 ms | 72-79 ms | 62-65 ms |
| `JsonConvert.DeserializeObject<JObject>` | 30 ms | 27-32 ms | 60-62 ms | 56-61 ms |

The attack cases also verified final values of `14999` and `29999`, confirming last-value-wins semantics.

### DIFF
--- a/Src/Newtonsoft.Json.Tests/Linq/JObjectAsyncTests.cs
+++ b/Src/Newtonsoft.Json.Tests/Linq/JObjectAsyncTests.cs
@@ -25,6 +25,7 @@
 
 #if !(NET20 || NET35 || NET40 || PORTABLE40)
 
+using System;
 using System.Collections.Generic;
 using Newtonsoft.Json.Tests.TestObjects;
 #if DNXCORE50
@@ -36,6 +37,7 @@ using NUnit.Framework;
 #endif
 using Newtonsoft.Json.Linq;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 
 namespace Newtonsoft.Json.Tests.Linq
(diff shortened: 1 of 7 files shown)


JamesNK:
Reviewed at `7ad27785`. No blocking issues found.

The loader-local index ownership, nested-object cleanup, writer cache invalidation, `SetItem` replacement semantics, sync/async dispatch, and legacy target compatibility all look correct. The operation-count tests prevent the original `IndexOfItem` scan from returning.

Proof on the reviewed commit:

- full `net8.0` suite: 3,524 passed, 0 failed
- 15K duplicate attack: parse 30 ms, deserialize 33 ms
- 15K unique control: parse 34 ms, deserialize 37 ms
- 30K duplicate attack: parse 81 ms, deserialize 79 ms
- 30K unique control: parse 70 ms, deserialize 68 ms
- duplicate results retained the expected counts and final values `14999` / `29999`

Residual test risk: the deterministic tests pin removal of the known linear `IndexOfItem` path, but do not instrument how often the lazy index is rebuilt. The implementation was reviewed directly for one-time backfill and the public-path scaling check covers that behavior.

PR #3125, opened 2026-08-13
Remove .NET Standard 1.x support

JamesNK:
## Summary
- remove the `netstandard1.0` and `netstandard1.3` package assets and their mapped test hosts
- simplify obsolete conditional compilation and remove redundant test dependencies and package validation suppressions
- derive release versions from `Build/version.json` and upgrade local/CI SDK configuration to .NET 10 LTS

## Validation
- `Build/localbuild.ps1` passes under .NET SDK 10.0.300
- all retained test targets pass: 22,284 tests across net20, net35, net40, net46, net5.0, net6.0, and net8.0
- public API counts are unchanged for every retained library target, with strict ApiCompat comparisons passing
- package validation passes and the package contains exactly net20, net35, net40, net45, net6.0, net8.0, and netstandard2.0 assets


Build/build.ps1
```diff
@@ -152,20 +154,60 @@ task Test -depends Build {
 function NetCliBuild()
 {
   $projectPath = "$sourceDir\Newtonsoft.Json.slnx"
+  $originalLocation = Get-Location
   $libraryFrameworks = ($script:enabledBuilds | Select-Object @{Name="Framework";Expression={$_.Framework}} | select -expand Framework) -join ";"
   $testFrameworks = ($script:enabledBuilds | Select-Object @{Name="Resolved";Expression={if ($_.TestFramework -ne $null) { $_.TestFramework } else { $_.Framework }}} | select -expand Resolved) -join ";"
+  $libraryFrameworkArgument = '/p:LibraryFrameworks=\"{0}\"' -f $libraryFrameworks
```
JamesNK: This fixes Windows PowerShell 5, but breaks the normal `pwsh` path used by `Build/localbuild.ps1`. I reproduced it with `& .\Build\runbuild.ps1 -parameters @{taskList='Build'} -properties @{ensureNetCliSdk=$false; buildNuGet=$false; treatWarningsAsErrors=$true}`: MSBuild reports `MSB1006` and treats `netstandard2.0` as a separate switch. The literal `\"...\"` quoting depends on legacy native-argument marshalling. Please use one argument representation that is proven under both Windows PowerShell 5 and PowerShell 7, and add both entry points to validation.

JamesNK (commented):
One blocking cross-shell issue found. The Windows PowerShell CI path is green, but the same build fails under PowerShell 7/localbuild because the MSBuild framework-list quoting is host-dependent. Other panel observations were either pre-existing or non-blocking cleanup.

Build/build.ps1
JamesNK: Fixed in 99652c9d by making PowerShell 7 the required build host, switching the CMD wrapper and Azure steps to `pwsh`, and passing framework lists as scoped MSBuild environment properties. Verified a clean PowerShell 7 build plus full tests/package pipeline; Windows PowerShell 5 now fails immediately with `ScriptRequiresUnmatchedPSVersion`.

PR #3126, opened 2026-09-09
fix: use read async with cancel token when available

sp-jcberleur:
Pass `CancellationToken` to `TextReader.ReadAsync` when the method is available (.NET Core 2.1+)

I put .net 6+ because thats the supported versions


JamesNK (approved):

JamesNK:
Thanks!

